PR #902, opened 2015-01-08
add-on heat pump final demand

ChaelKruip:
Added electricity use of add-on heat pump to space heating final demand chart.

Before:
energietransitiemodel_-_je_eigen_onafhankelijke__volledige_en_op_feiten_gebaseerde_energiescenario_

After:


AlexanderWirtz:
@ChaelKruip What I do not yet understand is why we do not ad the ambient heat and cold here. The converter does use this, as do the other heat pumps. It is not intuitive to me or the user why in this case it is not shown. I understand that this converter works a little different by reducing useful demand, but I do not see why we cannot simply add the ambient hear demand to the chart.

ChaelKruip:
> What I do not yet understand is why we do not ad the ambient heat and cold here. 

I forgot :blush: 
Very good point!

After adding it:

PR #913, opened 2015-01-28
Updated SER 2020 and 2023 and Urgenda presets

ChaelKruip:
Because of fuel cost updates.


AlexanderWirtz:
Urgenda scenario is not correct. Waaaaaay too many slider changes. 
The SER 2020 and SER 2023 slider settings are OK

Issue #921, opened 2015-05-14
P2H node should be properly connected to the graph


ChaelKruip:
Currently it has no connections. I think it should replace the add-on heat pump for hot water. 
The reasons for this are:
* The add-on heat pump currently does not produce hot water (only space heating) so the edge from the add-on to hot water could be connected to the P2H node instead
![etengine_dev_data_latest_layout_edit](https://cloud.githubusercontent.com/assets/1303760/7631645/2c16aba0-fa45-11e4-9c08-6176ee10479c.png)
* The P2H node is similar to the add-on heat pump (and solar thermal panels) in the sense that it **reduces** the hot water demand rather than fulfils it. 

The reason I am not doing this right now is that it is quite an operation with a lot of renaming of nodes, edges, gqueries, update statements and node analyses.

ChaelKruip:
I am logging what is being changed for compatibility reasons with ETDataset:

- [ ] rename `households_useful_demand_for_hot_water_for_houses_with_add_on` -> households_useful_demand_for_hot_water_for_houses_with_p2h
- [ ] rename `households_space_heater_heatpump_add_on_electricity-households_useful_demand_for_hot_water_for_houses_with_add_on@useable_heat` -> `households_flexibility_p2h_electricity-households_useful_demand_for_hot_water_for_houses_with_add_on@useable_heat`

ChaelKruip:
TODO:

- [ ] fix shares in `households_useful_demand_for_hot_water_after_solar_heater_and_add_on-households_useful_demand_for_hot_water_for_houses_with_p2h@useable_heat`

jorisberkhout:
Changing milestone to 'Deploy January 2016'.

PR #923, opened 2015-06-09
Storage edges

ChaelKruip:
This pull request includes the changes described in [this commit](https://github.com/quintel/etsource/issues/921)

Summarising:
* The P2H converter `households_flexibility_p2h_electricity` has been connected to the hot-water part of the households sector.
* The converters and nodes in the hot water sector have been renamed in accordance with the new node 
* gqueries have been updated to use the new names

The graph used to look like this:

Now it looks like this:

TODO: the initialisation for the new nodes need to be made and the renaming should be extended to ETDataset (see https://github.com/quintel/etdataset/issues/607)


jorisberkhout:
Nice solution, @ChaelKruip ! I will work on the open item you mentioned in https://github.com/quintel/etsource/issues/921#issuecomment-108346139 as there is a strong relation to ETDataset.

PR #928, opened 2015-08-04
Added rake task to update carrier attributes.

jorisberkhout:
I added a rake task that makes it possible to update the attributes defined in a `<carrier>.ad`. This rake task is almost identical to the rake task used to update the attributes for a `<node>.ad`.

The new rake task will be called from the `<carrier_source_analyses>.xlsx` on ETDataset using a VBA script.

See https://github.com/quintel/etdataset/pull/613 for more details.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -142,6 +142,61 @@ namespace :import do
 
     node.save
   end # :node
+
+  desc <<-DESC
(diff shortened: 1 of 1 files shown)


antw:
:+1:

PR #931, opened 2015-08-06
Moving co2 values to datasets

grdw:
Tested by hand with @jorisberkhout 

Keep in mind that `co2_conversion_per_mj:` is copied from `carriers`.


AlexanderWirtz:
merged

PR #932, opened 2015-08-06
Fixed input statement to be in line with other input statements in this share group; fixes failing turk

jorisberkhout:
After the mechanical turk failed I noticed that the input statement for `households_water_heater_heatpump_air_water_electricity` (which caused the failing spec) looked different from the others in the `water_heater` share group. I updated the input statement to be in line with the others. Locally the Turk passes again.


ChaelKruip:
@jorisberkhout thanks for fixing this. I don't understand why it would break in the first place though... @antw do you have any idea how this input statement ends up tripping up the Mechanical Turk all of a sudden?

PR #941, opened 2015-08-24
Initialisation files for P2H node and edges

### LINKED ISSUE #921: P2H node should be properly connected to the graph

jorisberkhout:
This pull request is a follow-up of https://github.com/quintel/etsource/pull/923 and replaces the hard-coded values that were introduced in several edges by proper parent / child shares. The current pull request also contains these parent / child share files for all datasets. The ETDataset part of these changes is in a separate pull request.

Closes #921.


jorisberkhout:
Closing this pull request as this was not working as expected. A new pull request will follow soon.

PR #945, opened 2015-09-01
Update inputs & queries to reflect addition of P2H node

antw:
Updates start_value queries which expected a link between a converter and `..._after_solar_heater_and_add_on`. Those converters now link to `..._after_solar_heater_and_p2h`.

This pull request fixes the queries and should restore the correct start values for hot water sliders.


ChaelKruip:
:+1:

PR #952, opened 2015-09-15
Improved input for setting full load hours in scenario files. 

AlexanderWirtz:
The final commit undoes the first two commits.
Setting the priority to 3, means the power plant FLH updates results in updated power and heat production. This works like a charm locally


AlexanderWirtz:
assigning @dennisschoenmakers

dennisschoenmakers:
Excellent @AlexanderWirtz !

PR #959, opened 2015-09-28
Rli presets ad files created. 

AlexanderWirtz:
@ChaelKruip assigned. Including @jorisberkhout to notify him

**Please note** that I am not entirely sure I did everything required to make this work. It works fine on my local machine. Some things did strike me though:

- the `- id` entry in the other presets files sometimes refers to a existing and corresponding scenarios on etengine, but not in all cases (e.g. SER2020 and 2023). 
- I did use the etengine scenario id for the `- id` entry. On my local machine this means I cannot load that particular scenario anymore using: `http://etmodel.dev/scenarios/423879` for example, as that now refers to the preset scenario. Since these two are identical, this is not a problem, of course. What happens if I overwrite the settings in the presets file or in the etengine scenario `user_values`, though?

### To be done
- [ ] After deploy, the redirect links for the Rli scenarios need to be updated to point to the preset scenarios on [etcentral](https://github.com/quintel/etcentral/blob/master/config/routes.rb)


AlexanderWirtz:
have you imported etsource?

ChaelKruip:
> have you imported etsource?

Yes.

PR #967, opened 2015-09-29
CO2 of produced electricity now includes units

RobTerwel:
Fixes https://github.com/quintel/etmodel/issues/1924


jorisberkhout:
Re-assigning to @ChaelKruip as this pull request is only part of the fix and he knows what else to do.

PR #969, opened 2015-10-05
Fixing shares for P2H

### LINKED ISSUE #921: P2H node should be properly connected to the graph

jorisberkhout:
This pull request includes:

- Added `residences_useful_demand_for_hot_water_after_solar_heater_and_p2h_parent_share` and updated the corresponding edges
- Added `residences_useful_demand_for_hot_water_for_houses_with_p2h_child_share` and updated corresponding edges
- Removed the no longer needed parent share `residences_space_heater_heatpump_add_on_electricity_parent_share` for all datasets and updated corresponding edges
- Removed the no longer needed parent share `residences_useful_demand_for_hot_water_after_solar_heater_and_add_on_parent_share` for all datasets and updated corresponding edges
- Added P2H shares for all datasets generated by the new residences analysis, see https://github.com/quintel/etdataset/pull/627.

Closes #921.


ChaelKruip:
:+1: :+1:

PR #983, opened 2015-10-28
LNG for Trucks, Ships and Regasification

RobTerwel:
Contains the necessary nodes and edges, including demands and shares (as obtained  from ET-Dataset).

Point of attention: the regasification terminal's efficiency is set to 1, as otherwise a smaller amount of regasified lng than reported would enter the national gas grid as the graph is initialised through the demand of `energy regasification lng`. Setting an `input.natural_gas` for `national gas network natural gas` won't do the trick, since it also receives natural gas from `energy treatment natural gas` - it would amount to setting a hard share for the natural gas/greengas ratio.


nodes/transport/transport_road_mixer_lng.converter.ad
```diff
@@ -0,0 +1,10 @@
+- use = energetic
+- energy_balance_group = local mixer
+- groups = []
+- free_co2_factor = 0.0
+
+#~ input.lng = EB(road,natural_gas) * SHARE(transport_final_demand_for_road_network_gas_parent_share, transport_final_demand_for_road_lng) / (EB(road,natural_gas) * SHARE(transport_final_demand_for_road_network_gas_parent_share, transport_final_demand_for_road_lng) + EB(road,biogases))
```
jorisberkhout: @RobTerwel, could you delete this line and the one below?

nodes/transport/transport_final_demand_diesel.final_demand.ad
```diff
@@ -3,4 +3,4 @@
 - groups = [final_demand_group]
 - free_co2_factor = 0.0
 
-~ demand = EB("transport", "gas/diesel_oil")
+~ demand = EB("transport", "gas/diesel_oil") - EB("non-specified (transport)", "gas/diesel_oil") - EB("pipeline transport", "gas/diesel_oil")
```
jorisberkhout: @RobTerwel , this means that we are ignoring any energy that is reported under `non-specified (transport)` and `pipeline transport`, right? Could you think of a fix for this?

RobTerwel:
> this means that we are ignoring any energy that is reported under non-specified (transport) and pipeline transport, right? Could you think of a fix for this?

Yes, this is what I mentioned to you earlier - I'll create a new issue for this.

PR #990, opened 2015-11-18
Efficiency update for CNG, diesel, electric and gasoline trucks

RobTerwel:
Update of  the output.truck_kms element of these converters to more realistic values (see references to ETDataset commits).


jorisberkhout:
:+1:

PR #1000, opened 2015-12-09
Includes LNG truck in the efficiency improvement truck input statement

RobTerwel:
As spotted by John, the LNG truck was not included in the truck efficiency improvement slider. In this pull request, the LNG truck is added to that input statement.


jorisberkhout:
:+1:

Issue #999, opened 2015-12-04
Add-on heat pumps need to be replaced by hybrid heat pumps in scenario's


ChaelKruip:
@antw didn't we have a script/task to update scenario's?

antw:
We have a script for ensuring share groups balance, and that each input is within the min/max, but not for changing one input into another.

It will be trivial to do this for presets using Atlas, and we can create an ETEngine migration to update the existing scenarios.

PR #1004, opened 2015-12-09
One node hhp

ChaelKruip:
This adds two hybrid heat pumps to the ETM:

* households_space_heater_hybrid_heatpump_air_water_electricity
* households_water_heater_hybrid_heatpump_air_water_electricity

The add-on heat pump has been removed. TODO: the presets still have to be updated (see https://github.com/quintel/etsource/issues/999).

The graph for space heating (before):

beta_et-engine_com_data_latest_layout_edit

The graph for space heating (after):

The Turks fail only on the RLI scenario's which is expected.


ChaelKruip:
> What are these to be replaced with? I can do that quite quickly for the presets.

Unfortunately, it is not that easy. The hybrid heat pump is a combination of a heat pump add on (removed) and a combi boiler (still there). So the removed add-ons would need to be replace by hybrid heat pumps while at the same time removing the equivalent number of combi boilers. I would have to look into the effects of this operation on some important presets before commencing. The PR can be merged IMO before I make the corrections.

Issue #972, opened 2015-10-08
Woody biomass should not be included in bio-footprint calculation


ChaelKruip:
See [the relevant node](https://github.com/quintel/etsource/blob/master/nodes/energy/energy_production_woody_biomass.ad#L3).

Because it does not have a footprint, there is currently no effect on the calculation.

PR #1017, opened 2015-12-23
Removed biofootprint from energy_production_woody_biomass (this node …

ChaelKruip:
…should not contribute to the biofootprint)

Fixes https://github.com/quintel/etsource/issues/972

@AlexanderWirtz why should energy_production_woody_biomass not contribute to the bio-footprint again?


AlexanderWirtz:
@ChaelKruip the reason is that this carrier (which was renamed from `bio_residues_for_firing`) refers to a waste stream of woody biomass. It is therefore not grown and requires no specific area.

Issue #943, opened 2015-08-27
agriculture_chp_supercritical_wood_pellets.ad needs to be renamed


ChaelKruip:
This should become `investment_costs_combustion_biomass_plant.ad`.

ChaelKruip:
The translation keys should be renamed as well. Don't forget to update the scenario's and presets :smile:

ChaelKruip:
`buildings_useful_demand_for_appliances` needs to be renamed to something involving efficiency.

ChaelKruip:
`buildings_useful_demand_cooling_electricity` needs to be renamed to refer to demand for electricity instead of cooling.

PR #1019, opened 2015-12-23
Renaming nodes and inputs

ChaelKruip:
Fixes part of https://github.com/quintel/etsource/issues/943


ChaelKruip:
@antw the same renaming of inputs need to happen in the scenario's stored on the database.

PR #1020, opened 2016-01-06
Hydrogen 2.0

RobTerwel:
In this PR the following additions are made:

- hydrogen compressor
- local electrolysis as a production pathway
- biomass gasification, with and without CCS (but will stay in the backend)
- all necessary edges for the above
- shares for the hydrogen production techniques (which will be exported from ETDataset in the next version)
- input statements for hydrogen production technologies, FCEVs and FCEV efficiency improvements

And some minor cleaning up.


jorisberkhout:
Looks good, @RobTerwel !

PR #1023, opened 2016-01-13
Hydrogen gqueries and new hydrogen production chart 169

RobTerwel:
Besides creating new carrier_groups for hydrogen, this PR adds hydrogen gqueries to the following charts

- 23 use of final demand in transport
- 43 use of transport fuels
- 49 renewability of final demand
- 123 sankey
- 146 final demand (MECE)
 - in order to have ETE pick up these gqueries , it was necessary to add a H2 final demand edge to the [final_demand] group
- 165 primary CO2 emissions of final demand
- 169 a new chart for hydrogen production

This 'should' be an exhaustive list of the charts hydrogen for transport needs to be part of; however, with the P2G implementation more gqueries will have to be added (a preliminary set of which I have locally).


jorisberkhout:
:+1:

PR #1025, opened 2016-01-15
Transport efficiency improvement input statements correction

RobTerwel:
Currently, the efficiency improvement sliders for trains, planes and ships do not behave as they should: they decrease the useful demand rather than improve the efficiency. This can be seen from the current statement on <a href="https://github.com/quintel/etsource/blob/master/inputs/demand/transport/transport_efficiency/transport_useful_demand_planes_efficiency.ad" >master</a>. This is the exact opposite of what happens 
in  the <a href="https://github.com/quintel/etsource/blob/master/inputs/demand/transport/transport_plane/transport_useful_demand_planes.ad" >mobility growth</a> input statement, where the useful demand is updated.

However, the fundamental point is here that efficiency improvement should not change the useful demand, i.e. in this case the kms driven. Efficiency improvements should ultimately change the demand of the specific technology converter, its output is still the same (i.e., the quantity required to meet the same useful demand). In other words, cars do not start driving more when their efficiencies increase - their fuel requirements are just smaller.

Cars and trucks already had correct efficiency improvement input statements.


jorisberkhout:
Well spotted, @RobTerwel !

ChaelKruip:
> ... they decrease the useful demand rather than improve the efficiency.

@RobTerwel As you correctly observe, this applies to many more 'efficiency sliders' in the ETM. Could you perhaps create a ticket about this?

PR #1026, opened 2016-01-15
Removed unnecessary update queries for EVs for loss conversion

RobTerwel:
As spotted by @jorisberkhout , the EV efficiency improvement input statement updates the loss output of `transport_car_using_electricity` and `transport_truck_using_electricity`. Since they have the attribute
``` 
output.loss = elastic
```

manipulating it with some numbers does not have any effect.


jorisberkhout:
:+1:

PR #1033, opened 2016-01-21
HHP insulation formula and input share fix, closes #1008

### LINKED ISSUE #1008: HHP IF-statement instead of fudge-factor in input statements
As of now, all HHP updating input statements have an ugly fudge factor in the insulation_fraction_weighted_average formula:

```Insulation_fraction_weighted_average = (insulation_fraction_new_houses * INPUT_VALUE(households_number_of_new_houses) + insulation_fraction_old_houses * INPUT_VALUE(households_number_of_old_houses)) / (INPUT_VALUE(households_number_of_new_houses) + INPUT_VALUE(households_number_of_old_houses) + 0.0000000000000000001);```

This is done to make sure that the denominator does not become zero, which is possible if a user decides to destroy all houses.

This fudge factor should be replaced by an IF-statement which (for instance) sets the `insulation_fraction_weighted average` to 1 IF `INPUT_VALUE(households_number_of_new_houses) + INPUT_VALUE(households_number_of_old_houses) = 0`

Re-assign at will.

RobTerwel:
This PR fixes some bugs:

- previously, the HHP insulation formula had a discontinuity since it divided a quantity mapped on [0,1] by  the efficiency of the gas part of the HHP. Now, we no longer divide by this and the mapping is continuous and smooth.
- previously, the HHP insulation formula had a fudge factor, this PR replaces it with an IF-statement and closes #1008 
- previously, the input.network_gas attribute of the HHP was in fact the share of heat output delivered by the gas part of the HHP. Now we divide by the gas_efficiency to obtain the correct input.network_gas.

### DIFF
--- a/inputs/demand/households/households_misc/households_insulation_level_new_houses.ad
+++ b/inputs/demand/households/households_misc/households_insulation_level_new_houses.ad
@@ -4,13 +4,9 @@
 #  
 #  COP = 4.5;
 #
-# This is the efficiency of the gas part of the HHP   
-#    
-#   gas_efficiency = 1.067;
-#
 # This is the initial input.network_gas attribute of the HHP
 #    
-#    initial_fraction_gas_to_heatpump = 0.2;
+#    initial_fraction_gas_to_heatpump = 0.187441424554827;
 #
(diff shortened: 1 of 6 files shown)


ChaelKruip:
Excellent work @RobTerwel! :+1:

RobTerwel:
Thanks, @ChaelKruip !

PR #1034, opened 2016-01-22
Made all HHP input statements pick up initial shares from the nodes

### LINKED ISSUE #1009: HHP input statement should get initial values from node instead of having them hard-coded
Currently all HHP updating input statements have hard-coded initial values. Now we set

``` 
    COP = 4.5;
    
    gas_efficiency = 1.067;

    initial_fraction_gas_to_heatpump = 0.2;
````
While these in fact are the following attributes of the HHP space heating node

```
COP = (input.ambient_heat + input.electricity) / input.electricity
    gas_efficiency = output.useable_heat
    initial_fraction_gas_to_heatpump = input.network_gas
```

We tried obtaining these through a query such as (for the latter) `Present:V(households_space_heater_hybrid_heatpump_air_water_electricity, network_gas_input_conversion)` but writing a query inside a query does not work. Perhaps CSV's could be a way to transmit the information.

Re-assign at will.

RobTerwel:
closes #1009


ChaelKruip:
Very nice indeed @RobTerwel!

Issue #1029, opened 2016-01-19
Exclude EB non-specified and pipeline energy use from demand initialised transport FD nodes


RobTerwel:
All FD_[carrier] transport nodes which have their demand EB-set and -initialized in the form of `EB(transport, [carrier]) `should have this value subtracted by `EB(pipeline transport,[carrier])` and `EB(non-specified,[carrier])`. 

## Explanation

This has to happen because currently the total, **EB** reported final demand of `non_specified` and that for `pipeline_transport`  energy use in *transport* is aggregated in the *other sector* in the **ETM** <a href="https://github.com/quintel/etsource/blob/master/nodes/other/other_final_demand_crude_oil_non_energetic.final_demand.ad#L30">here<a/> and <a href="https://github.com/quintel/etsource/blob/master/nodes/other/other_final_demand_network_gas.final_demand.ad#L9">here<a/>, respectively. 
They read:
````
 EB("Pipeline_transport", total)
EB("Non-specified_(transport)", "total")
````
But by initialising a transport FD node with

````
 EB(transport,[carrier])
````
we are actually counting
````
EB(transport,[carrier]) = EB(road,[carrier]) + EB(rail,[carrier]) + EB(aviation,[carrier]) + **EB(Pipeline_transport,[carrier])** + EB(domestic_navigation,[carrier]) + **EB("Non-specified_(transport)",[carrier])**
````
hence, we are potentially counting energy *twice*. This happens for any non-zero `pipeline_transport` and `non-specified`. And as it turns out, in several cases we are (e.g. for electricity, around 50 PJ for the EU dataset)). Therefore we should remove this EB reported `pipeline_transport` and `non-energy` demand from the transport final demand [carrier] nodes.

PR #1036, opened 2016-01-22
Removing non_specified and pipeline_transport from transport FD EB-initialised demand nodes

RobTerwel:
For more information, see the explanation in #1029

### DIFF
--- a/nodes/transport/transport_final_demand_bio_ethanol.final_demand.ad
+++ b/nodes/transport/transport_final_demand_bio_ethanol.final_demand.ad
@@ -3,4 +3,7 @@
 - groups = [final_demand_group]
 - free_co2_factor = 0.0
 
-~ demand = EB("transport", biogasoline)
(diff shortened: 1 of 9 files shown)


jorisberkhout:
Merging this as it indeed fixes the bug described in #1029 .

This issue did however make me realise that mistakes / bugs like the one described in #1029 are easily introduced; there is no check for consistency between the results of the dataset generation process and manipulations on the Energy Balance like the ones that are part of this pull request (and are done in more nodes on ETSource). Part of this concern is already voiced in https://github.com/quintel/etdataset/issues/295. When time permits, I will use that issue as a starting point for a discussion on a more permanent solution to issues like #1029.

Notifying @AlexanderWirtz and @ChaelKruip and thanking @RobTerwel !

AlexanderWirtz:
Well spotted @RobTerwel

PR #1045, opened 2016-01-27
Fixed typo's in attributes for some hydrogen converters 

RobTerwel:
As it turns out, there were some typo's in the `decommissioning_costs` and `variable_operation_and_maintenance_costs_per_full_load_hour` attributes in the node source analyses which were therefore not exported. This was fixed in quintel/etdataset@185028a61c436a9a851e1032acfe0d3abb95cc84 . In addition, there was an incorrect cell reference for the `fixed_operation_and_maintenance_costs_per_year` attribute in the electrolysis node source analysis which has been fixed in quintel/etdataset@ed54f4ee67656effc9040b05ee4a2a5660929594 .

This PR features the associated ETSource updates and makes the `number_of_units` calculation for all hydrogen converters in the energy sector fully functional.


jorisberkhout:
:+1:

PR #1046, opened 2016-01-27
Updated LNG and bio LNG regasification converters based on 

RobTerwel:
quintel/etdataset@50b1e0a63dde8d06278c3508c487f9a6e5df822c

Necessary for `number_of_units` calculation.


jorisberkhout:
:+1:

PR #1053, opened 2016-02-01
Restored previously deleted mech_turk query which was used in multiple tests

RobTerwel:
The `energy_power_ultra_supercritical_coal_demand` mech_turk query was replaced with `energy_power_supercritical_coal_demand` in quintel/etsource@449ad3c04103fc3dc7e630491322d08f61e7a4 . However, it was used in multiple mech_turk tests, so this PR restores it.


jorisberkhout:
:+1:

Issue #1038, opened 2016-01-25
Country-specific initial share or start_value_gql pick up from node


RobTerwel:
## Background

Nodes can retrieve certain country-specific data from files (mostly .csv and .ad) as outlined in <a href="https://github.com/quintel/atlas/blob/master/lib/atlas/runtime.rb"a>runtime.rb<a/> in Atlas. 

In <a href="https://github.com/quintel/etengine/commit/13dd4ce584b1fa33ef764b1105f9dc0aebe3a5af"a>these commits<a/>, @antw implemented a way to pick up the `number_of_units` attribute from a node which enabled us to have car technology input statements set the `number_of_units` for the **future**. However, the `number_of_units` node attribute for the **present** for car technologies is country specific and should be equal to the share of the useful demand times the total number of cars in that country: 

`
QUERY_PRESENT(-> { V(transport_car_using_hydrogen, share_of_transport_useful_demand_car_kms) }) * AREA(number_of_cars)
`

or 

`(present:)V(transport_car_using_hydrogen_share) * AREA(number_of_cars)
`
## Problem

Although nodes are already able to pick up `AREA(number_of_cars)`, the problem is that ETEngine and Atlas as of now (to my knowledge) don't enable nodes to pick up these shares from queries. We do however need the present share of the useful demand, since we would otherwise face a discontinuity when one of these input statements - affecting the future - is moved and as a result changes the `number_of_units` value from 0 to the newly calculated value. The question is how to retrieve this information on the initial (i.e. present) useful demand shares and get it to the node.

@antw , 
-  is it currently possible to have a node pick up the initial shares as calculated by Atlas, and if so, how? 
-  if not, it is it possible to make this happen and if so, how hard and time-consuming would that be?

Thanks in advance!

You can find the 'work in progress' containing edited input statements on the branch `h2_number_scaling` on ETSource.

P.S. I know it is possible to export csv's containing the shares of UD from the transport analysis, but this would amount to mimicking Atlas in ETDataset and therefore duplicating calculations, introducing unnecessary risks and complexity.

antw:
> is it currently possible to have a node pick up the initial shares as calculated by Atlas, and if so, how?

[...]

PR #1060, opened 2016-02-05
Additional investment costs slider for EVs, closes #800

### LINKED ISSUE #800: Electric vehicles need 'additional costs'  attribute wrt fossil fueled cars
After discussions with Ecofys we agreed to investigate the possibility of including  a costs parameter for vehicles.  Currently, the assumption is that vehicles cost the same regardless of their drive train technology. 

Because it would create different (unrecognized) cost numbers in the dashboard, I suggest instead of giving all cars and trucks cost attributes, we instead give them only to electric ones, in the form of `additional_costs`. This only applies to `installation costs` or a similar parameter, of course, as fuel costs are  taken care of automatically.

@jorisberkhout @wouterterlouw @ChaelKruip please comment. The data will be supplied by Ecofys.

Also see #797

RobTerwel:
Largely based on https://github.com/quintel/etdataset/pull/670 . This PR adds:

- a cost calculation for electric cars
- the additional investment costs slider for EVs
- a chart that shows the investment costs of combustion engine cars and that of electric cars
- a dataset-specific `number_of_cars` AREA attribute
- a `number_of_units` calculation for all cars which is based on the useful demand slider setting and `number_of_cars`

https://github.com/quintel/etsource/issues/1038 is still open but not required for the front-end given the current implementation.

Also closes the deploy parts of https://github.com/quintel/etdataset/issues/635 .


jorisberkhout:
:+1:

PR #1068, opened 2016-02-09
Updated input statements for car tech share for 'present' update period

AlexanderWirtz:
as well as the input statement for car kms useful demand for 'both' update periods to reflect changes introduced by #1060. Also cleaned up input statement for 'future' hydrogen car share.


RobTerwel:
Thanks @AlexanderWirtz , tested locally and it works like a charm! The `number_of_cars` AREA attribute is scaled to 519 in the present. Is that about accurate?

PR #1079, opened 2016-02-10
Prevent double counting output losses in storage

antw:
Ref quintel/etmodel#2070

Storage loss queries account for the input and output losses of storage technologies. Demand queries were querying the `demand` of the storage converters, which is the amount of energy after input loss, but *before* output loss. Therefore, output losses were being counted twice.

This PR changes the queries to show the amount of storage energy *after* all loss has been subtracted.


jorisberkhout:
:+1:

PR #1080, opened 2016-02-12
Carrier prices, dollar exchange rate and co2 prices updated. 

AlexanderWirtz:
Also updated NL arable land area and 1990 CO2 emissions. 
Also see quintel/etdataset/pull/674


jorisberkhout:
:+1:

PR #1097, opened 2016-03-15
Removed bio_oil and bio_ethanol from initial_updates_preset_demands input statement

### LINKED ISSUE #1093: Initial_updates_preset_demands input statement features carriers which lack a timecurve
The <a href= "https://github.com/quintel/etsource/blob/master/inputs/misc/initial_updates_preset_demands.ad"a>initial_updates_preset_demands<a/> input statement returns the production of a certain carrier from the timecurves based on the scenario year the user selects:

``
UPDATE( V(energy_production_bio_oil), preset_demand, TIME_SERIE_VALUE(energy_production_bio_oil,preset_demand,USER_INPUT())
      )
``

However, for `bio_oil` and `bio_ethanol`, these timecurves do not exist. As a result, this input updates the production with the value 0 and the primary production in the future is always zero; hence, we import that carrier based on the scenario's demand. This has implications for import, export and the biofootprint calculation.

To fix this, we could either add timecurves for `bio_oil` and `bio_ethanol` (the ETDataset way) or remove them from this input statement (which amounts to assuming that the future production will always be equal to that of the dataset's year's) (the ETSource way).

@ChaelKruip @jorisberkhout @AlexanderWirtz , what are your thoughts and preferences?

RobTerwel:
Since there are no timecurves for `bio_oil` and `bio_ethanol`, their production is always updated to 0. With this PR their production is assumed to stay constant and equal to the production of the initial year. 
Closes #1093 .

### DIFF
--- a/inputs/misc/initial_updates_preset_demands.ad
+++ b/inputs/misc/initial_updates_preset_demands.ad
@@ -1,10 +1,6 @@
 - query =
     EACH(
       UPDATE(
-        V(energy_production_bio_oil), preset_demand,
-        TIME_SERIE_VALUE(energy_production_bio_oil,preset_demand,USER_INPUT())
-      ),
-      UPDATE(
         V(energy_extraction_natural_gas), preset_demand,
         TIME_SERIE_VALUE(energy_extraction_natural_gas,preset_demand,USER_INPUT())
       ),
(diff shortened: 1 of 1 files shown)


ChaelKruip:
:+1:

PR #1111, opened 2016-04-29
Added compressed_network_gas to FD MECE network gas for transport query

### LINKED ISSUE #1110: compressed network gas should be included in final demand queries
Currently, it does not show up in the Mekko:

This query does not contain the carrier:
[https://github.com/quintel/etsource/blob/master/gqueries/general/final_demand/mece/final_demand_of_natural_gas_and_derivatives_in_transport.gql](https://github.com/quintel/etsource/blob/master/gqueries/general/final_demand/mece/final_demand_of_natural_gas_and_derivatives_in_transport.gql)

RobTerwel:
Closes #1110.

### DIFF
--- a/gqueries/general/carrier_groups/all_carriers/carrier_group_natural_gas.gql
+++ b/gqueries/general/carrier_groups/all_carriers/carrier_group_natural_gas.gql
@@ -1,4 +1,4 @@
 # Carrier group of natural gas
 
-- query = CARRIER(natural_gas, low_caloric_gas, high_caloric_gas, lng)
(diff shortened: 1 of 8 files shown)


ChaelKruip:
@RobTerwel do you think there might be other place where this carrier needs to be added now that we are at it? 😄

RobTerwel:
> @RobTerwel do you think there might be other place where this carrier needs to be added now that we are at it? 😄

I'll go over it again; the carrier already existed before the introduction of the compressor, but I agree a little searching wouldn't hurt!

RobTerwel:
Good point you raised there, apparently it has never been part of many queries! I will add it right away. 🎅

RobTerwel:
This should be it. 😄

PR #1118, opened 2016-05-18
Added inputs to tweak the demand of several carriers used by appliances

RobTerwel:
Our interface (ETModel) does not allow the user to alter all energy flows. For scaled scenarios these unchangeable flows can become quite noticeable: appliances in the buildings sector have a non-zero demand for coal, wood pellets, crude oil and network gas. This PR features inputs (`update_period = both`) which allow us to 'reduce' and remove these energy flows in the present and future simultaneously.


AlexanderWirtz:
notifying @dennisschoenmakers and @ChaelKruip 

We needed some more input statements to 'tweak' regional scenarios

PR #1120, opened 2016-05-19
Made inputs update the demand of appliances (instead of the FD) in buildings sector

RobTerwel:
This is a follow-up to https://github.com/quintel/etsource/pull/1118 . Setting the demand of some FD converters in the buildings sector to zero was reflected in the dashboard, but not in the graph. Nodes between the FD and UD still had non-zero demand. This PR is an attempt to fix that.


AlexanderWirtz:
Notifying @dennisschoenmakers @ChaelKruip @jorisberkhout

PR #1124, opened 2016-05-24
Belatedly updated rli preset scenario prices based on 2015 NEV number…

AlexanderWirtz:
…s (also used in GEA 2050 scenario for example).

I suppose there is no way this can be pushed to Live, right? I forgot these were preset scenario's and updated the database files instead...


ChaelKruip:
> I suppose there is no way this can be pushed to Live, right?

I don't see why not. @antw can you cherry-pick this commit to production?

antw:
> can you cherry-pick this commit to production?

Done.

AlexanderWirtz:
Great! Thanks @antw!

PR #1127, opened 2016-05-26
Merit table

ChaelKruip:
* Split up wind turbines in three types
* Added geothermal
* Added CSP

Also see https://github.com/quintel/etmodel/pull/2141


ChaelKruip:
Possible improvement: also split up solar PV into its four components similarly to wind.

ChaelKruip:
Fixes https://github.com/quintel/etmodel/issues/2102

jorisberkhout:
This also fixes quintel/etmodel#2103 which was assigned to me. Thanks @ChaelKruip 😄

Issue #1133, opened 2016-05-31
Dashboard cost item includes the fuel costs of just one application of a combi technology


jorisberkhout:
The mechanical turk fails since we imported 634787b28eaa87e1cbc54948d8a4394b888debd1 in Beta ETEngine. In this commit we fixed the double counting of the investment costs for the combi technologies by taking into account only the `total_costs` for the most expensive of both applications of that combi technology.

I now realise that by correcting the `total_costs`, we not only correct the `fixed_costs` but also the `variable_costs`, which include the fuel costs. Hence, the dashboard cost item now only takes the fuel costs for one of the two applications (the most expensive) into account.

I think the solution would be to replace `total_costs` with `fixed_costs` in all Gqueries in 634787b28eaa87e1cbc54948d8a4394b888debd1, but would like to sit down with @ChaelKruip before we do this to avoid any further mistakes.

I believe that fixing this issue, will also fix the mechanical turk failing.

ChaelKruip:
Well spotted @jorisberkhout! 😨

jorisberkhout:
Assigning myself.

PR #1134, opened 2016-06-01
Cost correction now only takes fixed costs into account.

ChaelKruip:
Fixes https://github.com/quintel/etsource/issues/1133


jorisberkhout:
Cool! As expected, the mechanical turk passes after this fix, as verified on my local machine!

PR #1135, opened 2016-06-08
Limit electricity import

jorisberkhout:
As discussed in https://github.com/quintel/merit/issues/139, limiting the interconnecter capacity limits the amount of electricity that can be imported. This might lead to electricity shortages. The commits in this pull request add a converter that keeps track of how much this electricity shortage actually is.

I have also reverted the commit the removed import from the merit order; as a result the interconnector capacity for import is limited if the merit order is switched on.

In addition, I have added a Gquery to keep track of how much electricity is imported on an annual basis. This Gquery is used in chart 99 (Electricity production).

Closes https://github.com/quintel/merit/issues/139.


jorisberkhout:
I ran the mechanical turk locally and it passed without issues.

ChaelKruip:
@jorisberkhout what is the marginal cost of this new merit order participant?

jorisberkhout:
> @jorisberkhout what is the marginal cost of this new merit order participant?

To be honest, I don't know. Would you know the answer, @antw ?

antw:
> what is the marginal cost of this new merit order participant?

It's given [a fake price](https://github.com/quintel/etengine/blob/6e164161d0d0c7233cdf9a2ebcff113e76e79d08/app/models/qernel/plugins/merit/import_adapter.rb#L27) to ensure it is only used after the must-runs and dispatchables.

PR #1148, opened 2016-08-08
Added newest encrypted energy balances for all datasets

RobTerwel:
The newest encrypted EBs were apparently not included in https://github.com/quintel/etsource/pull/1147


jorisberkhout:
Thanks for fixing this, @RobTerwel !

PR #1151, opened 2016-08-11
Presets with chemical industry update

antw:
This pull request updates all presets to account for the recent chemical industry changes. The script used is, for all intents and purposes, identical to the one used for ETEngine scenarios, so there shouldn't be any surprises here.

There are two commits: one to update the inputs with the industry changes, and another to fix share groups which weren't adding up to 100.


jorisberkhout:
Assigning @RobTerwel .

RobTerwel:
I checked the post migration values once more for a scenario against our expected values, and the migration seems successful once more. I opened a few scenario's, checked every industry slide with developer tools, pulled some sliders and found no errors or unexpected behaviour.

Merging time!

PR #1152, opened 2016-08-22
Added central ICT sector to the Industry

RobTerwel:
This PR introduces the central ICT sector in the industry. It solely involves its electricity use.

This electricity use is aggregated by the IEA in the `public and commercial services` post, hence the electricity of the central ICT comes from the services sector.

This PR features the required graph modifications as well as sliders and FD queries. NB: FD queries for all carrier groups have been included, although electricity is the only carrier group that is present in this industry.

Based on https://github.com/quintel/etdataset/pull/692


jorisberkhout:
:+1:

Issue #761, opened 2014-02-10
Own use in refineries is included twice


wouterterlouw:
The own use of refineries in currently included twice in the model:

1. The transport fuel converters (energy_production_diesel, _gasoline, _heavy_fuel_oil, _kerosene and _lpg) do have an efficiency of appr. 92%, calculated by the aggregate `transformation output / (transformation input + own use)`. See the *Calculation of efficiency* part of https://github.com/quintel/etsource/issues/725. This method leads to an additional crude oil consumption of 42 PJ (consumption * (1 - eff) = 482 PJ * 0.08).

2. The own use of the energy industry (except for own use in electricity, CHP and heat plants; blast furnaces and coke ovens) is included in the final demand of the industry sector. See for example the [industry_final_demand_coal](https://github.com/quintel/etsource/blob/master/nodes/industry/industry_final_demand_coal.final_demand.ad) node. This result also in the inclusion of the own use in refineries, resulting in an additional consumption of 173 PJ: 
 - 51 PJ natural gas
 - 84 PJ refinery gas
 - 10 PJ non-specified oil products
 - 11 PJ electricity
 - 16 PJ heat

In the first case, the refinery own use is dependent on the demand of transport fuels, where in the second case, the refinery own use is included in the Other industry sector and does not follow demand for refinery products.

ChaelKruip:
I think this is still relevant because the pull requests mentioned above have not been merged in the end. Re-opening.

Issue #725, opened 2014-01-14
Improve refinery modelling


wouterterlouw:
*Follow-up to issue https://github.com/quintel/etsource/issues/615*

#### Introduction

Refineries transform crude_oil into various oil products. These oil products are subsequently consumed in various sectors, including the industry (e.g. naphtha) and transport sector (e.g. gasoline and diesel).

Currently, only in the transport sector we convert `crude_oil` into more specific carriers, such as `gasoline` and we do that with an efficiency that resembles the aggregated efficiency of the refinery sector (e.g. in the [energy_production_diesel](https://github.com/quintel/etsource/blob/master/nodes/energy/energy_production_diesel.converter.ad#L3) converter).

However, all other sectors just use `crude_oil` that does not flow though a converter that accounts for the refinery losses. Nevertheless, the '`crude_oil`' these sectors use are merely 'oil_products' produced by a refinery.

For the NL 2011 this results in disregarded refinery losses for oil use outside the transport sector. In the NL the oil products use is appr. 600 PJ in the transport sector and appr. 700 PJ in the other sectors; resulting in 700 PJ / 0.92 = 760 PJ --> 60 PJ of disregarded primary crude oil demand.

#### Calculation of efficiency

On the IEA energy balance refineries are specified under transformation and own use. The aggregated efficiency of refineries is defined as `transformation output / (transformation input + own use)`, i.e. output divided by all inputs. 

![refinery](https://f.cloud.github.com/assets/4319656/1910618/ee962c62-7d1b-11e3-894c-9517f85ccb54.png)

For the EU (2011) this result in an efficiency of 26,994 PJ / (27,216 PJ + 1,934 PJ) = 0.926. See also the [*refinery efficiency analysis*](https://www.dropbox.com/s/3qmdco9qvihsa7j/20140114%20Refinery%20efficiency%20source%20analysis%20%28without%20EB%29.xlsx).

#### Solution

To model also the refinery properties for the non-transport oil-product streams, the following graph structure could be implemented:

![including refinery](https://f.cloud.github.com/assets/4319656/1910717/1be8167e-7d1e-11e3-9110-0c68dd9fb0b5.png)

1. The crude_oil_distribution converter will manage the production, import and export of crude oil.
2. The crude_oil_refinery converter will convert crude_oil into oil_products an efficiency of appr. 92%. These oil_products are delivered to the sectors or converted further in case of the transport sector.
[...]

PR #1154, opened 2016-08-24
Improved refinery modelling; fixes double counting, corrects scaling and carrier inclusion

RobTerwel:
This PR features an improved modelling of the refineries. The main issues with the current modelling, as outlined in https://github.com/quintel/etsource/issues/761 and https://github.com/quintel/etsource/issues/725 , are fixed with this PR:

- no more double counting and modelling of energy flows
- taking into account *all* carriers dynamically for refinery production (not just oil)

The refinery is **not demand-driven**. That is, the consumption of transport fuels and ultimately the user-set transport shares does not determine the output of refineries. The user will therefore be able to set both independently, allowing for greater freedom. The refinery flows (incl. import, export, production, transport consumption, industry consumption) will be communicated to the user by means of a oil/refinery sankey, which @jorisberkhout and I are working on. 

In this modelling it is assumed that refinery efficiency is equal for every pathway, i.e. the production of `gasoline` is as efficient as the production of `diesel`.

The refinery now has some additional features, a.o.

- country-dependent efficiency (as opposed to an efficiency that's the EU's)
- producing refinery_products other than transport fuels

The latter point deserves some more attention. The refinery now also produces two new carriers: `oil_products` and `refinery_gas`, both of which have very heterogeneous properties and therefore simply received the carrier attributes of `crude_oil`. As explained in https://github.com/quintel/etdataset/pull/693 , the `oil_products` (mostly naphtha and heavy fractions) are considered to be crude oil which (partially or fully) meet the non-energetic `crude_oil` demand of the petrochemical industry. In addition, `refinery_gas` is produced, which is a mixture mainly consisting of light hydrocarbons and hydrogen. In practice, (also see EB), this `refinery_gas` is used to produce heat for the refinery. In our model, this would lead to a feedback loop: heat and crude oil generating `refinery_gas`, generating heat, together with crude oil generating `refinery_gas` a.i. Therefore, this `refinery_gas` is currently not used. I will make an issue for this.

This PR closes https://github.com/quintel/etsource/issues/761 , https://github.com/quintel/etsource/issues/725 and https://github.com/quintel/etengine/issues/860


jorisberkhout:
Major improvement! Let's test the sh*t out of this 😄

PR #1161, opened 2016-09-01
Subtracted export from refinery production in dashboard cost query

RobTerwel:
Closes https://github.com/quintel/etmodel/issues/2204


ChaelKruip:
Merging this as this is in line with the previous modelling.

PR #1164, opened 2016-09-06
Included flexibility options in excess events chart

jorisberkhout:
This fixes https://github.com/quintel/etmodel/issues/2188.


ChaelKruip:
Very elegant to fix quintel/etmodel#2188 like this! 👍

PR #1170, opened 2016-10-03
Fixes https://github.com/quintel/etmodel/issues/2267.

ChaelKruip:
I'm not entirely sure if the https://github.com/quintel/etsource/blob/master/gqueries/general/costs/costs_of_refinery_in_energy_sector.gql still works properly...


ChaelKruip:
Great work @jorisberkhout! This was a tough one!

PR #1171, opened 2016-10-07
Replaced load profiles 

### LINKED ISSUE #1145: Improve load profiles for must-run technologies
As I discovered while looking at one of the new merit order charts, the load profiles for the must-run technologies (`agriculture.csv` and `buildings.csv`) are no longer aligned with the total demand profile (I would expect the five peaks in e.g. dark red to coincide with the five highest peaks of the (bright red) demand profile in the chart below).

I suggest picking this up as part of the `NL 2014 update` since 2014 starts on a different day than 2013 and we need to reconsider the profiles anyway.

Notifying @AlexanderWirtz and @ChaelKruip .

screen shot 2016-07-19 at 17 19 35

TessColijn:
Replaced the load profiles for agriculture chps is all countries depending on the year used for the dataset. The new load profiles are generated with the new script.

Fixes #1145


jorisberkhout:
Nice work @TessColijn !

PR #1188, opened 2016-11-07
Adjusted max value of new residences slider. Closes #1187

### LINKED ISSUE #1187: Slider ranges for scaled scenarios insufficient for number of old and new houses
## Problem
The slider ranges for these input statements are thwarting my efforts at building some scaled scenarios for Paddepoel Noord. :

    households_number_of_old_houses.ad
    households_number_of_new_houses.ad

range defines in input statement as follows:

    - max_value_gql = present:Q(number_of_new_residences) * 0.000001 * 10
    - min_value = 0.0

The reason is that we start off with all residences technically being `old` (they were built in the 60's). Instead I created the scenario to have 1 new residence, as otherwise the slider for new residences has a range of 0 and is greyed-out. This is clear from the definition of `max_value_gql`. 

Choosing 1 new residence maximizes the slider to 10, so there is no way I can tear down all old residences and build a whole new neighborhood. Boooh!

## Suggested fix
I suggest re-defining these `max_value_gql` as follows:

The new residences slider should have `2x` the range of the total number of residences in the present. I think like so  (works in sandbox):

    - max_value_gql = present:SUM(Q(number_of_old_residences),Q(number_of_new_residences))*2 * 0.000001

This would ensure that you can always replace all old residences with new ones and have room to build more. 
I cannot implement this on my local machine, because after the Ruby upgrade made yesterday, nothing runs locally any more and there is no standard solution for this (not one comprehensible to me, anyway)

Notifying @dennisschoenmakers @antw @DorinevanderVlies and @jorisberkhout

jorisberkhout:


AlexanderWirtz:
I have asked @jorisberkhout to pick this up, as @dennisschoenmakers is sick today and this has high priority for me.

PR #1191, opened 2016-11-07
Old new residences split scaling

AlexanderWirtz:
Please review and merge.

Needed for creating a scenario for Paddepoel.


AlexanderWirtz:
I also renamed an input statement that had a duplicate (i.e. conflicting) name. I do not understand how come this input statement does not cause any problems.

PR #1193, opened 2016-11-08
Rake task for decrypting energy balances uses a more secure and robus…

ploh:
…t way

The old method...
- did not work on some machines
- was not secure on multi-user machines


antw:
Thanks!

> was not secure on multi-user machines

I'm guessing this is because the password would appear in the process list?

ploh:
> I'm guessing this is because the password would appear in the process list?

Yep

PR #1195, opened 2016-11-09
Rake task for decrypting energy balances uses a multi-user secure way

ploh:
The old method was not secure on multi-user machines since the secret
password showed up in the (world-readable) process list while executing the task

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -308,32 +308,21 @@ desc <<-DESC
   Decrypts the energy balance for every area.
 DESC
 task :decrypt do
-  if ENV['ETS_PASSWORD']
-    puts 'Using password from ETS_PASSWORD environment variable'
-    password = ENV['ETS_PASSWORD'].strip
-  elsif File.exists?('.password')
+  if File.exists?('.password')
     puts "Using password that is stored in .password..."
-    password = File.read('.password').strip
+    puts "Start Decrypting..."
+    Dir.glob('datasets/*').each do |area|
+      if File.exists?("#{ area }/energy_balance.gpg")
+        puts "* #{ area }"
+        cmd = "gpg --batch --yes --no-use-agent --passphrase-file .password" +
+              " --output #{ area }/energy_balance.csv --decrypt #{ area }/energy_balance.gpg"
+        puts `#{ cmd }`
+      end
+    end
+    puts "Done!"
   else
-    puts "File .password not found in root."
-    puts "Please Enter Passphrase to decrypt files:"
-    password = $stdin.gets.strip
+    puts "ERROR: File .password not found in current directory!"
   end
-
-  puts "Start Decrypting..."
-  Dir.glob('datasets/*').each do |area|
-    if File.exists?("#{ area }/energy_balance.gpg")
-      puts "* #{ area }"
-      cmd = "gpg --passphrase-fd 0 -d #{ area }/energy_balance.gpg > #{ area }/energy_balance.csv"
-      pipe = IO.popen(cmd, 'r+')
-      pipe.print(password)
-      pipe.close_write
-      puts pipe.read
-      pipe.close
-     end
-  end
-
-  puts "Done!"
 end
 
 desc <<-DESC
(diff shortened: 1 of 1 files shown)


ploh:
Now, it is done by piping the password directly from the rake task into gpg without writing it in the command-line of some intermediate process

antw:
Thanks @ploh! This is a clever way of doing things, I like it. :+1:

I see the reason we do this is because the rake task also supports supplying the password as an environmental variable or using stdin. I wonder if we really need those features?

* I can't imagine anyone ever using the stdin method.
* Automated ETSource tests on Semaphore *do* use the environmental variable, but we could work around that easily:

  In the Semaphore test setup, instead of:
  ```sh
  ETS_PASSWORD=abc bundle exec rake decrypt
  ```

  ... we might ...

  ```sh
  echo "abc" > .password
  bundle exec rake decrypt
  ```

If we did that, we could just use the `gpg --passphrase-file` option instead. What do you think?

ploh:
> If we did that, we could just use the gpg --passphrase-file option instead. What do you think?

antw:
[It's here](https://semaphoreci.com/dennisschoenmakers/etsource/settings), but it seems like you don't have access. Nor does Gerard for that matter! I don't have the ability to add new collaborators to the Semaphore projects; @dennisschoenmakers could you do this?

For simplicity's sake, I've changed the setup script myself so it should be good-to-go.

jorisberkhout:
I ran `rake decrypt` on the `master` branch, which resulted in the following error message (for all datasets, I'm just showing the `nl` one) and all my `energy_balance.csv` files to be empty 😖 . I then checked out this branch, which gives the same results. It might well be that I am doing something wrong here; do you know @ploh ?

```
* datasets/nl
gpg: AES encrypted data
gpg-agent[58458]: command get_passphrase failed: Inappropriate ioctl for device
gpg: problem with the agent: Inappropriate ioctl for device
gpg: encrypted with 1 passphrase
gpg: decryption failed: No secret key
```

antw:
It does work for me. For what it's worth I'm using OSX 10.11.6, and `gpg --version` reports "gpg (GnuPG) 1.4.15".

> I ran rake decrypt [..] which resulted in the following error message

As a temporary fix, in the Rakefile find the "decrypt" task near the bottom, and replace everything between `puts "start decrypting"` and `puts "done"` with the original version. It might be useful to establish that this still works and that it's not something unrelated:

```ruby
Dir.glob('datasets/*').each do |area|
  if File.exists?("#{ area }/energy_balance.gpg")
    puts "* #{ area }"
    cmd = "gpg -d --passphrase #{ password } #{ area }/energy_balance.gpg > #{ area }/energy_balance.csv"
    puts `#{ cmd }`
  end
end
```

You might also try the proposed alternative:

```ruby
Dir.glob('datasets/*').each do |area|
  if File.exists?("#{ area }/energy_balance.gpg")
    puts "* #{ area }"
    cmd = "gpg -d --passphrase-file .password #{ area }/energy_balance.gpg > #{ area }/energy_balance.csv"
    puts `#{ cmd }`
  end
end
```

jorisberkhout:
Thanks, @antw . The first fix you describe works. I am guessing that it has something to do with the version of gpg I have installed. I installed `GnuPG 2` recently which might be the reason. I'll ask @grdw to help me out when he is back in the office.

ploh:
@jorisberkhout please check out the branch again and verify that it now works on your system

ploh:
@antw please review my changes one last time (hopefully ;) and check why semaphore failed

jorisberkhout:
> @jorisberkhout please check out the branch again and verify that it now works on your system

Unfortunately it still doesn't work. I get the same error message as before and empty `energy_balance.csv` files.

jorisberkhout:
I forgot to pull the latest changes on this branch 😊 . Everything works fine now.

antw:
> check why semaphore failed

The Semaphore failure was my error, not yours. Merging! :shipit:

Issue #1213, opened 2017-01-20
Discussion: Disallow fixed values in scaleable inputs and remove step value


antw:
Following on from quintel/etmodel#2328 and quintel/etengine#902:

It is apparently quite tricky to scale a scenario which is based on a derived region since there are two sets of "scalings" required for input values to make sense in the front-end: full dataset → derived dataset, derived-dateset scenario → scaled-down scenario.

In order to solve this, I suggest we...

### Disallow the use of fixed input values when using a scaleable `unit`

Every input with a scaleable unit should use `start_value_gql`, `min_value_gql`, and `max_value_gql`, and would not be permitted to use `start_value`, `min_value`, or `max_value`. This would eliminate the need to scale these in ETEngine; it would happen automatically by virtue of the graph and dataset having been scaled down.

IMO this would be a step in the right direction even without the problems described in quintel/etmodel#2328. I see no intuitive reason for the P2G input to have [a hard-coded range of 0 to 100000 units](https://github.com/quintel/etsource/blob/680b817ed3d83b9d319e4c2d8816d44fe088418b/inputs/flexibility/energy/number_of_energy_flexibility_p2g_electricity.ad). Surely we would expect a German scenario to be able to install far more than a Netherlands scenario? This is probably true for many (all?) scaleable inputs with hard-coded values.

At present, there are [85 such inputs](https://gist.github.com/antw/fc64b1357efdbe694cac3c8d4a57f4fc).

The simplest way would be to replace these with a query referencing the number of inhabitants, residences, buildings, etc.

```
# number_of_energy_flexibility_p2g_electricity

- min_value_gql = 0.0
- max_value_gql = 0.005959627 * AREA(number_of_inhabitants)
- start_value_gql = "present:V(energy_flexibility_p2g_electricity, number_of_units)"
- unit = #
```

It would still be acceptable to use hard-coded values [for unscalable inputs](https://github.com/quintel/etengine/blob/470256764853fa338a13543e272de324bd46610b/app/models/scenario_scaling.rb#L7-L8). For example:

```
# agriculture_geothermal_share

- min_value = 0
- max_value = 100.0
- start_value_gql = "present:V(agricult[...]ble_heat) * 100)"
- unit = %
```

### Get rid of step values in ETSource inputs

This eliminates the need to scale step values in ETEngine.

1. Steps are not enforced by ETEngine,
2. They are overridden in ETModel with a custom value (`input_elements.step_value`). I believe ETFlex also uses custom values.
[...]

PR #1216, opened 2017-01-24
No more max values

ChaelKruip:
Removes from the inputs:
* max_value
* start_value

The main thing to worry about is if there are slider caps (`max_value_gql`) that are smaller than the previous cap and therefore could change the slider settings of existing scenario's. 

I don't worry very much though because `max_value_gql` was leading in the first place and there are just a few inputs for which there was no `max_value_gql` and I checked that their new max is higher than previously set by the `max_value`. Still, it might be worth a bit of (automated?) testing.

Fixes https://github.com/quintel/etsource/issues/1213


antw:
@ChaelKruip I pushed a small change; apparently ETE won't let you simply specify a number as a `_gql` attribute, but needs to be told which period to use:

* No: `- max_value_gql = 25000`
* Yes: `- max_value_gql = present:25000`

That seems an unnecessary requirement when the value is numeric, but for the sake of simplicity I fixed the three offending inputs and we can look to change ETE later.

Here are the changes from old to new:

* number_of_energy_chp_supercritical_waste_mix_present
  * max 21.325 → 426.497
* number_of_energy_flexibility_p2g_electricity
  * max 100000.0 → 31140.266
* number_of_industry_flexibility_p2g_electricity
  * max 100000.0 → 12436.644
* number_of_energy_chp_supercritical_waste_mix
  * max 21.325 → 426.497
* number_of_energy_heat_network_backup_heater_network_gas
  * max 1000.0 → 286.765
* number_of_energy_power_hydro_mountain
  * max 44.913 → 45.158
* policy_area_green_gas
  * max 5000.0 → 55427.1

Some of the changes are quite large, but if they seem right to you we can merge it.

ChaelKruip:
> Some of the changes are quite large, but if they seem right to you we can merge it.

I think these changes are quite fine. 

@jorisberkhout @AlexanderWirtz do you agree? The maximum number of P2G units is now based on the assumption that you can build enough P2G to convert the total electricity production in the present year, twice.

jorisberkhout:
I think the new max value for the P2G sliders are to tight. Although the exact value is arbitrary, I would suggest to keep them close to what they are at the moment (`100,000`). I have changed the `max_value_gql` for both sliders by increasing the factor `2` by which the total electricity production is multiplied to `6` for `energy_p2g` and `16` for `industry_p2g`. I have pushed these changes to this PR.

ChaelKruip:
@antw can you merge this?

PR #1224, opened 2017-01-26
Local datasets

grdw:
I rebased everything against master. :man_shrugging:  

The reason: https://github.com/quintel/etsource/pull/1225#issuecomment-275662765


Rakefile
```diff
@@ -335,3 +335,25 @@ task :fix_line_endings do
     end
   end
 end
+
+desc <<-DESC
+  Scale a dataset to amount_of_residences
+
+  Arguments:
```
antw: This documentation is a little out-of-date now:

* DATASET → FULL_DATASET
* LOCAL_DATASET → DERIVED_DATASET

It would be nice if the task could do some basic error checking – that values were provided for all three options and the full dataset exists – because the default error isn't going to be very helpful to researchers:

```
rake aborted!
Could not find a fulldataset with the key nil
```

grdw:
Merge #1230 first please @antw before merging this.

PR #1225, opened 2017-01-26
Add specs to validate Initializer Inputs

### LINKED ISSUE #1222: Add specs for local datasets
Validate:

- `DerivedDataset`
  - Validate that the referenced initializer inputs exist
- `InitializerInput`
  - Check if `update_type` is one of `nil`, `factor`, or `%`
  - Check if `query` is applicable and doesn't contain syntax errors

grdw:
~~Dependent on: https://github.com/quintel/atlas/pull/86~~

---

Fixes #1222 
Contains part of #1227

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GIT
   remote: git://github.com/quintel/atlas.git
-  revision: 1a67953fc974177bad9eb194750d4677b4b99880
-  ref: 1a67953
+  revision: e8d4af7b726c7a87ba6d80b25082284e1b23bf3b
+  ref: e8d4af7
   specs:
     atlas (1.0.0)
       activemodel (~> 4.0)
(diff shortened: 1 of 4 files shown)


antw:
Since this PR removes the Ameland dataset which was only just added to "local-datasets", perhaps you might use `rebase -i` to remove b9d55c1 thus negating the need for 02103c2.

Are the new initializer inputs added in cd2dbabe related to the specs?

Gemfile.lock
```diff
@@ -51,8 +51,8 @@ GEM
     equalizer (0.0.11)
     i18n (0.7.0)
     ice_nine (0.11.2)
-    json (1.8.3)
-    minitest (5.9.1)
+    json (1.8.6)
+    minitest (5.10.1)
```
ploh: Is this on purpose? Otherwise we should probably "not risk it".

PR #1226, opened 2017-01-26
Make conditionals in gqueries more robust

ploh:
Rationale: If the values are basically identical, we
still want to have a determinate outcome - even in
case of rounding errors


ploh:
@antw If you agree with the changes, please squash-merge.

ploh:
@antw The semaphore error looks to me as if this is running with ruby 1.9.3 :open_mouth:

antw:
> @antw The semaphore error looks to me as if this is running with ruby 1.9.3 😮

Yup! 😱 

I bumped it to 2.3 and it passed. Merged! :+1:

Issue #1227, opened 2017-01-27
Initializer inputs that don't do anything


grdw:
The migration left with some initializer inputs without a query. This is ofcourse impossible, these are the one's without a `query` or an empty looking `query`:

Migrated: | Originates from:
-------------|-------------------------
[agriculture_burner_network_gas_share.ad](https://github.com/quintel/etsource/pull/1228/files#diff-925b3b2775fbb523b4ff582e46e5daf3) | [agriculture_burner_network_gas_share_present.ad](https://github.com/quintel/etsource/blob/master/inputs/adjust_scaling/agriculture/agriculture_burner_network_gas_share_present.ad)
[households_cooling_airconditioning_electricity_share.ad ](https://github.com/quintel/etsource/pull/1228/files#diff-59ace0e6ce5e1e58a2f5cea02740e0a9) | [households_cooling_airconditioning_electricity_share_present.ad](https://github.com/quintel/etsource/blob/master/inputs/adjust_scaling/households/households_cooling_airconditioning_electricity_share_present.ad)
[households_heat_network_connection_steam_hot_water_share.ad ](https://github.com/quintel/etsource/pull/1228/files#diff-be42d19e4d611753a817ee0e5e61e790) | [households_heat_network_connection_steam_hot_water_share_present.ad](https://github.com/quintel/etsource/blob/master/inputs/adjust_scaling/households/households_heat_network_connection_steam_hot_water_share_present.ad)
[households_lighting_incandescent_electricity_share.ad ](https://github.com/quintel/etsource/pull/1228/files#diff-eafde6b382047e6f0cf75e7f6bf48f9d) | [households_lighting_incandescent_electricity_share_present.ad](https://github.com/quintel/etsource/blob/master/inputs/adjust_scaling/households/households_lighting_incandescent_electricity_share_present.ad)
[households_water_heater_network_gas_share.ad ](https://github.com/quintel/etsource/pull/1228/files#diff-9fcb37ad9c38047f4af589a62aeb3a32) | [households_water_heater_network_gas_share_both.ad](https://github.com/quintel/etsource/blob/master/inputs/adjust_scaling/households/households_water_heater_network_gas_share_both.ad)

---

This [one](https://github.com/quintel/etsource/pull/1228/files#diff-037b4900364e968e16e08e6127cbc297R7) has a non-supported attribute called `dependent_on`. What is it and what does it do?

grdw:
[...]

PR #1228, opened 2017-01-27
Local datasets migrate adjust scaling inputs

grdw:
This is a very basic migration from the `adjust_scaling` inputs to the `initializer_inputs` folder. I made sure all the attributes which are not needed for initializer inputs were removed from the initializer input itself. 

#1227 

--- 

- [x] Removed `min_value`, `min_value_gql`, `max_value`, `max_value_gql`, `start_value_gql`, etc. 
- [x] The `update_period` `%y` is removed from the inputs (so they default to absolute values)
- [x] Removed `present`, `both` from filenames


initializer_inputs/agriculture/agriculture_burner_crude_oil_share.ad
```diff
@@ -0,0 +1,10 @@
+- query =
+    EACH(
+      UPDATE(LINK(agriculture_burner_crude_oil,agriculture_useful_demand_useable_heat), share, DIVIDE(USER_INPUT(),100)),
+      UPDATE(LINK(agriculture_useful_demand_useable_heat,agriculture_burner_network_gas), share,
+        ->{SUM(NEG(SUM(V(EXCLUDE(INPUT_LINKS(V(agriculture_useful_demand_useable_heat)),UPDATE_COLLECTION()), share))), 1)}
+      )
+    )
+- share_group = agri_heat_present
```
ploh: I guess the share group names do not need the `_present` anymore either

initializer_inputs/agriculture/agriculture_burner_network_gas_share.ad
```diff
@@ -0,0 +1,4 @@
+- id = 246
```
ploh: Is this used for anything? Is this supported for init. inputs?

initializer_inputs/industry/industry_aluminium_production.ad
```diff
@@ -0,0 +1,3 @@
+- query = UPDATE_WITH_FACTOR(V(industry_aluminium_production), preset_demand, USER_INPUT())
+- priority = 10
+- update_type = %
```
ploh: If we change the %y to absolute, then maybe also the % to absolute!?
But please talk to @AlexanderWirtz about this. Maybe we do not even need the update_type anymore for init. inputs?
I guess for new datasets/scenarios that would be desirable. For the old ones it might make the migration a lot more complicated though (see #1229).

initializer_inputs/agriculture/agriculture_burner_network_gas_share.ad
grdw: See [this comment](https://github.com/quintel/etsource/issues/1227#issuecomment-276033822)

initializer_inputs/industry/industry_aluminium_production.ad
grdw: The % `update_type` needs to stay because it's used to 'scale' certain industry sectors according to @ChaelKruip

initializer_inputs/agriculture/agriculture_burner_network_gas_share.ad
ploh: That does not really say anything about the `id` though. But I guess it makes my question irrelevant ;-)

grdw: Yes, it's more about the absence of the `query` key ;)

Issue #1223, opened 2017-01-26
Mistakes in transport_car carrier share inputs


ploh:
As discovered in https://github.com/quintel/etengine/issues/904#issuecomment-275199374 while migrating a scenario to a new derived dataset, there are probably some errors in the `transport_car_using_..._share(_present)` inputs:

In [these lines of the present inputs](https://github.com/quintel/etsource/blob/ea7cc5af9be787f76df760b01f32744e77b8f561/inputs/adjust_scaling/transport/transport_car_using_hydrogen_share_present.ad#L8-L12) (the ones from this share group are all similar):
The use of `INPUT_VALUE(transport_car_using_electricity_share)` although we are updating the present graph, i.e. it should be `INPUT_VALUE(transport_car_using_electricity_share_present)` here.

And in [these lines of the future inputs](https://github.com/quintel/etsource/blob/ea7cc5af9be787f76df760b01f32744e77b8f561/inputs/demand/transport/transport_car/transport_car_using_hydrogen_share.ad#L7-L12):
The use of `QUERY_PRESENT(-> { AREA(number_of_cars) })` is probably a mistake since the future number of electric cars should be relative to the future total number of cars and not to the present total number.

AlexanderWirtz:
well spotted!

ChaelKruip:
@ploh what is the action here? Have you tried your own suggestions for fixing this? I don't know what is expected of me here 😄

AlexanderWirtz:
@ChaelKruip I'll venture a guess: you are expected to fix these. 😄  

They look like obvious mistakes, but a closer look might be warranted. @BerendSmits this is relevant to our work, as a fix to this issue will affect the outcome of the Drenthe dataset. Only in cases where `future` and `present` are not equal, though.

ploh:
The action is to verify that my ideas about this are right and implement them :wink:
I did not try myself because I was not sure about my solution (especially regarding the future inputs) and because I did not have the time for it

ChaelKruip:
> The action is to verify that my ideas about this are right and implement them

I still have some questions before I invest a considerable amount of time into this:
* How do I verify these ideas? I have not yet worked with the new scaling locally...
* Is this even relevant for the new scaling methods? Aren't we using "both"-statement only for that? Or is this for purpose of migrating old scaled scenario's? 
* Are there many (any?) scenario's where these queries are used and could we fix these in some other way?

@ploh could you please provide some insight?

ploh:
[...]

PR #1235, opened 2017-02-07
Car input fix

ChaelKruip:
Fixes https://github.com/quintel/etsource/issues/1223

### DIFF
--- a/gqueries/general/costs/costs_of_transport_vehicles.gql
+++ b/gqueries/general/costs/costs_of_transport_vehicles.gql
@@ -2,7 +2,7 @@
 
 - query =
     SUM( 
-      V((transport_car_using_electricity), total_costs_per(:plant)) * DIVIDE(INPUT_VALUE(transport_car_using_electricity_share),100) * QUERY_PRESENT(-> { AREA(number_of_cars) }),
-      V((transport_car_using_hydrogen), total_costs_per(:plant)) * DIVIDE(INPUT_VALUE(transport_car_using_hydrogen_share),100) * QUERY_PRESENT(-> { AREA(number_of_cars) })
+      V((transport_car_using_electricity), total_costs_per(:plant)) * V(LINK(transport_car_using_electricity, transport_useful_demand_car_kms), share) * QUERY_FUTURE(-> { AREA(number_of_cars) }),
+      V((transport_car_using_hydrogen), total_costs_per(:plant)) * V(LINK(transport_car_using_hydrogen, transport_useful_demand_car_kms), share) * QUERY_FUTURE(-> { AREA(number_of_cars) })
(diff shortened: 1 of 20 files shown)


antw:
@ChaelKruip Do you know why these inputs balance the share of gasoline automatically?

```ruby
UPDATE(LINK(transport_useful_demand_car_kms,transport_car_using_gasoline_mix), share,
  ->{SUM(NEG(SUM(V(EXCLUDE(INPUT_LINKS(V(transport_useful_demand_car_kms)),UPDATE_COLLECTION()), share))), 1)})
```

This is pretty non-standard (e.g. heating inputs [only set the share of the link](https://github.com/quintel/etsource/blob/cf29e6472830246907b99ef163c7f73ec391171e/inputs/demand/households/households_heating/households_space_heater_micro_chp_network_gas_share.ad#L1)); the user is expected to provide the entire group of inputs or permit the auto-balancer to make adjustments automatically. This balancing [predates "modern" ETSource+Atlas](https://github.com/quintel/etsource/blame/1ec83d5ee69814a5eb4b616bd5bd7ab07c8415ed/data/inputs/transport/transport_car/transport_car_using_compressed_natural_gas_share.ad) so it might be that we can remove it.

ChaelKruip:
> @ChaelKruip Do you know why these inputs balance the share of gasoline automatically?

No :blushing:
We currently have about 200 inputs which use this type of 'balancing' inside the input-statement. If that can be removed I'd be happy to do that!

antw:
> We currently have about 200 inputs which use this type of 'balancing' 

Right you are! I guess it's not as "non-standard" as I thought. Still, it's strange that this [is used in heating for buildings](https://github.com/quintel/etsource/blob/cf29e6472830246907b99ef163c7f73ec391171e/inputs/demand/buildings/buildings_heating/buildings_space_heater_electricity_share.ad) but not [heating for households](https://github.com/quintel/etsource/blob/cf29e6472830246907b99ef163c7f73ec391171e/inputs/demand/households/households_heating/households_space_heater_electricity_share.ad). I don't see any obvious difference in the graph: both have a useful demand node with input links from each technology. In fact, the network gas heater links are `flexible` which means they'll be automatically "balanced" anyway, even without the manual balancing in the input statements.

Would you mind if I experimented with this briefly before merging? It would be a small win if it turns out we can eliminate the manual balancing in the input statement.

ChaelKruip:
> Would you mind if I experimented with this briefly before merging?

Not at all!

>  It would be a small win if it turns out we can eliminate the manual balancing in the input statement.

👍

antw:
Superseded by #1237. Removing INPUT_VALUE helped fix #1236; thanks!

PR #1238, opened 2017-02-07
Local dataset for Drenthe

BerendSmits:
This branch includes the newly made local dataset for Drenthe. As per the provinces request, i've taken out all the national mix energy production and only added what is present within the province.


datasets/drenthe/time_curves/energy_distribution_woody_biomass_time_curve.csv
```diff
@@ -0,0 +1,83 @@
+year,max_demand
```
antw: These time curve files don't look right to me. It's probably an issue with the scaling script and not your fault; I will look into it in the morning.

antw: I fixed the broken time curves, it was indeed an issue with the scaling script. You can grab the fixed curves by running `git pull`.

PR #1246, opened 2017-02-16
changed name of ad file from npd to noorderplantsoen. Also changed th…

AlexanderWirtz:
…e name in the ad file and of the folder. I HAVE NOT CHECKED IF THIS WORKS LOCALLY AS MY SERVER STILL DOESN'T WORK.


grdw:
I've just created a scenario with this dataset locally and everything looks fine (name of the dataset is "Noorderplantsoen" and nothing crucial breaks as I'd expect).

PR #1254, opened 2017-02-22
Remove 'number_of_cars' from car kms

### LINKED ISSUE #1253: Number of cars in local datasets
The number of cars in local datasets does not seem to follow the .ad file. For instance, in Drenthe the .ad file lists number of cars as: 
screen shot 2017-02-22 at 13 59 26

But when you open a scenario and query the number of cars, it gives: 

screen shot 2017-02-22 at 14 00 50

Where the queried number comes from is unknown, it is not the total number of cars in NL (its about 500x larger). For Paddepoel, its 1,470,126.568, for Noorderplantsoen its 8,876,542.882 (more than NL).

grdw:
I turned this initializer input to a WYSIWYG (what you see is what you get) initializer. Please correct me if this is wrong. 

Fixes #1253

### DIFF
--- a/initializer_inputs/transport/transport_car/transport_useful_demand_car_kms.ad
+++ b/initializer_inputs/transport/transport_car/transport_useful_demand_car_kms.ad
@@ -1,13 +1,3 @@
-- query =
-    EACH(
-      UPDATE(V(transport_useful_demand_car_kms), preset_demand, USER_INPUT()),
-      UPDATE(AREA(), number_of_cars, USER_INPUT()),
-      UPDATE(V(transport_car_using_compressed_natural_gas), number_of_units, USER_INPUT()),
-      UPDATE(V(transport_car_using_diesel_mix), number_of_units, USER_INPUT()),
-      UPDATE(V(transport_car_using_electricity), number_of_units, USER_INPUT()),
-      UPDATE(V(transport_car_using_gasoline_mix), number_of_units, USER_INPUT()),
-      UPDATE(V(transport_car_using_hydrogen), number_of_units, USER_INPUT()),
-      UPDATE(V(transport_car_using_lpg), number_of_units, USER_INPUT())
-    )
+- query = UPDATE(V(transport_useful_demand_car_kms), preset_demand, USER_INPUT())
 - priority = 10
 


grdw:
I have no idea who to assign this to in @ChaelKruip's absence. So if any of you have time, please take a look. Thanks!

antw:
> Please correct me if this is wrong.

I'm not 100% sure myself, but I think it is probably incorrect to remove the adjustment to `number_of_units` since ETEngine is not capable of calculating this attribute for cars; see https://github.com/quintel/etsource/issues/1223#issuecomment-277316959.

As far as I know, the input works by pretending that more (or fewer) cars have been purchased in the region and therefore the amount of energy used has increases. By removing the adjustment to `number_of_units`, the behaviour changes to having the same number of cars but each using more energy.

Note that both [the scaling](https://github.com/quintel/etsource/blob/9807a5fa380621197ddf41aa1027aa90ea56f5fe/inputs/adjust_scaling/transport/transport_car/transport_useful_demand_car_kms_both.ad) and [ordinary input](https://github.com/quintel/etsource/blob/9807a5fa380621197ddf41aa1027aa90ea56f5fe/inputs/demand/transport/transport_car/transport_useful_demand_car_kms.ad) also do this.

To be honest I don't know enough about how transport is modelled to offer a better solution to #1253. We could remove the input entirely, but I'm guessing that increasing the number of cars in the area `.ad` file won't change the demand of the node.

---

Pinging @jorisberkhout.

dennisschoenmakers:
@grdw: What's the reason and motivation for this change?

grdw:
> @grdw: What's the reason and motivation for this change?

See: https://github.com/quintel/etsource/issues/1253

grdw:
> I'm not 100% sure myself, but I think it is probably incorrect to remove the adjustment to number_of_units

Probably, but the way it's currently used with initializer inputs is completely wrong aswell.  If you look at the [old adjust scaling input](https://github.com/quintel/etsource/blob/9807a5fa380621197ddf41aa1027aa90ea56f5fe/inputs/adjust_scaling/transport/transport_car/transport_useful_demand_car_kms_both.ad), you see that the `update_type` is `%y`. 

The initializer input doesn't know what that `update_type` is, so it just uses a `#` unit, so therin lies the problem. The input's name is `transport_useful_demand_car_kms` people tend to fill in a demand. That demand is than used as a 'number_of_units' which is as far as I know not correct. 

So maybe a solution would be to split these two inputs up? One for `number_of_units_of_cars`? And another for the `demands`? Than again I'm no energy expert.

grdw:
What's the status of this issue @ChaelKruip @DorinevanderVlies - I heared from @BerendSmits that you were taking some steps in fixing this issue? 

I need this to be resolved for https://github.com/quintel/etlocal/issues/46 . 😄

DorinevanderVlies:
@antw can I close this issue as it is overwritten by [ETSource commit 1283](https://github.com/quintel/etsource/pull/1283/commits)

PR #1259, opened 2017-03-15
Added cost of flex to the queries for chart 48

ChaelKruip:


ChaelKruip:
This fixes https://github.com/quintel/etmodel/issues/2357

PR #1265, opened 2017-03-29
Changed wind on land potential 

BerendSmits:
Changed wind on land potential of province from wind concentration area, to no-barriers approach


AlexanderWirtz:
Does this also change the maximization of the slider value for installing onshore wind?

BerendSmits:
> Does this also change the maximization of the slider value for installing onshore wind?

Yes it does

PR #1269, opened 2017-04-13
Solar plants maximum amount

BerendSmits:
Changed the max for the amount of solar plants from depending on the total electricity produced, to depending on the total land available, as referenced in https://github.com/quintel/etmodel/issues/2367


jorisberkhout:
Nice work @BerendSmits !

PR #1271, opened 2017-04-21
Adding gquery for number of residences

grdw:


gqueries/modules/etmoses/number_of_residences.gql
```diff
@@ -0,0 +1,2 @@
+- query = AREA(number_of_new_residences)
```
antw: Why `number_of_new_residences` and not `number_of_residences`? It would be a significant change in behaviour for ETMoses to only include new residences.

gqueries/modules/etmoses/number_of_residences.gql
```diff
@@ -0,0 +1,2 @@
...
+- unit = number
```
antw:
```diff
  | - query = AREA(number_of_new_residences)
- | - unit = number
+ | - unit = #
```

(I do prefer `number` and there are some inputs which use it, but at the moment `#` is the unit for queries which represent a raw numbers).

grdw:
@antw Turns out that the `number_of_residences.gql` is already present in the list of gqueries. Also, it was already used in the [`Export` class of ETMoses](https://github.com/quintel/etmoses/blob/master/app/models/export.rb#L92).

PR #1273, opened 2017-05-16
Network cost queries

antw:
This updates the network cost queries to include the explicitly-modelled technologies (EVs, household space heating, household hot water).

@ChaelKruip: **I've not merged this myself because I'd appreciate a quick sanity check from yourself.** I created a scenario using my local ETE/ETS and compared it with a live scenario. The only change from default settings was to increase household electric heaters from the default 1.2%, up to 100%.

investment2

There are two things to note:

1. The cost of additional infrastructure is *lower* due to the use of a "real" load, instead of taking the converter's peak load and [multiplying it](https://github.com/quintel/etsource/blob/234f86f9fd7fcabdbd2f7d59d0576c2c5486059b/nodes/households/households_space_heater_electricity.converter.ad#L19-L20) by a constant. The actual "peak load" of the explicitly modelled technologies [is not *hugely* different](https://gist.github.com/antw/1bf5386099f19bbfd7380da760a21369), so I think the blame for the different costs is the `simult_*` constants.

2. There is a large rise in electricity demand due to 100% of households having electric heaters. 70% of houses are old and poorly insulated, so the demand curve becomes very spikey. I think this makes sense, but it is quite a significant departure from the previous model with many more blackout hours (1249 on production, 2616 with the dynamic curve). Fully insulating all homes has [a dramatic effect on peak load and costs](https://cloud.githubusercontent.com/assets/4383/26110577/a87f20f4-3a4a-11e7-867b-057232af4def.png).

If this looks good to you, please merge at your convenience. If not, let me know if there are any other charts or data I can get for you.

### DIFF
--- a/gqueries/modules/network/network_impact_calculations/hv_mv_trafo/network_hv_mv_trafo_peak_load_sd.gql
+++ b/gqueries/modules/network/network_impact_calculations/hv_mv_trafo/network_hv_mv_trafo_peak_load_sd.gql
@@ -7,7 +7,7 @@
         SUM(
             SUM(
                 QUERY_DELTA( -> {SUM(V(Q(network_lv_net_converters),"simult_sd*peak_load_in_mw")) }),
-                GRAPH(peak_load_of_explicitly_modelled_lv_technologies_sd)
+                V(GRAPH(), peak_load_of_explicitly_modelled_lv_technologies(sd))
             ) * V(energy_power_lv_network_electricity,simult_sd)
                 * V(energy_power_transformer_lv_mv_electricity,simult_sd)
                 * V(energy_power_mv_distribution_network_electricity, simult_sd)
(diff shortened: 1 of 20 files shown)


ChaelKruip:
Thanks @antw for this great overview! 

> The cost of additional infrastructure is lower due to the use of a "real" load, instead of taking the converter's peak load and multiplying it by a constant.

I am suspicious of this. I'd expect the peaks to be much higher in the new modelling (as the visuals show). The simult multipliers should be 1.0 or lower to make sense to me (which unfortunately is not true for the electric heater). So, the fault is probably in the old implementation.

My guess is that the effect of this technology has been artificially boosted in the past and the current modelling is better. @AlexanderWirtz @jorisberkhout what do you think?

AlexanderWirtz:
> My guess is that the effect of this technology has been artificially boosted in the past and the current modelling is better. @AlexanderWirtz @jorisberkhout what do you think?

My hunch is that you are right. Quite excited about this. Let's not forget to run it by Berenschot once things are semi-stable, as they have a good feeling for these things (i.e. Bert does).

PR #1274, opened 2017-05-19
Added missing area attributes for EV profiles to local datasets

jorisberkhout:
Fixes: https://github.com/quintel/etengine/issues/923


grdw:
Fixes: https://github.com/quintel/etengine/issues/923

PR #1276, opened 2017-05-22
Add task to refresh a graph from a local dataset

grdw:
Part of: https://github.com/quintel/atlas/pull/98


grdw:
Still not 100% convinced though that `refresh_graph` is the correct name for this task. Maybe `update_graph`? Do you have any strong opinions on this matter? @antw

PR #1283, opened 2017-06-07
Update car converters to inherit number of units from parent

antw:
This is part of quintel/etmodel#2360 and changes ETSource so that the individual car nodes (transport car using gasoline, transport car using electricity, etc) have a correct `number_of_units` attribute. This attribute will be recomputed if you increase the demand of "car kms" or change the share of car technologies.

This is achieved by:

1. Setting a `number_of_units` attribute on "transport_useful_demand_for_car_kms",
2. Assigning the individual car nodes to the `inheritable_nou` group which will trigger the behaviour in ETEngine (only once quintel/etengine#926 is merged).

It is possible that we could decouple *demand* for cars from *number* of cars by changing some queries; this would permit us to add a separate slider for the number of vehicles.

### Caveats:

1. The number of units of cars [is read from the area/dataset `number_of_cars` attribute](https://github.com/quintel/etsource/blob/cars-nou/nodes/transport/transport_useful_demand_car_kms.demand.ad#L6). This is not ideal since it means we now have two places where the number of cars is available in ETEngine (the area data and the `number_of_units` attribute on "transport_useful_demand_for_car_kms".
  
   Queries have been adjusted to update both attributes, but ideally we might remove the area attribute and use a CSV to store the value for cars and trucks.

2. You may not set the number of units for the individual car types; you must set the total number of cars by changing `number_of_units` on the "transport_useful_demand_for_car_kms" node and adjust the shares of each technology as desired. ETEngine will then work out the number of units for each technology.

3. I have not tested the behaviour in local datasets and therefore cannot *yet* answer [these questions](https://github.com/quintel/etmodel/issues/2360#issuecomment-306500961). This testing is next on my list and I will update this issue once completed.

### Examples:

Here are a couple of example of how this change affects the number of units attribute on the various car nodes both before this change ("old number of units") and after ("new number of units").

#### NL, 2050, default settings:

| Node | Old number of units | New number of units |
| :---: | ---: | --: |
| `transport_useful_demand_car_kms` | 0 | 7,916 K	|
| `transport_car_using_compressed_natural_gas` | 0 | 0 |
| `transport_car_using_diesel_mix` | 0 | 2,147 K |
| `transport_car_using_electricity` | 11 K | 11 K |
| `transport_car_using_gasoline_mix` | 0 | 5,345 K |
| `transport_car_using_hydrogen` | 0 | 0 |
| `transport_car_using_lpg` | 0 | 406 K |

#### NL, 2050, Demand increased by 2% per year

| Node | Old number of units | New number of units |
| :---: | ---: | --: |
| `transport_useful_demand_car_kms` | 0 | 16,469 K	|
| `transport_car_using_compressed_natural_gas` | 0 | 0 |
| `transport_car_using_diesel_mix` | 0 | 4,468 K |
| `transport_car_using_electricity` | 23 K | 23 K |
| `transport_car_using_gasoline_mix` | 0 | 11,120 K |
| `transport_car_using_hydrogen` | 0 | 0 |
| `transport_car_using_lpg` | 0 | 845 K |

#### NL, 2050, EV share = 100%

| Node | Old number of units | New number of units |
| :---: | ---: | --: |
| `transport_useful_demand_car_kms` | 0 | 7,916 K	|
| `transport_car_using_compressed_natural_gas` | 0 | 0 |
| `transport_car_using_diesel_mix` | 0 | 0 |
| `transport_car_using_electricity` | **11 K (!!!)** | 7,916 K |
| `transport_car_using_gasoline_mix` | 0 | 0 |
| `transport_car_using_hydrogen` | 0 | 0 |
| `transport_car_using_lpg` | 0 | 0 |

---

FYI @ChaelKruip @DorinevanderVlies

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gem 'rake'
 
 group :development, :test do
-  gem 'atlas',    ref: '6eb0ad5', github: 'quintel/atlas'
+  gem 'atlas',    ref: 'cec2d64', github: 'quintel/atlas'
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gem 'rake'
 
 group :development, :test do
-  gem 'atlas',    ref: 'cec2d64', github: 'quintel/atlas'
+  gem 'atlas',    ref: '1ab0482', github: 'quintel/atlas'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GIT
   remote: git://github.com/quintel/atlas.git
-  revision: 6eb0ad5f9a721ee31c4afa36b97a357d3d42ae9d
-  ref: 6eb0ad5
+  revision: cec2d6452ed6b5d3ea7386d3450561c9e33b1109
+  ref: cec2d64
   specs:
     atlas (1.0.0)
       activemodel (>= 4.0)
@@ -47,7 +47,7 @@ GEM
       thread_safe (~> 0.3, >= 0.3.1)
     diff-lcs (1.2.4)
     equalizer (0.0.11)
-    i18n (0.8.1)
+    i18n (0.8.4)
     ice_nine (0.11.2)
     minitest (5.10.2)
     rake (10.1.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GIT
   remote: git://github.com/quintel/atlas.git
-  revision: cec2d6452ed6b5d3ea7386d3450561c9e33b1109
-  ref: cec2d64
+  revision: 1ab0482f69b026b64f02ea9b7afa81dd3b878300
+  ref: 1ab0482
   specs:
     atlas (1.0.0)
       activemodel (>= 4.0)
--- a/Rakefile
+++ b/Rakefile
@@ -375,20 +375,34 @@ DESC
 task :refresh_graph do
   require 'bundler'
 
-  raise ArgumentError, 'missing DERIVED_DATASET argument' unless ENV['DERIVED_DATASET']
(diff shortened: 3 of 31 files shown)


antw:
I have pushed my (final?) changes based on a discussion with @DorinevanderVlies earlier today.

* The initializer input for car demand [will now *only* set the demand](https://github.com/quintel/etsource/pull/1283/files#diff-a535ef9585e1f37c0dffbcc7e499dd67), and will not adjust the number of cars (replaces #1254).

* [A separate initializer input has been added to set the number of cars](https://github.com/quintel/etsource/pull/1283/files#diff-ed9406b52b9d282079ff44a6a16bd44a). The existing local datasets have been updated so as to set the correct value for this initializer input.

* The front-end slider will [still set demand *and* number of cars](https://github.com/quintel/etsource/pull/1283/files#diff-235cad95e7b62ef1316c79561646a671). We can separate these into separate sliders if you want.

* The Atlas graph scaler will set the correctly scaled number of cars as a default value; therefore if you don't set a value using the input, you will still get a value which makes sense for the size of the region.

From my testing I can confirm that electric vehicles now appear to behave correctly as a flexibility option in derived datasets:

* if there are no electric cars, EVs will not be used for flexibility.
* if you set a large share of EVs, the number of units is correctly adjusted (previously the number of vehicles was massively inflated in derived datasets; Groningen had 3.92B EVs!).

@DorinevanderVlies @ChaelKruip: If you wish, I can merge this and the ETEngine pull request immediately so that testing may be done on beta. If you prefer to test locally first, that's also fine.

DorinevanderVlies:
Nice work @antw . I'm looking forward to test these improvements. 

My preference is to do this on beta as it will be faster. I also suppose that the export from ETM and ETM-SA will work better from beta (I doubt I can make it work at all locally) @ChaelKruip do you agree with this?

ChaelKruip:
Very nice work indeed! It would have my strong preference to deploy and test this next week.

Issue #1266, opened 2017-03-31
Little bit of extra energy in import/export waterfall


ChaelKruip:
The waterfall adds up to 3.19 in my local scenario (essentially a start-scenario for NL 2030) but if you run 
```
SUM(
        V(G(energy_import),demand).sum,
        NEG(V(G(energy_export),demand).sum)
  )
```
it returns 3.17

![screen shot 2017-03-31 at 16 26 47](https://cloud.githubusercontent.com/assets/1303760/24554770/092bc71c-162f-11e7-8bd0-ed17e25de032.png)

jorisberkhout:
Removing my assignment.

BerendSmits:
I think this is caused by double counting LNG imports/exports. 

The natural gas column in the waterfall uses the following query: 

gas_in_future_energy_imports.gql:
>- query =
    future:
    DIVIDE(
        V(G(energy_import),"output_of_natural_gas + output_of_lng").sum - 
        V(G(energy_export),"input_of_natural_gas + input_of_lng ").sum
    ,BILLIONS)

And the LNG column in the waterfall uses: 

lng_in_future_energy_imports.gql
>- query = future:DIVIDE(V(energy_import_lng, output_of_lng),BILLIONS)

There would then be two ways of solving this problem:
- Removing the LNG column
- Removing LNG from the natural gas column

@ChaelKruip do you agree with what the problem is, and which solution would you prefer? 

NB the solution needs to be implemented for both present and future

ChaelKruip:
> do you agree with what the problem is, and which solution would you prefer?

Yes and option 2!

BerendSmits:
Fixed by https://github.com/quintel/etsource/pull/1286

PR #1286, opened 2017-06-14
Import export waterfall

BerendSmits:
As described in issue #1266, I removed LNG from the import/export waterfall. This was done because there already was a separate LNG column, causing LNG to be double counted. I also added network gas to the calculation of future imports and exports, to make present and future match. That shouldn't have an impact on outcome.


ChaelKruip:
Fixes  #1266.

PR #1291, opened 2017-07-03
Adding dataset for Friesland

jorisberkhout:


ChaelKruip:
I think it looks good but there is one thing that I can imagine people are going to find frustrating: the CO2-emissions are 7% up in a blank scenario:

This is due (I think) to the fact that in the start-year more imported electricity is used:

I don't know what to do about this... Perhaps we are underestimating full load hours of CHPs in the present year?

jorisberkhout:
> I don't know what to do about this... Perhaps we are underestimating full load hours of CHPs in the present year?

Well spotted, @ChaelKruip . This has to do with the FLH of the only dispatchable plant in Friesland (a CCGT). In the start year it has only 2000 FLH, meaning that more electricity is imported (at lower CO2 emissions per kWh); in the end year, this plant runs almost 7800 FLH to reduce the need for imported electricity, but given the higher CO2 emissions of the CCGT, this increases the overall CO2 emissions in the scenario. I don't think we can do anything about this, as this is just the way the merit order module works.

PR #1293, opened 2017-07-07
Subtracted the start-year impact of heating and EV from the ENTSO-E c…

ChaelKruip:
…urve. This results in the measured curve for the start-year.

BETA:

LOCAL:

I was hoping this would solve the issue with network costs in the start-year resulting in this spec failing:

```
1) Network costs In a start scenario no infrastructure investments should be needed
     Failure/Error: expect(@scenario.lv_mv_transformer_in_additional_infrastructure_investments.value).to be == 0.0
       expected: == 0.0
            got:    5367617.713900197
     # ./spec/network/network_infrastructure_investments_spec.rb:12:in `block (3 levels) in <top (required)>'
```

Strangely though, the investment calculated by the mechanical turk is the same between BETA and my LOCAL machine. 

Otherwise, this PR is still a useful improvement IMO.


antw:
> Strangely though, the investment calculated by the mechanical turk is the same between BETA and my LOCAL machine.

This is because the peak load queries use only the explicitly-modelled technologies in Merit, and `simult_*` and `peak_load_in_mw` attributes for everything else. For example:

```ruby
MAX(
  SUM(
    # Attributes...
    QUERY_DELTA( -> {SUM(V(Q(network_lv_net_converters),"simult_sd*peak_load_in_mw")) }),
    # Explicitly modelled technologies load curve...
    V(GRAPH(), peak_load_of_explicitly_modelled_lv_technologies(sd))
  ) * V(energy_power_lv_network_electricity,simult_sd),
  0.0
)
```

PR #1303, opened 2017-08-07
Initializer input change

BerendSmits:
As discussed with @jorisberkhout , i've assigned both @ChaelKruip and @jorisberkhout. 

I made some changes to the set of initializer inputs, mostly removing redundant ones:

 - Households cooling: there are UD inputs for cooling in households for old and new houses separately, and there is also one which sets both (households_useful_demand_for_cooling). The latter has been removed

- Households heating: there are UD inputs for heating in households for old and new houses separately, and there is also one which sets both (households_useful_demand_heat_per_person). The latter has been removed. 

- The building sector has an initializer input for 'buildings_useful_demand_electricity' which updates the UD's for appliances and light. Since seperate inputs for these demands already exist (and these demands should definitely not be set to the exact same value), I removed it.


BerendSmits:
> I agree with everything except the splitting of the solar heater. As far as I understand it, this initialiser actually sets the shares which represent the "number of houses with a solar heater". In this case, I think it is wise to keep those shares linked. It is the same technology in the end. In other words: the solar heater might be the only combi-technology which is modelled in the graph to truly fulfil a combined purpose. Let me know if you want to discuss this point :)

I agree that for most cases, this number should be the same. I had this discussion with @jorisberkhout this morning, and we decided to split up the initializer input for multiple reasons:

1. Wanting our set to be MECE, meaning the demand for each node should be adjustable indendent of all others. 
2. Leaving the possibility open that people only use their solar thermal for one application (for instance in warmer countries where people only use it for hot water but never for space heating)
3. The percentage of UD which is supplied through solar thermal for a house that has solar thermal panels, is hardcoded into the ETM, and cannot be influenced with initializer inputs (15% for space heating and 50% for hot water). If we want to influence the amount of energy running through these nodes separately, this is a way to do it. I agree that this is not the cleanest way to influence the demand delivered by solar thermal. 

For argument 1, I would say we can accept violating MECEness here, if we see this value as strictly a percentage of households with this technology installed. We could allow (new) initializer inputs to work on the percentage of UD delivered through solar thermal in houses with solar thermal panels installed (the earlier mentioned 15 and 50%), solving number 3. As for 2, these cases might be rare, and we can solve them by setting the percentage of UD delivered through solar thermal in a household equipped with panels, to 0%. 

If we were to do that, and keep the current initializer input as it was, I would like to change the name, since now from the name it seems to only affect water heating, while it does more. 

What do you think @jorisberkhout? Should we make new ones for the percentage of useful heat in hot water delivered by solar thermal panel (if household is equipped)?

BerendSmits:
As we decided in our discussion earlier, i'll keep the original initializer input for percentage of houses that have solar thermal (with perhaps a new name), and make new initializer input for the percentage of demand which is filled by solar thermal in a house that has it. 

I'll do this in a separate branch (it will also require small changes in the local datasets). @grdw and I removed the solar thermal commit from this one, @jorisberkhout can you review the other changes so we can merge it?

PR #1304, opened 2017-08-08
Initializer input for hydrogen trucks

BerendSmits:
I added an initializer input for the hydrogen trucks, we did not yet have one


jorisberkhout:
🎉

PR #1314, opened 2017-08-23
As discussed with the province, I removed the UD for aviation

BerendSmits:
I removed the UD for aviation from the Flevoland dataset, as we discussed with the province.

### DIFF
--- a/datasets/flevoland/flevoland.derived.ad
+++ b/datasets/flevoland/flevoland.derived.ad
@@ -295,7 +295,7 @@
 - init.transport_useful_demand_ship_kms = 1355794.2
 
 
-- init.transport_useful_demand_planes = 674695000
+- init.transport_useful_demand_planes = 0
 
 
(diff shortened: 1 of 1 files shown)


grdw:
https://semaphoreci.com/quintel/etsource/branches/flevoland_airport2 this one actually succeeds but Semaphore is not recognizing it as though it is. 🤔 Rebuilding also doesn't work. Do you might know what might be causing this @antw ?

antw:
Semaphore was switched from @dennisschoenmakers' personal account to a Quintel company account. Perhaps that's broken the integration with GitHub...

antw:
I've reauthorised Semaphore but the notifications still aren't reaching GitHub for some reason. I'll take a more thorough look at what's going wrong later when I have a bit more time to spare.

*For now* I've set it that ETSource PRs may be merged without passing Semaphore.

PR #1318, opened 2017-08-24
Fixed final demand queries for household heat and hot water

antw:
Updates the final demand queries used in the "Final energy demand for heating in residences" and "Final energy demand for hot water in residences" charts.

I changed the `*_converters_in_households` converter groups, removed the deficit node (so that its heat is never accidentally considered to be part of the demand), and then changed all of the chart queries to reference the group instead of hard-coding the node name(s).

Note that there is a bug with ambient heat and ambient cold being wrongly reported as zero (visible in the charts below). This is not related to the gqueries, and I'm already looking into it.

screen shot 2017-08-24 at 10 57 43

screen shot 2017-08-24 at 10 57 52


DorinevanderVlies:
nice fix @antw

ChaelKruip:
👍

PR #1322, opened 2017-08-28
Space heating HPs are aliases of the hot water HPs

antw:
Sets the electric- and hybrid- heat pumps for space heating to be aliases of their hot water equivalents. This ensures that the capacity of the heat pump is not considered twice.

screen-shot-2017-08-28-at-17 21 45

Currently the following heat producers act this way:

* `households_space_heater_combined_network_gas` → `households_water_heater_combined_network_gas`

* `households_space_heater_heatpump_air_water_electricity` → `households_water_heater_heatpump_air_water_electricity`

* `households_space_heater_heatpump_ground_water_electricity` → `households_water_heater_heatpump_ground_water_electricity`

* `households_space_heater_hybrid_heatpump_air_water_electricity` → `households_water_heater_hybrid_heatpump_air_water_electricity`

@ChaelKruip @DorinevanderVlies Do any others *also* need to behave this way?

### DIFF
--- a/nodes/households/households_space_heater_district_heating_steam_hot_water.converter.ad
+++ b/nodes/households/households_space_heater_district_heating_steam_hot_water.converter.ad
@@ -2,6 +2,7 @@
 - energy_balance_group = technologies
 - groups = [cost_traditional_heat, heat_production, demand_driven, etmoses, application_group, aggregator_producer]
 - output.useable_heat = 1.0
+- fever.alias_of = households_water_heater_district_heating_steam_hot_water
 - fever.defer_for = 4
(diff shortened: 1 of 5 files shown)


ChaelKruip:
> @ChaelKruip @DorinevanderVlies Do any others also need to behave this way?

I think it goes for all combined technologies in general. So, also the heat network and micro CHP (`households_space_heater_micro_chp_network_gas`) would qualify. I'm not sure how difficult these two are though to add...

antw:
> So, also the heat network and micro CHP would qualify

Done in de52944.

PR #1325, opened 2017-08-30
added inputs for costs storage volume

DorinevanderVlies:


inputs/costs/investment_costs_households_storage_space_heating.ad
```diff
@@ -0,0 +1,19 @@
...
+      cost_per_mwh,
+      DIVIDE(USER_INPUT(), 0.001)
+    )
+- priority = 0
+- max_value = 500.0
+- min_value = 0.0
+- start_value_gql = present:V(households_space_heater_heatpump_air_water_electricity,"storage.cost_per_mwh")*1000
+- step_value = 0.5
+- factor = 1.0
+- unit = euro_per_MWh
```
antw: Should be `euro/kWh`, since that's the value the user actually sees.

inputs/costs/investment_costs_households_storage_space_heating.ad
```diff
@@ -0,0 +1,19 @@
+- query =
+    UPDATE(
+      V(
+        households_space_heater_heatpump_air_water_electricity,
+        households_space_heater_heatpump_ground_water_electricity,
+        households_space_heater_hybrid_heatpump_air_water_electricity,
+        storage
+      ),
+      cost_per_mwh,
+      DIVIDE(USER_INPUT(), 0.001)
```
antw: Did you mean for this to be `DIVIDE(USER_INPUT(), 1000)` to convert from euro/kWh (what the user inputs) to euro/MWh (what ETEngine uses)?

DorinevanderVlies: yes indeed. 1 euro/kWh = 1000 euro/mWh right?
So that is time 1000, or divide by 0.001.

Or should I not do this conversion here?

inputs/costs/investment_costs_households_storage_space_heating.ad
DorinevanderVlies: ok

inputs/costs/investment_costs_households_storage_space_heating.ad
antw: > So that is times 1000, or divide by 0.001.

Sorry, you're completely right. What was I thinking... 🙄

It might be clearer to multiply (`USER_INPUT() * 1000` or `PRODUCT(USER_INPUT(), 1000`), but I don't have a strong opinion. It's fine as it is if you prefer using `DIVIDE` here.

inputs/costs/investment_costs_households_storage_space_heating.ad
DorinevanderVlies: On second thoughts I am not sure. 

Firstly the unit "euro_per_kwh" does not exist yet (am I correct in that?)
Secondly, this seems to be working fine (I checked setting 1 euro/kWh and a 2 kWh buffer, total initial investment costs of the technology increased with 2 euro)

Because the user input is multiplied by 1000 the user actually sets euro/mwh, not euro/kwh.

@antw is my reasoning correct?

DorinevanderVlies:
> Do the heat pump nodes need a default cost, or is are the costs zero unless the user specifies otherwise?

These will be default 0 for the time being. Adjustment may come after extra information is available (I am working on this)

inputs/costs/investment_costs_households_storage_space_heating.ad
antw: > Because the user input is multiplied by 1000 the user actually sets euro/mwh, not euro/kwh.

Indeed. The value the user sees is euro/kWh, but the value set in the graph by the input is euro/MWh.

> Firstly the unit "euro_per_kwh" does not exist yet (am I correct in that?)

`euro/kWh` (not `euro_per_kwh`) does exist already and will not be scaled in local scenarios.

PR #1333, opened 2017-09-05
Change buffer and  costs and buffer size and added sliders for buffer size

DorinevanderVlies:
Updated buffer costs and buffer size according to information from DHPA


nodes/households/households_space_heater_heatpump_air_water_electricity.converter.ad
```diff
@@ -15,8 +15,8 @@
 - fever.efficiency_based_on = electricity
 - fever.group = space_heating
 - fever.type = producer
-- storage.volume = 0.0017
-- storage.cost_per_mwh = 0.0
+- storage.volume = 0.0
```
ChaelKruip: Is this the default / initial value? Why is it zero now?

DorinevanderVlies: Yes, it is 0 for all space heating buffers.

The hot water boiler vessel is used as storage device to protect the compressor. Peter Wagener from DHPA explained today that most HP's use the boiler vessel (which is 5 kWh in our default setting) to protect the compressor. There is no need to have both a hot water boiler and a space heating buffer.

A HHP does not need compressor protection as it is "self-modulling". So, it has 0 buffers by default. That is also how for example the Techneco Elga is generally used.

Issue #1339, opened 2017-09-11
The problem with FIRST(LINKS(OUTPUT_SLOTS(V(<some node>),not_defined))


grdw:
While cleaning up the initializer inputs and specifying which node attributes and edge attributes to set I stumble upon this magic:

```
FIRST(LINKS(OUTPUT_SLOTS(V(<some node>),not_defined))
```

This query looks for all the output slots of `<some node>`, selects all the edges (LINKS) and picks the first edge (FIRST). This is pretty magical. If I were to alter the graph at `<some node>` and add a new node and edge, this query will not function as intended and might cause unforseen bugs. What I would suggest is to turn it into something explicit:

```
LINK(<some node>, <first node>)
```

This is logical, involves no magic, is exactly as it is in other inputs.

Files that have this magic:

- [x] inputs/adjust_scaling/transport/transport_train/transport_train_using_coal_share_present.ad
- [x] inputs/adjust_scaling/transport/transport_train/transport_train_using_diesel_share_present.ad
- [x] inputs/adjust_scaling/transport/transport_train/transport_train_using_electricity_share_present.ad
- [x] inputs/demand/buildings/buildings_misc/buildings_lighting_savings_from_daylight_control_light.ad
- [x] inputs/demand/buildings/buildings_misc/buildings_lighting_savings_from_motion_detection_light.ad
- [x] inputs/demand/transport/transport_plane/transport_plane_using_kerosene_share.ad
- [x] inputs/demand/transport/transport_plane/transport_plane_using_bio_ethanol_share.ad
- [x] inputs/demand/transport/transport_plane/transport_plane_using_gasoline_share.ad
- [x] inputs/demand/transport/transport_train/transport_train_using_coal_share.ad
- [x] inputs/demand/transport/transport_train/transport_train_using_diesel_share.ad
- [x] inputs/demand/transport/transport_train/transport_train_using_electricity_share.ad

grdw:
Re-assign at will

BerendSmits:
@AlexanderWirtz I don't think we use adjust_scaling anymore right?  Is it maybe time to throw those away? Or should we keep them for the old Gelderland dataset?

AlexanderWirtz:
@BerendSmits I am not sure they are not used. How would we find out? Query the database?

ChaelKruip:
Assigning @grdw. Let me know if you need help!

BerendSmits:
[...]

PR #1343, opened 2017-09-12
Refactor the selection of a LINK to something straight-forward

grdw:
Uses of 'FIRST(LINKS(OUTPUT_SLOTS(V(<some node>),not_defined))' are way to fragile.

Fixes: #1339


grdw:
Tested by checking my local machine against beta. Also I fetched the correct links by running the 'dynamic' selection on the `beta` version of etengine

PR #1360, opened 2017-09-26
Remove the yellow color

grdw:
Yellow color on a white background should be avoided in the graph.


grdw:
@ChaelKruip re-assign at will

PR #1385, opened 2017-10-17
added several nodes to the production_parameters-export

DorinevanderVlies:
Closes https://github.com/quintel/etmodel/issues/2512#issuecomment-337206630


jorisberkhout:
@ChaelKruip , would you mind looking at this? I do not oversee the potential fallout of adding these converters to the respective groups (`cost_other` and `electricity_production`).

DorinevanderVlies:
> Why not add it to cost_other as well?

See commit

PR #1388, opened 2017-10-24
Updated heat production queries to actually compute heat (and cold) production

MichielDenHaan:
Chart 42 (heat and cold production) first showed the carrier inputs for heat production, not the amount of heat/cold produced by each carrier. For CHPs, this meant that all energy going into the CHP showed up in the chart. This fix ensures that the graph shows the (actual) heat and cold produced by the carriers (i.e. by multipling inputs by heat_and_cold_output_conversion).

### DIFF
--- a/gqueries/general/heat/heat_production_from_ambient.gql
+++ b/gqueries/general/heat/heat_production_from_ambient.gql
@@ -3,8 +3,8 @@
 - unit = PJ
 - query =
     SUM(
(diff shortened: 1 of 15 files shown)


ChaelKruip:
Good catch @MichielDenHaan!

This looks good and I'll merge it. Will you take care that the new hydrogen query gets a series in the chart?

PR #1389, opened 2017-10-25
Replace the need for VBA with Ruby for:

grdw:
- carrier analysis
- FCE import


grdw:
> Is everyone happy about hard-coding the path to ETDataset?

Well.... It's more so that this was already hard-coded for the dataset import in the first place, but this can be made dynamically with an `.env` file ofcourse 👍

DorinevanderVlies:
> Otherwise, looks fine to me. 👍 If @DorinevanderVlies can confirm that it works as expected then I'll be very pleased for this to be merged.

Works as expected!

PR #1410, opened 2017-10-31
Make the ETDataset path dynamic

grdw:
Because of the minor effort I thought [this](https://github.com/quintel/etsource/pull/1389#issuecomment-339356749) might actually be doable.

Note: everybody does need to do a `mv .env.default .env` + run `bundle install` again whenever they want to make use of the rake tasks.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,30 +1,21 @@
-require 'dotenv/load'
+task :environment do
+  require 'bundler'
+  require 'dotenv/load'
+  require 'pathname'
 
-def encrypt_balance(directory)
-  if File.exists?('.password')
-    password = File.read('.password').strip
-  else
-    puts "File .password not found in root. Cannot encrypt energy balance CSV"
-    exit(1)
-  end
+  require 'fileutils'
+  require 'yaml'
+  require_relative './helpers/encrypt_balance'
 
-  csv  = directory.join('energy_balance.csv')
-  dest = directory.join('energy_balance.gpg')
+  Bundler.require(:development)
+  Dotenv.load('.env', '.env.default')
 
-  if csv.file?
-    dest = csv.dirname.join('energy_balance.gpg')
-    system("gpg --batch --yes --passphrase '#{ password }' -c --output '#{ dest }' '#{ csv }'")
-    puts "  - Encrypted energy balance"
-  end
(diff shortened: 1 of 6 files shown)


jorisberkhout:
Could you assign someone to review this PR, @grdw ?

Rakefile
```diff
@@ -1,3 +1,5 @@
+require 'dotenv/load'
```
antw: This is not quite the right way to do this, since it's loading dotenv directly instead of via Bundler. You might get a different version than expected.

We could make this nicer by adding an `:environment` Rake task which is a prerequisite to all the tasks which need a third-party library (including the `import:*` tasks). The [Atlas `:environment`](https://github.com/quintel/atlas/blob/master/lib/tasks/environment.rb) might serve as an example:

```ruby
task :environment do
  require 'bundler'
  Bundler.require(:development)

  Atlas.data_dir = File.expand_path(File.dirname(__FILE__))
end

namespace :import do
  task dataset: :environment do
    # ...
  end

  task node: :environment do
    # ...
  end

  task carrier: :environment do
    # ...
  end
end
````

File: Rakefile [hunk 2]
```diff
@@ -47,7 +49,7 @@ namespace :import do
 
     datasets.each do |country, year|
       dest = Pathname.new("datasets/#{ country }")
-      csvs = Pathname.glob("../etdataset/data/#{ country }/#{ year }/*/output/*.csv")
+      csvs = Pathname.glob("#{ ENV['ETDATASET_PATH'] }/data/#{ country }/#{ year }/*/output/*.csv")
```
antw: Pass the dataset path to `encrypt_balance` as an argument rather than accessing the global variable directly. This means that we can also check in the Rake task that a value has been set.

antw:
Thanks @grdw, I merged this. :+1:

I think the amount and complexity of the Rake tasks is getting a little silly though (which I think you recognised by extracting `encrypt_balance` to a separate file). If we add more in the future, perhaps it will be time to use something like [Thor](https://github.com/erikhuda/thor) to build a proper CLI for ETSource.

grdw:
427 lines for a single Rakefile is a bit pushing it, 😅 so Thor seems like a pretty good alternative.

PR #1411, opened 2017-10-31
Bump Ruby version of ETSource

grdw:


antw:
I haven't merged this yet since I guess most won't have 2.4.2 installed yet. I'll write up some instructions for everyone else and will merge this at the same time as the 2.4.2 branches for ETE and ETM (probably next week).

grdw:
Can this be merged now, or is this still something which we need to hold off?

PR #1416, opened 2017-11-02
Fixed energetic demand of fertilizer industry

jorisberkhout:
This pull request, combined with https://github.com/quintel/etengine/pull/958 and https://github.com/quintel/etdataset/pull/737, fixes https://github.com/quintel/etmodel/issues/2354.

The main idea behind this fix is that experts told me that about 4.3 GJ per tonne of NH3 is needed for what we in the ETM call the Haber Bosch process. This energy is supplied by the waste heat of the SMR process, which in turn requires about 35 GJ per tonne of NH3. I translated this into the SMR node having an output of useable heat, with an output efficiency of (4.3/35)=12%.

If the user decides to produce hydrogen for the fertilizer industry by P2G, the source of waste heat evaporates and the model will use burners to supply the heat to the Haber Bosch node directly.


jorisberkhout:
Closes https://github.com/quintel/etsource/issues/1417. See https://github.com/quintel/etsource/issues/1417#issuecomment-351737367 for explanation.

PR #1418, opened 2017-11-08
Remove deprecation warning for gpg

grdw:


grdw:
Option is still required for semaphore eventhough it's derecated. Close this PR.

PR #1423, opened 2017-11-15
Baarland init input

MichielDenHaan:


datasets/baarland/curves/heat/1987
```diff
@@ -0,0 +1 @@
+../../../nl/curves/heat/1987
```
grdw: These are symlinks 👍 very proud 🥇

PR #1437, opened 2017-11-17
Chart207 kerosene production

marliekeverweij:
Hoi!

Ik heb net zes queries gemaakt voor een nieuwe chart voor synthetische kerosine. Zou jij er even naar willen kijken?

### DIFF
--- a/gqueries/output_elements/output_series/vertical_stacked_bar_207_synthetic_kerosene_production/transport_final_demand_kerosene.gql
+++ b/gqueries/output_elements/output_series/vertical_stacked_bar_207_synthetic_kerosene_production/transport_final_demand_kerosene.gql
@@ -1,0 +1,9 @@
+- unit = PJ
+- query =
+    DIVIDE(
+        SUM(
+            V(bunkers_useful_demand_planes,demand),
+            V(transport_final_demand_kerosene,demand)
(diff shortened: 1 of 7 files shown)


gqueries/output_elements/output_series/vertical_stacked_bar_207_synthetic_kerosene_production/transport_final_demand_kerosene.gql
```diff
@@ -0,0 +1,9 @@
+- unit = PJ
+- query =
+    DIVIDE(
+        SUM(
+            V(bunkers_useful_demand_planes,demand),
```
grdw: Just me not understanding energy maybe, but isn't there a `final_demand` for bunkers?

marliekeverweij: Yes you are right! But this `final_demand` is varying for kerosene when synthetic kerosene is produced and that should not be the case in this chart. Therefore I chose to use the useful demand in this chart.

grdw:
@marliekeverweij would you be so kind as to delete your own branch 👍

PR #1441, opened 2017-11-21
Add ptl to chart 198

marliekeverweij:
I have added a output_element_serie for synthetic kerosene for chart 198. This is not the total production, but the total production minus the export of synthetic kerosene. So what results is the final demand of synthetic kerosene for bunkers.


marliekeverweij:
Besides the new output_element_serie, I have renamed the chart-folders with the right chard id's

jorisberkhout:
Cool! This closes https://github.com/quintel/etmodel/issues/2559, right @marliekeverweij ?

PR #1444, opened 2017-11-23
Remove coal freight train from efficiency input

antw:
Adjusts the `transport_trains_efficiency` input to remove a nonexistent coal freight train converter, which was causing this input to fail.

Airbrake refs: [[1]](https://quintel.airbrake.io/projects/35333/groups/2089917896139495527), [[2]](https://quintel.airbrake.io/projects/35333/groups/2089879752283281218), [[3]](https://quintel.airbrake.io/projects/35333/groups/2089227749026716000), [[4]](https://quintel.airbrake.io/projects/35333/groups/2089492077646733032).


antw:
I'm noticing that there's also a `transport_freight_train_using_coal_share` input. Perhaps either the freight-coal converter hasn't been added, or that input should also be deleted. There's no slider in the ETM for freight-coal, so I'm guessing it's the latter?

MichielDenHaan:
yeah it's the latter, we decided not to add freight coal trains

PR #1454, opened 2017-11-24
New new initializer inputs

grdw:
*!!! WIP !!!*

Need somebody, either @michieldenhaan @marliekeverweij or anybody for that matter to 

- [x] revisit all the elements and see if the graph_methods are correctly whitelisted. 
- [x] Re-implement the transport sector for initializer methods


edges/agriculture/agriculture_final_demand_steam_hot_water-agriculture_useful_demand_useable_heat@useable_heat.ad
```diff
@@ -1,3 +1,5 @@
 - type = share
 - reversed = false
-- groups = [final_demand]
```
michieldenhaan: why is this in final demand?

edges/agriculture/agriculture_final_demand_steam_hot_water-agriculture_useful_demand_useable_heat@useable_heat.ad
```diff
@@ -1,3 +1,5 @@
...
\ No newline at end of file
+- groups = [final_demand]
```
michieldenhaan: why is this in final demand?

grdw:
@michieldenhaan please ignore that. I need to redo the whitelisting properly anyway 😅

nodes/buildings/buildings_solar_pv_solar_radiation.central_producer.ad
```diff
@@ -36,3 +37,6 @@
 - hours_maint_nl = 0.33
 - hours_remov_nl = 0.0
 
+~ demand = CENTRAL_PRODUCTION(buildings_solar_pv_solar_radiation, demand)
```
grdw: @antw do you know why if I do:

```ruby
node = Atlas::Node.find(:buildings_solar_pv_solar_radiation)
node.save
```

these lines re-appear?

Is this a bug?

antw: > Is this a bug

Kind of, yes. The queries [are defined in Ruby](https://github.com/quintel/atlas/blob/3367cff931387fe7ca211b998266b558f390b270/lib/atlas/node/central_producer.rb) so that modellers wouldn't have to write them over and over for 50+ central producer nodes. Unfortunately Atlas doesn't know this and `save` writes them to the file anyway.

If there aren't too many affected files, I suggest just deleting the newly-added `~ demand` and/or `~ full_load_hours` since they're not related to the PR.

---

*We might reconsider this at a later date. Not including the lines was supposed to be a benefit since it means we could easily change the queries in Ruby. However, we've not done that even once in five years, so perhaps it will makes sense for us to remove this magic and include the queries with the files (not in this PR though). If you have strong feelings either way, we can take this discussion to Basecamp and involve the modellers.*

grdw: > so that modellers wouldn't have to write them over and over for 50+ central producer nodes.

Ah. Next time I'll just comment that line 👍 It's a bit tedious removing it all manually

grdw:
I'm not moving any datasets over to the new style. Will do that in a separate pull request.

PR #1458, opened 2017-11-28
Add P2L to investment table

marliekeverweij:
I have done the following things:
- created three P2L-queries for in investment table
- changed the node name `bunkers_flexibility_p2l_electricity` to `energy_flexibility_p2l_electricity`
- this change caused that I had to change many other output element series, inputs, edge etc


marliekeverweij:
Some of the changes in the local datasets are expected, because of the node name has changed.

However, I do not know why the `freight_tonne_kms` of many applications have changed!
screen shot 2017-11-28 at 10 47 33

Let's not merge it yet....

PR #1461, opened 2017-11-30
Fixing minor mistake p2g nodes source analysis

marliekeverweij:


marliekeverweij:
Two little mistakes in calculation land_use_per_unit:
- /1000 from m2 to km2, but should be /1000000
- took 6.2 instead of 6.3 for the multiplication

PR #1465, opened 2017-12-04
Fuel prices update 2017

michieldenhaan:
Note: the change in MJ per kg for bio_kerosene is correct (change was made a few weeks ago in ETDataset but not raked to ETSource)


AlexanderWirtz:
DEBTS:
1. Find a good source for biogas and greengas costs. Preferably in comparison to natural gas.

PR #1469, opened 2017-12-05
Updating gem groningen with new transport values

michieldenhaan:


grdw:
@michieldenhaan I should not be the one to clear this pull request simply because the numbers all don't make sense to me. I can say though that this will _technically_ work 👍 

Is there anybody else that could review if this makes works 'energy-wise' for the municipality of Groningen? What I'm trying to say: if I were to test this branch locally and run it in my ETModel and for example: the "Bio-footprint" in my local ETModel would be 10.000x NL, it wouldn't strike me as odd. 😅 Simply, because 10.000x NL sounds fine in my dictionary, if that's what's intended. 🤓 

If you feel hower that you checked this yourself thoroughly than I'm willing to merge this without any problem.

Issue #1330, opened 2017-09-04
Solar roof surface in buildings sector doesn't increase with number of buildings


BerendSmits:
The input number_of_buildings.ad increases demand: 

 `- query = UPDATE(V(buildings_useful_demand_for_appliances,buildings_useful_demand_light,buildings_useful_demand_for_space_heating,buildings_useful_demand_cooling), preset_demand, USER_INPUT())`

but not the roof surface available for solar PV. It think it should, we already do it in the household sector (households_number_of_new_houses.ad and households_number_of_old_houses.ad). 

@AlexanderWirtz @DorinevanderVlies what do you think? Should I make a proposal for a new input? And do you have any idea how many scenario's will be affected (I would assume almost all of them, but the significant impact should be limited to the ones with considerable buildings sector growth).

AlexanderWirtz:
> @AlexanderWirtz @DorinevanderVlies what do you think? Should I make a proposal for a new input? And do you have any idea how many scenario's will be affected (I would assume almost all of them, but the significant impact should be limited to the ones with considerable buildings sector growth).

1. Well spotted!
2. This is a bug not a feature and needs to be fixed: so yes, please propose a fix
3. Agree with you on the impact assessment. Featured scenarios on the front page and Groningen neighborhood and GEA provincial scenario's first come to mind
4. The fix for these affected scenarios would be to keep the solar PV power production in these scenarios constant. It should be possible to work out a database migration which scales back the %-slider in order to counteract the increased roof surface.

DorinevanderVlies:
I agree with @AlexanderWirtz

antw:
> The fix for these affected scenarios (after fixing this issue that is) would be to keep the solar PV power production in these scenarios constant. It should be possible to work out a database migration which scales back the %-slider in order to counteract the increased roof surface.

@nairnamratha has adjusted important presets to keep PV production constant, but personally I'm unconvinced that we should do this for *all* scenarios in the database.

If a visitor set a certain number of buildings and a *percentage* of solar PV, then I think we should respect the choices they made even if it increases the amount of solar *production* in their scenario. They chose that amount of PV after all; they didn't get it previously due to a bug.

---

Including @ChaelKruip

nairnamratha:
Just discussed it with @AlexanderWirtz, he confirmed that in most cases the users have put more thought in to the actual PV production than the slider values themselves, which is why we should keep the PV production constant for all cases even if that requires adjusting the slider values in the scenarios.

[...]

PR #1492, opened 2017-12-19
Scaling roof surface available for PV with growth in number of buildings

nairnamratha:
This refers to the issue: [https://github.com/quintel/etsource/issues/1330](url)
1. Update (by scaling) the roof surface available in buildings for pv with an increase in number of buildings

2. Update total solar electricity production according to new roof surface available in buildings.  (Method already used in households translated for buildings)

3. Changed seven existing preset scenarios.


antw:
@nairnamratha If you're happy with it, this can be merged. The migration in ETEngine has been completed.

PR #1500, opened 2017-12-22
Clean up co2 biomass queries

grdw:
Decided to remove both 'future' and 'present' queries for co2 biomass (liquid/gas/solid) and replace it with a single query that dynamically figures out which `INPUT_VALUE` to take.

I decided to declare the INPUT_VALUE key as a variable, as such:

```ruby
input_query = IF(
  scope.graph.future?,
  climate_relevant_co2_biomass_gas_future,
  climate_relevant_co2_biomass_gas_present
);

INPUT_VALUE(input_query)
```

It can also be done as (losing the idea of declaring a variable and the mandatory `;`):

```ruby
INPUT_VALUE(IF(
  scope.graph.future?,
  climate_relevant_co2_biomass_gas_future,
  climate_relevant_co2_biomass_gas_present
))
```

.. but I think it looses it's readability in this fashion, especially in this case - since the `INPUT_VALUE` is already mixed up in a lot of other different statements.

If there are smarter ways of doing this, than I'm open to changes.


antw:
:+1:

> It can also be done as (losing the idea of declaring a variable and the mandatory ;)

I think I prefer the second style since variables aren't an "official" part of GQL. But these wouldn't be the only queries to use variables *anyway*, so I don't particularly mind if you leave it as it is.

@grdw I'll leave this for you to merge with the ETModel PR.

grdw:
> I think I prefer the second style since variables aren't an "official" part of GQL. But these wouldn't be the only queries to use variables anyway, so I don't particularly mind if you leave it as it is.

Ok. I actually changed it. I don't feel to strongly about either of the solutions to be really honest. With a few line-breaks it actually becomes kinda readable-ish. The other one also has it's downsides, *sway sway*.

PR #1509, opened 2018-01-05
Fuel prices - Update

nairnamratha:
Changed slider LHS to show absolute fuel prices instead of percentage changes for natural gas, oil, coal, biomass and CO2 emissions. 
- Initial, Max, Min values for these items updated to show absolute prices
- Preset scenarios updated with absolute fuel price values instead of percentage change values. 

Note: RHS (showing percentage change values) still needs to be fixed. Check [#2622](https://github.com/quintel/etmodel/issues/2622) for further reference.


gqueries/input_elements/label/price_of_co2.gql
```diff
@@ -1,3 +1,5 @@
-- query = PRODUCT(V(AREA(co2_price)),KG_PER_TONNE)
+- query =
+  PRODUCT(
+    V(AREA(co2_price)),KG_PER_TONNE)
```
grdw: Nitpicky: can this be cleaned up to:

```ruby
- query =
  PRODUCT(
    V(AREA(co2_price)), KG_PER_TONNE)
```

gqueries/input_elements/label/relative_price_of_co2.gql
```diff
@@ -0,0 +1,8 @@
+# Relative price of CO2
+
+- query =
+    100.0 * DIVIDE(
+      SUM(QUERY_FUTURE(-> {Q(price_of_co2)}), PRODUCT(-1,QUERY_PRESENT(-> {Q(price_of_co2)}))),
```
grdw: `PRODUCT(-1, x)` can also be written as `-x` in GQL, the same for adding `+`, dividing `/` and subtracting `-`. At least - in the Sandbox of ETEngine I'm allowed to write: `(V(households_final_demand_electricity, demand) * -1 / 5 + 2 - 10)`. So I think this can become:

```ruby
100.0 * 
((QUERY_FUTURE(-> {Q(price_of_co2)}) - QUERY_PRESENT(-> {Q(price_of_co2)})) /
QUERY_PRESENT(-> {Q(price_of_co2)})
```

inputs/costs/costs_co2.ad
```diff
@@ -1,11 +1,19 @@
-- query = UPDATE(AREA(), co2_price, USER_INPUT())
-- label = &euro;/tonne
-- label_query = Q(price_of_co2)
+
+# - label = &euro;/tonne
+# - label_query = Q(price_of_co2)
```
grdw: If these lines are intended to be removed, please do so. The same for the blank lines following.

inputs/costs/costs_biomass.ad
```diff
@@ -1,13 +1,23 @@
...
-- step_value = 5.0
-- unit = %
+- max_value = 1550.0
+- min_value = 0.0
+- start_value_gql = present:Q(price_of_wood_pellets)
+- step_value = 0.01
+- unit = euro
 - update_period = future
-- update_type = %
+- update_type = mixed
```
antw: FYI: `mixed` is not a valid `update_type`. It should be blank, `%`, or `%y`. In this case I expect it should be blank.

Ditto for the other cost inputs.

inputs/costs/costs_co2.ad
nairnamratha: I left them as comments as these lines refer to the RHS of the slider ( which should show the percentage change in fuel price) but this part of the issue is postponed for later as changes to the ETengine need to be made to facilitate this. 

So if these changes will be made at some point, I guess it's better to leave them as comments so that it's easier to fix. Do you agree?

inputs/costs/costs_biomass.ad
```diff
@@ -1,13 +1,14 @@
 # This now updates ALL biomass products, even woody_biomass (but not bio-fuels). This is required as fix to prevent negative costs from the export of wood_pellets. When a flex_max link type is made possible, we should rethink the workings of this slider
-
-- query = UPDATE(CARRIER(greengas,biogas,wood_pellets,woody_biomass,wood,torrified_biomass_pellets), cost_per_mj, USER_INPUT())
-- label = &euro;/tonne
-- label_query = Q(price_of_wood_pellets)
+- query =
+  present_value = QUERY_PRESENT(-> {Q(price_of_wood_pellets)});
+  relative_value = USER_INPUT() / present_value;
```
grdw: I also created [variables](https://github.com/quintel/etsource/pull/1500#issue-284126917) in a query once, yet refrained from doing so. I'm still not sure where the balance in this case lies, maybe @antw has a stronger opinion in this case. 

One side-note: what is the difference between: 

```
QUERY_PRESENT(-> {Q(price_of_wood_pellets)});
```

and

```
present:Q(price_of_wood_pellets)
```

Doesn't this return the same thing?

michieldenhaan: I first tried ``present:Q(price_of_wood_pellets)`` but it didn't work

antw: > One side-note: what is the difference between: `QUERY_PRESENT` and `present:`

`present:` is used as a prefix for the entire query, not a single expression.

For example:

**Yes:**

```ruby
- start_value_gql = present:Q(price_of_wood_pellets)
```

**No:**

```ruby
- start_value_gql =
    value = present:Q(price_of_wood_pellets) # ← syntax error
    value * 10
```

---

Nonetheless, I think...

```ruby
QUERY_PRESENT(-> {Q(price_of_wood_pellets)})
```

... can be simplified as ...

```ruby
QUERY_PRESENT(price_of_wood_pellets)
```

antw: > I also created variables in a query once, yet refrained from doing so. I'm still not sure where the balance in this case lies, maybe @antw has a stronger opinion in this case.

This is personal opinion, but I would refrain from using variables unless they make a complex query much easier to read and understand. Variables are a quirk of GQL being implemented in Ruby rather than something which is officially supported. However, I doubt support for variables will ever be removed, so... */shrug*.

Here's how I would write the query:

```ruby
UPDATE_WITH_FACTOR(
  CARRIER(
    biogas,
    greengas,
    torrified_biomass_pellets,
    wood,
    woody_biomass,
    wood_pellets
  ), 
  cost_per_mj,
  USER_INPUT() / QUERY_PRESENT(price_of_wood_pellets)
)
```

PR #1519, opened 2018-01-11
insulation bug in residences Reitdiep

MartLubben:
Small mistake in the excel :)


DorinevanderVlies:
Thanks for the fix @MartLubben !

PR #1527, opened 2018-01-11
Updated gem Groningen dataset

michieldenhaan:
Updated:
- green gas in gas network
- extraction/production of carriers to 0
- various (minor) changes in sectors due to mistakes in excel


DorinevanderVlies:
Thanks @michieldenhaan!

PR #1531, opened 2018-01-16
Fixed negative energy problem local datasets

michieldenhaan:
Removed biogas and wood pellet production statements as they led to negative energy (losses for export) / energy that 'appears' out of nowhere
neggreengas


ChaelKruip:
"...as they led to negative energy (losses for export) / energy that 'appears' out of nowhere"

Is this a problem in the modelling of these carriers and should we worry about this happening for other regions as well @michieldenhaan?

michieldenhaan:
I think we simply shouldn't use the ``energy_production_biogas: 0`` and ``energy_production_wood_pellets: 0`` update statements. When I remove them, the ETM automatically sets these demands correctly.

PR #1533, opened 2018-01-19
Added new dataset of s Heer Hendrikskinderen and a small change in th…

MartLubben:
…e rakefile


MartLubben:
@grdw Is it ready to be merged? Not sure what the status quo is of the symlinks.

MartLubben:
> @MartLubben could you add a translation for this area in ETmodel.


@grdw Yes, if with the translation you meant this: 

screen shot 2018-01-22 at 10 07 04

Issue #1289, opened 2017-06-23
Lot of redundancy in the data of datasets


grdw:
Just looking at all the existing local datasets , there's a lot of redundancy in the load profiles, central producers, time curves etc. 

I think a nice way to tackle this problem is as follows:

If a csv or yml file is exactly the same to the one in the base dataset, remove that file and point it to the one of the base dataset. This could be in the form of a symlink. 

If anybody has a better idea, please do suggest.

antw:
> If a csv or yml file is exactly the same to the one in the base dataset, remove that file and point it to the one of the base dataset. This could be in the form of a symlink.

I agree very much. In fact I did exactly this with the recently-added profiles for electric vehicles and household heating: every other dataset just links to the NL profiles (for example, see [UK](https://github.com/quintel/etsource/tree/e97083d319ca7a7b9313d7a78bf865da35662f98/datasets/uk/load_profiles)).

This has the benefit of saving space, but also making it clearer when two or more datasets use the same profiles.

Even outside of local datasets there's a lot of duplication:

```
+-----------------------------------+------------+---------------------------------------------------------------------------------------------------------------------------+
| File                              | Duplicates | Files                                                                                                                     |
+-----------------------------------+------------+---------------------------------------------------------------------------------------------------------------------------+
| buildings_chp.csv                 | 16         | barendrecht, be, br, de, de_Biezen, drenthe, es, eu, fr, groningen, nl, nl2012, noorderplantsoen, paddepoel_noord, pl, uk |
| dhw_normalized.csv                | 16         | barendrecht, be, br, de, de_Biezen, drenthe, es, eu, fr, groningen, nl, nl2012, noorderplantsoen, paddepoel_noord, pl, uk |
| electric_vehicle_profile_1.csv    | 16         | barendrecht, be, br, de, de_Biezen, drenthe, es, eu, fr, groningen, nl, nl2012, noorderplantsoen, paddepoel_noord, pl, uk |
| electric_vehicle_profile_2.csv    | 16         | barendrecht, be, br, de, de_Biezen, drenthe, es, eu, fr, groningen, nl, nl2012, noorderplantsoen, paddepoel_noord, pl, uk |
| electric_vehicle_profile_3.csv    | 16         | barendrecht, be, br, de, de_Biezen, drenthe, es, eu, fr, groningen, nl, nl2012, noorderplantsoen, paddepoel_noord, pl, uk |
[...]

PR #1560, opened 2018-02-12
Update rivierenland profiles with symlinks

grdw:
Part of #1289


grdw:
@michieldenhaan I just did this manually. I've learned that there was an issue in Atlas in regards to the scaling, which I've solved [here](https://github.com/quintel/atlas/pull/115).

PR #1566, opened 2018-02-15
added Amersfoort dataset by E&E

michieldenhaan:


DorinevanderVlies:
### Maximum slider settings

- The slider `wind onshore inland` has a max of 10.0. I think this might be too low for some scenario's.
- The slider `Solar PV plants` has a max of 3.75. I think this might be too low for some scenario's.

###  Error when increasing non-renewable electricity production
**I am not sure whether this issue is dataset related**
When I increase the number of pulverized coal plants import of electricity decreases, but is not replaced with coal as happens in the NL-dataset 
image

michieldenhaan:
The maximum slider settings is a deliberate choice by E&E but I'll double check if they're happy with it. Coal bug is fixed, very nice catch

DorinevanderVlies:
> The maximum slider settings is a deliberate choice by E&E but I'll double check if they're happy with it. Coal bug is fixed, very nice catch

Ok, if it is a deliberate choice It's fine for me

nice, quick fix

PR #1569, opened 2018-02-17
Added MECE queries for energetic final demand.

ChaelKruip:
These queries split the total energetic final demand into the following categories:
* oil and coal products
* gas households
* gas other
* collective heat 
* biofuels
* renewable heat
* electricity


ChaelKruip:
> I’m a bit in doubt about the ‘renewable heat’ category. Currently, it only contains solar thermal but judging from the title I would expect it to include things like ambient heat and geothermal as well. I think solar thermal is too small a carrier to ‘deserve’ its own category.

Yes, I agree. Ambient heat and geothermal are currently not final demand though... Not sure what to do.

> We may want to include it in the ‘collective heat’ category and rename the category to ’heat’.. especially since not all steam hot water in agriculture/industry is really ‘collective’

Just calling it 'heat' does not do justice to the category as that implies that all heat is in there... This is 'sold heat'. Should we call it that?

> Another option is to create a ‘renewable other’ category with woody biomass and solar thermal (such that the biofuels category only contains transport fuels). But 'other' categories aren't very sexy.

I agree. 

> Hydrogen: I’m in doubt whether to add hydrogen as a separate category or not. On the one hand it is already covered in part 3 of the factsheet. But on the other hand it might be confusing that at first glance gas use in a scenario doesn’t change much while in fact there is a huge shift from natural gas to hydrogen. I think most people's first association with the word gas is natural gas, not renewable gas.

I am in doubt too but I'd like to limit the number of series.

If you have no real objections, I would like this to be merged. If you have concrete improvements, please let me know!

michieldenhaan:
I've added buildings to the gas households query (since the current factsheet reads 'gas gebouwde omgeving'. Maybe you can double check it?

> This is 'sold heat'. Should we call it that?

Not too happy with that term. Let's stick with collective.

How about the following category names?
Olie en kolen
Gas gebouwde omgeving
Gas overige (o.a. industrie)
Collectieve warmte
Biomassa en -brandstof
Zonthermie
Elektriciteit

gqueries/modules/factsheet/energetic_final_demand/factsheet_energetic_final_demand_gas_households_buildings.gql
```diff
@@ -0,0 +1,7 @@
+- query =
+    SUM(
+      V(INTERSECTION(USE(energetic),SECTOR(households)),"input_of_network_gas + input_of_hydrogen").sum,
+      V(INTERSECTION(USE(energetic),SECTOR(buildings)),"input_of_network_gas + input_of_hydrogen").sum
```
antw: I think this query is incorrect. `demand_gas_other` uses  

```ruby
INTERSECTION(USE(energetic), G(final_demand_group))
```

(energetic nodes in the final demand group), while this query uses:

```ruby
INTERSECTION(USE(energetic), SECTOR(households))
```

(**all** energetic nodes in households/buildings).

This means that "final_demand_gas_households_buildings" can return a higher value than the first part of the query in "final_demand_gas_other", which in turn results in that query returning a negative (beta scenario 808292):

screen shot 2018-02-20 at 10 55 20

```ruby
INTERSECTION(USE(energetic),SECTOR(households)).length
# => 122

INTERSECTION(INTERSECTION(USE(energetic),SECTOR(households)), G(final_demand_group)).length
# => 7
```

I suspect that what you actually want is something like:

```ruby
INTERSECTION(INTERSECTION(USE(energetic), SECTOR(households)), G(final_demand_group))
```

It might best to define a new query which returns the energetic final demand nodes:

```ruby
INTERSECTION(G(final_demand_group), USE(energetic))
```

... and then reuse that in the two gas queries:

```ruby
SUM(
  V(INTERSECTION(Q(factsheet_energetic_gas_converters),SECTOR(households)),"input_of_network_gas + input_of_hydrogen"),
  V(INTERSECTION(Q(factsheet_energetic_gas_converters),SECTOR(buildings)),"input_of_network_gas + input_of_hydrogen")
)

V(Q(factsheet_energetic_gas_converters),"input_of_network_gas + input_of_hydrogen + input_of_lng + input_of_compressed_network_gas").sum - Q(factsheet_energetic_final_demand_gas_households_buildings)
```

ChaelKruip: Great catch @antw! I am certain that was the idea 👍

ChaelKruip: Fixed and pushed to the new `factsheet-queries-chart-2` branch

PR #1598, opened 2018-03-08
Connected the input of the co2 per imported electricity slider to the…

MartLubben:
… carrier imported electricity. Also rearranged some folders

1. Created the right conversion query between carrier and input: etsource/gqueries/general/co2/average_co2_emissions_of_imported_electricity.gql 

2. Connected the input to the new query

3. Rearranged some folders.


DorinevanderVlies:
This change is a temporary fix because the carrier imported electricity can currently not have dataset dependent attributes. It gives all dataset the CO2 of the Dutch electricity mix for imported CO2

Once we have found a solution for this (probably a similar approach as FCE analysis) the CO2 of imported electricity will be different for each dataset.

PR #1615, opened 2018-03-19
Fix CO2 biomass chart (#188); international transport was added twice

marliekeverweij:
Subtracted international transport emissions from the total co2 emissions in chart #188

This was the issue:
pastedgraphic-8

pastedgraphic-10


jorisberkhout:
This does fix the original issue. I did, however, notice that the CO2-emissions of fossil kerosene are double-counted as well. Compare the numbers in the red rectangles below and the number in the blue rectangle. Could you look into this @marliekeverweij ?

screen shot 2018-03-19 at 15 05 33

jorisberkhout:
Never mind my previous comment. Apparently I got lost in switching branches. This PR indeed fixes the original issue.

One small improvement that @marliekeverweij and I identified is that the CO2 of imported electricity is double-counted.

PR #1625, opened 2018-03-21
Sorted all areas except Brabant in groups. Brabant is done by Michiel

MartLubben:


MartLubben:
https://github.com/quintel/etsource/pull/1624

datasets/roosendaal/time_curves/energy_extraction_uranium_oxide_time_curve.csv
```diff
@@ -1,42 +0,0 @@
-year,preset_demand
```
grdw: Is there a reason these files are removed?

MartLubben: Yes Roosendaal is double

MartLubben: So I left the bd_roosendaal and removed the other

grdw:
screen shot 2018-03-21 at 14 55 02

'Brazil' should be in the 'Third party and other regions' group. Much as it is on beta:

screen shot 2018-03-21 at 14 55 33

The same for EU-27

datasets/nl2013/nl2013.full.ad
```diff
@@ -1,4 +1,5 @@
 - area = nl2013
+- group = country
 - id = 1
 - parent_id = 1
 - enabled.etengine = true
```
grdw: One line down there's a `enabled.etmodel` which is on true. This should be false I guess.

datasets/roosendaal/time_curves/energy_extraction_uranium_oxide_time_curve.csv
grdw: Cool! 👍 would love it, if these sort of things can be done in a separate pull request in the future, but I'm fine with this for now.

MartLubben:
Did the requests with this pull request: https://github.com/quintel/etsource/pull/1628

PR #1657, opened 2018-04-03
Hydrogen

michieldenhaan:


gqueries/general/costs/costs_of_used_hydrogen.gql
```diff
@@ -3,7 +3,6 @@
 - query =
     SUM(
       V(G(cost_hydrogen_production), total_costs_per(:converter)),
-      V(G(cost_hydrogen_infrastructure), total_costs_per(:converter)),
```
ChaelKruip: This line is removed here but not added elsewhere. Where are the infrastructure costs for hydrogen included in the cost queries now?

PR #1667, opened 2018-04-13
Updated GEA solar potentials

michieldenhaan:
+ fixed mistakes in agriculture, industry heat queries/efficiencies (for central geothermal heater)


michieldenhaan:
``energy_hydrogen_flexibility_p2g_electricity`` is removed automatically by the rake task, I have not yet added this new node to the Excel generating the datasets (but it is 0 anyway so shouldn't matter).

The queries are fixes of mistakes I found when I was testing the geothermal slider

PR #1669, opened 2018-04-16
Fixed various factsheet queries

michieldenhaan:


gqueries/modules/factsheet/technologies_section/biomass/factsheet_demand_of_biomass_other.gql
```diff
@@ -1,11 +1,6 @@
 - query =
-      SUM(V(
-            agriculture_final_demand_wood_pellets,
-            other_final_demand_wood_pellets,
-            industry_final_demand_wood_pellets,
-            bunkers_final_demand_bio_kerosene,
-            demand
-          )
-      )
+      SUM(Q(factsheet_supply_woody_biomass),
```
ChaelKruip: I understand why you did this @michieldenhaan but it will obscure any errors as this will always correct for missing energy...

gqueries/modules/factsheet/technologies_section/solar/factsheet_number_of_solar_pv.gql
```diff
@@ -3,6 +3,6 @@
 			households_solar_pv_solar_radiation,
 			buildings_solar_pv_solar_radiation,
 			"number_of_units")),
-			Q(factsheet_OM_solar_pv_number_of_units_correction))
+			Q(factsheet_OM_solar_pv_capacity_correction))
```
ChaelKruip: I cannot find this query anywhere: factsheet_OM_solar_pv_capacity_correction

michieldenhaan: Just a file rename: https://github.com/quintel/etsource/blob/d96340d01b3139505a54debdb40a55a8137a7757/gqueries/modules/factsheet/technologies_section/solar/factsheet_OM_solar_pv_capacity_correction.gql

Renamed it because not only number of units of roof pv but also ha of solar farms is lower when capacity of pv is higher than ETM value

PR #1696, opened 2018-05-08
Sparse graph queries

grdw:
PR for all sparse graph query related work


grdw:
Due to removing a lot of attributes from the `graph_methods` in the nodes, the specs now detect that there are datasets that use these methods in the `graph_values.yml` - which is not allowed - because they aren't whitelisted.

michieldenhaan:
To keep the number of sparse graph queries and user inputs as low as possible I removed whitelistings for demands/parent_shares/child_shares/inputs/outputs that are either always zero/always equal to NL/too obscure to bother users with. See https://github.com/quintel/etlocal/issues/80 for a list

PR #1706, opened 2018-05-16
Ccs industry

grdw:
Implementation of CCS for industry - will be superseded by more detailed modelling

Includes:
- slider to set CCS costs 
- slider to quantify percentage of industrial direct emissions captured
- query which calculates quantity of CO2 thus captured and stored


ChaelKruip:
Great work @RobTerwel!

PR #1712, opened 2018-05-22
Network calculation cleanup 

marliekeverweij:
**This pull request includes:**
1. Removal of gqueries related to the old network calculation in the folder etsource/gqueries/modules/network. Before deleting the gqueries I first checked if they were still used somewhere in other gqueries, charts, tables etc.... and if that was the case I updated the gquery with one from the new network calculation. After deleting a gquery, I checked whether my local model was still loading and this was the case. 
2. Removal of the old network calculation attributes (`simult_sd`, `simult_se`, `simult_wd`, `simult_we` and `peak_load_units_present`) from all the nodes where they are used.
3. Updated the network popup gqueries with new network calculation gqueries. Except for the gquery `network_investment_costs_needed_check`, because in this gquery old network queries are used that cannot be directly replaced by a new one (they do not exist yet). So I have set this gquery temporarily to `query = 1`. 

**The next steps will be:**
1. ETEngine: removal of the old network calculation attributes:
- `simult_sd`
- `simult_se`
- `simult_wd`
- `simult_we`
- `peak_load_units_present`
- `required_additional_network_capacity_in_mw`
- `peak_load_of_explicitly_modelled_lv_technologies`
Are there more?
2. Atlas: delete these attributes (with help of @antw)
3. ETDataset: removal of the old network calculation attributes (simult_sd, simult_se, simult_wd, simult_we and peak_load_units_present) from all the nodes_source_analyses where they are in.


ChaelKruip:
> I'm not sure if we're keeping that...

I vote to remove it.

jorisberkhout:
Thanks for your review @antw . I will move the discussion on the pop up to a separate issue to make it more visible.

jorisberkhout:
Thanks a lot @marliekeverweij !

PR #1750, opened 2018-06-19
Network start costs & bug fix 

MartLubben:
The total costs of electricity now add up to 2.7 Billion, like in NvdT. 

Also found that the the interconnection net returned _nill_ in the total costs of the net query. Q(network_calculation_total_costs_present), this was because it was named: interconnection_net_costs_present, instead of interconnection_net**_total_**costs_present


jorisberkhout:
:+1:

Issue #1773, opened 2018-07-04
parent_values.csv should be moved


antw:
Running `rake import` to import a new dataset from ETDataset will cause the parent_values.csv file to be removed. I understand this is used by ETLocal.

@michieldenhaan How is this file created? If it exists in ETDataset I can add it to the import script so that changes are imported with the rest of the dataset. If not, I think it might have to be moved to a new directory.

michieldenhaan:
This file was created manually with the following query:
``TXT_TABLE(ALL(),key, demand, input_of_electricity, input_of_network_gas, input_of_coal, input_of_crude_oil, input_of_wood_pellets, input_of_steam_hot_water, input_of_coal_gas, input_of_useable_heat, input_of_hydrogen, input_of_diesel, input_of_biodiesel, input_of_gasoline, input_of_bio_ethanol, output_of_useable_heat)``

The file is mainly used to allocate a certain share of the NL steel/aluminium/refineries/fertilizers to a local dataset. Due to certain nodes in these sectors with specific carrier input ratios the user cannot freely set carrier demands for these sectors (see https://github.com/quintel/etlocal/issues/72).

I'm happy with moving it to a new directory. I think it's not so trivial to generate it in ETDataset since it contains information on nodes quite 'to the left' in the graph, e.g. electricity input of ``industry_aluminium_electrolysis_bat_electricity``. But it's not impossible. In the future, we probably want to have a separate parent_values.csv for NL2016, so in that sense it is nice if it's linked to the dataset in some way.

Maybe @ChaelKruip or @jorisberkhout have an opinion on this?

ChaelKruip:
> Maybe @ChaelKruip or @jorisberkhout have an opinion on this?

Whatever seems most clean and maintainable is alright we me. I think @jorisberkhout is more at home in this region of the model. I'll trust his judgement.

antw:
> I would suggest to move this file to /datasets. 

I think the file contains demands and shares specific to NL. Perhaps it could be moved to `/datasets/nl` alongside the central producers and primary production CSVs?

PR #1774, opened 2018-07-04
`rake import` also imports bicycles_demand.csv

### LINKED ISSUE #1772: Rake import dataset script should also import bicycles_demand.csv
With the most recent changes to the transport sector in the ETM, the transport analysis on ETDataset now generates a file called `bicycles_demand.csv`. When importing a dataset from ETDataset to ETSource, this file included in the import. Would you have time to pick this up, @antw ? Knowing my own Ruby skills this will probably take me a while.

antw:
Updates `rake import` to also copy the bicycles_demand.csv file from ETDataset.

Note that the NL2015 dataset has had a "parent_values.csv" manually added to the "demands" directory; running `rake import` will remove this file *even though it is needed by ETLocal*. Therefore if you choose to import the NL2015 dataset, prior to committing your changes you should follow it up with `git checkout datasets/**/parent_values.csv` to restore the parent_values.csv file. This is being tracked separately in #1773.

Closes #1772.

### DIFF
--- a/tasks/etdataset/import_dataset.rb
+++ b/tasks/etdataset/import_dataset.rb
@@ -45,6 +45,8 @@ namespace :import do
           cp_csv(csv, dest.join('shares'))
         when /time_curve$/
           cp_csv(csv, dest.join('time_curves'))
+        when /^bicycles_demand/
+          cp_csv(csv, dest.join('demands/bicycles_demand.csv'))
         when /^metal_demands/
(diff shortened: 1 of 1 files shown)


jorisberkhout:
Thanks @antw ! Works like a charm. We'll only be needing this for the German and Danish dataset for now, so we won't phase this problem.

Issue #1788, opened 2018-08-08
Transport p2p should be included in energy sankeys


michieldenhaan:
Transport p2p is currently not included in the energy flow sankeys because it's not in ``Q(p2p_producing_converters_sankey)``. This is because this query only includes nodes in the  ``cost_flexibility`` group. ``transport_car_flexibility_p2p_electricity`` is not (and cannot be) in this group because it has no cost attributes.

To include transport p2p in the sankey we can let ``Q(p2p_producing_converters_sankey)`` include all nodes in a (to be created) ``p2p`` group or add transport p2p manually to ``Q(p2p_producing_converters_sankey)``. I prefer the former.

PR #1793, opened 2018-08-21
created new p2p group for sankey

michieldenhaan:
Created new ``p2p`` group that includes all p2p converters (households p2p, car p2p and pumped storage).

Fixes: https://github.com/quintel/etsource/issues/1788


michieldenhaan:
Added groups for p2h, p2l, p2g as well for consistency; found out that the refineries p2h node wasn't included in the sankey either for a similar reason as transport p2p. Separate groups (unrelated to the cost calculation) solves this.

PR #1822, opened 2018-09-21
Add percentage sliders for carriers used in other industry

marliekeverweij:
This PR includes:
- add new input statements for the percentage sliders used in other industry
- delete old input statements for size and efficiency in other industry
- change in country datasets (BR, ES, EU, FR, NL2012, PL, UK, example): now food and paper sectors are separated for the other industry sector
- delete old queries for charts that are going to be deleted (other industry OTHER charts)


ChaelKruip:
> Do you think that's worth the effort @ChaelKruip, @antw?

Although I dislike such improper naming very much, at the moment I think we have bigger fish to fry. @michieldenhaan can you make an overview issue about bad naming of queries, inputs, nodes etc. and so we can make it part of a bigger maintenance effort?

PR #1834, opened 2018-10-08
hiding Bommelstein

michieldenhaan:
@AlexanderWirtz Hiding Bommelstein for now. Since Bommelstein is not on BETA the dataset will be removed automatically after the deploy. Is that okay or do you want to retain a copy of it?


AlexanderWirtz:
Excellent idea. No reason to retain a copy. Does this mean the scenarios need to be deleted or will they simply not work any more?

PR #1836, opened 2018-10-09
Minor change due to multiplication of efficiencies of compressor and …

DorinevanderVlies:
…pipeline instead of adding them


michieldenhaan:
Hm, what's happening here?

PR #1872, opened 2018-12-11
Fix wind & solar flh sliders

marliekeverweij:
Closes https://github.com/quintel/etmodel/issues/2906


antw:
Unfortunately this doesn't appear to work as expected. Adding the solar PV nodes to the preset_demand group prevents ETEngine from calculating their demand (which is what the preset_demand group is mean to do). This means that the demands in the graph don't make sense when the user changes the number of panels installed.

For example, increasing buildings solar PV from the default 2.8% to 50%:

screen shot 2018-12-17 at 09 57 51

I haven't tested them, but I would guess this is also the case for the other PV producers.

Perhaps it's possible for these inputs to also update the demand of the producer?

marliekeverweij:
Thanks for testing.

> Perhaps it's possible for these inputs to also update the demand of the producer?

I will test this!

marliekeverweij:
Closing this PR as it does not fix the issue. Saving the branch for later, when somebody has time to fix it!

PR #1890, opened 2019-01-30
Add missing Stedin municipality datasets to ETSource

marliekeverweij:
Criteria: more than half of the neighbourhoods of the municipality should be lying in the Stedin region.

Fixing this BC to-do:
https://3.basecamp.com/3797409/buckets/10235523/todos/1503995676#__recording_1564685526


redekok:
We forgot to remove the curves from the datasets (sorry, my bad for not telling you, @marliekeverweij!) so we should do that first before merging this branch into the master. @michieldenhaan, you have a nice script to do that, haven't you?

PR #1891, opened 2019-01-31
Rebase regional report

MartLubben:


MartLubben:
@antw I deleted the double queries and kept the right ones.

PR #1893, opened 2019-02-01
Needed new queries for scenario report

MartLubben:


MartLubben:
Together with ETModel PR: https://github.com/quintel/etmodel/pull/2941

PR #1917, opened 2019-02-15
Update specs (mainly costs) heat networks built environment

marliekeverweij:
Closes https://github.com/quintel/etmodel/issues/2962

The costs for heat networks in the built environment have been updated. We only assume **secondary infrastructure costs for heat networks in households and buildings**: this contains the distribution pipes and the connections to the households/buildings including heat exchangers. For agriculture and industry the heat network costs are assumed to be zero for now: usually the heat source is relatively close to the industry. 

**Households**

- Investment costs per connected house: 6750 euro
- Yearly costs per connected house: 92.5 euro/year

**Buildings**

- Investment costs per 11kW connection (building of 143m2 and 0.08kW/m2): 10221.8 euro
- Yearly costs per 11kW connection: 194 euro/year

The cost of primary (transport) infrastructure is mainly determined by the distance between source and feed-in point. These are assumed to be zero for now, because we do not have a good estimate distance yet. 

Nodes_source_analyses for heat networks built environment have been added to ETDataset https://github.com/quintel/etdataset/commit/65bdd8ce8cc616e9274e0a029a3074e78a99d1f3


michieldenhaan:
Not sure if this is a problem but reducing the capacity of ``households_space_heater_district_heating_steam_hot_water`` from 10 kw to 8 kw leads to quite some deficits (without additional insulation):

etmodel test - energy transition model - your free independent comprehensive fact-based scenario builder 2019-02-18 09-14-10

To get rid of those insulation has to be increased to +/- 35%

etmodel test - energy transition model - your free independent comprehensive fact-based scenario builder 2019-02-18 09-15-52

For Buildings it seems fine except for the large spike in February
etmodel test - energy transition model - your free independent comprehensive fact-based scenario builder 2019-02-18 09-17-35

marliekeverweij:
Interesting!

Maybe it is realistic that houses need to be insulated to +/- label A when connected to a heat network? I don't know, maybe @AlexanderWirtz knows this.

AlexanderWirtz:
Very interesting indeed, but that does not seem reasonable. THe whole reason we or anyone else likes heat networks is that they don't require extensive insulation to be done. I wonder what that says about our profiles.

michieldenhaan:
> I wonder what that says about our profiles.

I think it's a mix of things, the differences between housing types also play a role:

Here's deficits per housing type for 100% district heating, no additional insulation on BETA (10 kw per residence):

beta

For apartments and terraced houses (almost) no deficits occur, for the other housing types deficits are quite big.

Here's the same scenario but now on the ``heatnetworkcosts`` branch (8 kw per residence):

networkcostbranch

Again, apartments are fine. However, for terraced houses deficits now occur almost everyday from mid January till April.

I think we should stick with 10 kw

marliekeverweij:
Okay, let's set the heat network to 10 KW. (Even though we got advised to set it to 8 KW)

As @michieldenhaan suggested the heat demand profile for houses connected to a heat network should actually be flatter than the spiky profile we have. Because in case of a heat network the house will be heated more constant throughout the day. This applies to more technologies (like heat pumps). I will create an issue for that.

AlexanderWirtz:
> For apartments and terraced houses (almost) no deficits occur, for the other housing types deficits are quite big.

That makes sense. A gas-fired heater has a capacity of > 20 kW typically. This is enough to heat a sizeable house. 8 or even 10 kW should not be enough. We should probably keep it at 10 kW, but also ask our friends at Berenschot, Over Morgen or Greenvis, what the typical range of capacity is for different houses. I don't know if you dare to keep calling Ewald @marliekeverweij, but I do feel we are getting to the heart of this matter here.

PR #1927, opened 2019-02-21
Update Leeuwarden dataset: increase biogas used in CHPs agriculture

marliekeverweij:
I have updated the biogas used in CHPs in agriculture to match with the 41PJ electricity production through biogas Klimaatmonitor reports for Leeuwarden here: https://klimaatmonitor.databank.nl/dashboard/Dashboard/Hernieuwbare-Energie/Hernieuwbare-elektriciteit--315/?regionlevel=gemeente . 

The only thing is that the total heat and electricity demand for agriculture simultaneously increases while doing that! Heat from 1 to 35 TJ and electricity from 39 TJ to 80 TJ. Shall we leave that?

### DIFF
--- a/datasets/GM0080_leeuwarden/graph_values.yml
+++ b/datasets/GM0080_leeuwarden/graph_values.yml
@@ -1,12 +1,12 @@
 ---
 agriculture_chp_engine_biogas:
-  demand: 0.0
+  demand: 97.6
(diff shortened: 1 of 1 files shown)


marliekeverweij:
Adjustment to biogas WKK in agriculture is not needed anymore.

PR #1931, opened 2019-02-21
Update oil and gas shares for buildings

redekok:
Some of the household final demand shares are also updated. Is this because of the Belgium parent file you've just created @michieldenhaan ?


michieldenhaan:
I don't think so but the changes are small so it might have something to do with how ETLocal exports floats or so

Issue #1766, opened 2018-06-27
The CO2 emissions of imported electricity should be dataset dependent. 


MartLubben:
The CO<sub>2</sub> emissions of imported electricity should be dataset dependent, but is independent, see
[this issue](https://github.com/quintel/etmodel/issues/2817).

The carrier [imported electricity](https://github.com/quintel/etsource/blob/master/carriers/imported_electricity.ad#L4) has as attribute `co2_conversion_per_mj`

How could we set the attribute`co2_conversion_per_mj` of the carrier [imported electricity](https://github.com/quintel/etsource/blob/master/carriers/imported_electricity.ad#L4) with the dataset?

The FCE method could be a possible direction of a solution. @ChaelKruip or @antw Do you have an idea how we could do this?

antw:
Adding an imported_electricity.yml file in [datasets/??/fce](https://github.com/quintel/etsource/tree/master/datasets/nl/fce) would probably do the trick; the [wood_pellets.yml](https://github.com/quintel/etsource/blob/master/datasets/nl/fce/wood_pellets.yml) file provides a good example. I'd also have to enable it in ETEngine but that's simple enough...

michieldenhaan:
Will this also work for countries with ``has_fce = false``? It looks like NL is the only country with an FCE analysis. And is it possible to specify imported electricity emissions for local datasets? All FCE carriers are currently symlinked to NL2015. I suspect this might become relevant for a project.

antw:
> Will this also work for countries with `has_fce = false`?

Should do. `has_fce = false` will remove the "Fuel chain emissions" checkbox and sidebar item in ETModel, but won't affect the behaviour in ETEngine. `co2_conversion_per_mj` will be calculated, but [the other attributes](https://github.com/quintel/etengine/blob/master/app/models/qernel/carrier.rb#L11-L15) (exploration, extraction, treatment, etc) will not be.

AlexanderWirtz:
@michieldenhaan it looks like we have a few days to fix this in the Antwerp project, no? The work for that project is scheduled to be done in March.

michieldenhaan:
Yes, I might need a little help from @antw to make the required ETEngine changes. I'll write up a small plan in due course.

marliekeverweij:
We would like to have the `co2_conversion_per_mj` dataset dependent for the following carriers:

- imported_heat
- imported_electricity
- biogenic_waste
- non_biogenic_waste

To accomplish this we need to:

[...]

PR #1959, opened 2019-03-14
CO2 emissions of imported electricity and heat are now dataset dependent

marliekeverweij:
Closes https://github.com/quintel/etsource/issues/1766

In this PR:
- yml-files for these two carriers are added to all datasets
- CO2 of imported electricity now differs per country and is based on [European Environment Agency](https://www.eea.europa.eu/data-and-maps/daviz/co2-emission-intensity-5#tab-googlechartid_chart_11_filters=%7B%22rowFilters%22%3A%7B%7D%3B%22columnFilters%22%3A%7B%22pre_config_ugeo%22%3A%5B%22European%20Union%20(current%20composition)%22%5D%7D%7D):
Screen Shot 2019-03-14 at 13 44 03

- CO2 of imported heat is assumed to be the same as NL ([35.96 kg/GJ](https://www.co2emissiefactoren.nl/lijst-emissiefactoren/))
- input statements are updated: now they set the FCE start_values
- max value of `co2_emissions_of_imported_heat.ad` is lowered: 100 kg/GJ instead of 500 kg/GJ


gqueries/general/co2/average_co2_emissions_of_imported_electricity.gql
```diff
@@ -1,14 +1,11 @@
...
-            V(
-               CARRIER(
-                imported_electricity), co2_conversion_per_mj),
-                1000 * 3.6
-          )
-    
-- unit = gCO2/KWh
+    PRODUCT(
+      V(CARRIER(imported_electricity), co2_conversion_per_mj),
+      1000 * 3.6
```
ChaelKruip: Better to use `MJ_PER_KWH * THOUSANDS` instead of hard-coded numbers

inputs/flexibility/import_export/co2_emissions_of_imported_electricity_future.ad
```diff
@@ -1,8 +1,17 @@
-#The user insert g/KWH and the ETM processes kg/MJ. For this reason we divide it by 1000 and with MJ_PER_KWH
+# The co2_conversion_per_mj defined in the FCE-module can only be updated by
+# updating shares. Therefore, we have to calculate the shares of renewable electricity 
+# (0 g/kWh) and fossil electricity (1000 g/kWh) and then update the FCE.
 
-- query = UPDATE(CARRIER(imported_electricity), co2_conversion_per_mj, USER_INPUT() / 1000 / MJ_PER_KWH )
+- query =
+    share_fossil_electricity = USER_INPUT() / 1000;
```
ChaelKruip: Dividing through integers is dangerous! Please use 1000.0 or, even better, `THOUSANDS`

inputs/flexibility/import_export/co2_emissions_of_imported_electricity_present.ad
```diff
@@ -1,8 +1,17 @@
-#The user insert g/KWH and the ETM processes kg/MJ. FOr this reason we divide it by 1000 and with MJ_PER_KWH
+# The co2_conversion_per_mj is defined in the FCE-module can only be updated by
+# updating shares. Therefore, we have to calculate the shares of renewable electricity
+# (0 g/kWh) and fossil electricity (1000 g/kWh) and then update the FCE.
 
-- query = UPDATE(CARRIER(imported_electricity), co2_conversion_per_mj, USER_INPUT() / 1000 / MJ_PER_KWH)
+- query =
+    share_fossil_electricity = USER_INPUT() / 1000;
```
ChaelKruip: Same here

inputs/flexibility/import_export/co2_emissions_of_imported_electricity_future.ad
marliekeverweij: Yes! I will use 1000.0 though, because this is the max slider value. I think it is going to be a bit intransparent when we would use `THOUSANDS`. When people want to update the slider_max its maybe not clear that this value also has to be updated

PR #1961, opened 2019-03-15
Add new input statement for solar parks slider and remove old one

redekok:
Changed the slider for solar pv parks from `number_of_energy_power_solar_pv_solar_radiation` into `capacity_of_energy_power_solar_pv_solar_radiation`, which now allows the user to set the capacity of solar parks instead of the number of units.

@ChaelKruip @marliekeverweij If you approve this change, we should migrate the existing scenarios before merging the pull request.

Fixes https://github.com/quintel/etmodel/issues/2992.


ChaelKruip:
@redekok can you check why/which checks fail and fix them?

redekok:
> @redekok can you check why/which checks fail and fix them?

They fail because the presets haven't been updated yet. I'll update those as part of the dataset migration which I've mentioned above. Furthermore, as discussed, I will likely commit more slider changes to this pull request (changing other inputs from number of units to capacity) before we should merge it into the master branch.

PR #1962, opened 2019-03-15
Update gqueries needed for ETFlex after updates heat section

marliekeverweij:
Part of https://github.com/quintel/etflex/pull/450

### DIFF
--- a/gqueries/modules/etflex/etflex_score_deviations_from_reference_scenario.gql
+++ b/gqueries/modules/etflex/etflex_score_deviations_from_reference_scenario.gql
@@ -5,10 +5,10 @@
 # Reference values:
 #
 # DEMAND
-# * Isolation:         2.8
+# * Isolation:          67
 # * Electric cars:      75%
 # * LED:                90%
-# * Fuel cell:           5%
+# * Electric trucks:    75%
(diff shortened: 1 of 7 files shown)


gqueries/modules/etflex/etflex_score_deviations_from_reference_scenario.gql
```diff
@@ -22,11 +22,11 @@
 
 - query =
     - [
-      0.111111 * (INPUT_VALUE(households_insulation_level_old_houses) - 2.8)**2,
+      1.0e-4 * (AVG(INPUT_VALUE(households_insulation_level_apartments),INPUT_VALUE(households_insulation_level_detached_houses),INPUT_VALUE(households_insulation_level_semi_detached_houses),INPUT_VALUE(households_insulation_level_terraced_houses),INPUT_VALUE(households_insulation_level_corner_houses)) - 2.8)**2,
```
ChaelKruip: I don't understand why this line uses the average of all insulation sliders.

Also, the number 2.8 seems to refer to the 'optimal' reference value for the old Rc slider. A percentage (67%?) should be likely used instead.

PR #1967, opened 2019-03-25
Update queries to fix ETFlex

redekok:
- Replace district heating by a hybrid heat pump
- Remove the `households_water_heater_...` queries from the investment query (since these do not exist anymore)

### DIFF
--- a/gqueries/modules/etflex/etflex_level_2/etflex_households_heating_technology_in_use.gql
+++ b/gqueries/modules/etflex/etflex_level_2/etflex_households_heating_technology_in_use.gql
@@ -1,5 +1,5 @@
 - query =
-    IF(INPUT_VALUE(households_heater_hybrid_heatpump_air_water_electricity_share) > 10,
+    IF(INPUT_VALUE(households_heater_hybrid_heatpump_air_water_electricity_share) > 50,
       "hybrid_heatpump",
       IF(INPUT_VALUE(households_heater_combined_network_gas_share) > 50,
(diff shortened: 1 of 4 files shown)


gqueries/modules/etflex/etflex_level_2/etflex_households_heating_technology_in_use.gql
```diff
@@ -1,6 +1,6 @@
 - query =
-    IF(INPUT_VALUE(households_heater_district_heating_steam_hot_water_share) > 0,
-      "district_heating",
+    IF(INPUT_VALUE(households_heater_hybrid_heatpump_air_water_electricity_share) > 10,
```
ChaelKruip: Why has the number 0 been changed to 10? Does the group still add to 100%?

redekok: I think this query is supposed to determine which radio button is active in level 2 of ETFlex. It doesn't set the initial input shares of the heating technologies. These are defined by the [relevant input statement](https://github.com/quintel/etflex/blob/fix-etflex/config/inputs/house/household_heating_technology.yml) in ETFlex. However, if the values of the initial input shares are changed, the active radio button might be changed by this query. I'm actually not entirely sure how the order of these if-statements was determined. And now that I think about it, it might make more sense to change the number for the hhp share from 10 to 50 as well, for sake of consistency. @ChaelKruip what are your thoughts on this?

PR #2003, opened 2019-04-19
Set maximum value of the onshore wind turbine and solar parks sliders to total land area

redekok:
Fixes https://github.com/quintel/etmodel/issues/2912:

- Export ETLocal datasets to ETSource to add `total_land_area` attribute to all AREA-files (and keep the old `land_available_for_solar` and `onshore_suitable_for_wind` attributes for migrations)
- Update input statements and replace `land_available_for_solar` and `onshore_suitable_for_wind` by `total_land_area`
- Update Atlas ref in Gemfile


redekok:
There are two things we have to check before this PR can be merged:

1) I'm not sure why the graph values of most datasets are updated as well. It seems to be the case that the values haven't changed, just the order of attributes. Would there be an easy way to remove the changes made to the graph values from this PR?

2) A lot of changes have been made to (among others) the Antwerpen dataset. @marliekeverweij @michieldenhaan, I'm not sure if all changes are included in this PR. Let's have a look at it next week!

michieldenhaan:
> There are two things we have to check before this PR can be merged:
> 
>     1. I'm not sure why the graph values of most datasets are updated as well. It seems to be the case that the values haven't changed, just the order of attributes. Would there be an easy way to remove the changes made to the graph values from this PR?
> 
>     2. A lot of changes have been made to (among others) the Antwerpen dataset. @marliekeverweij @michieldenhaan, I'm not sure if all changes are included in this PR. Let's have a look at it next week!

I've removed all graph_values.yml files including the Antwerp changes.

PR #2006, opened 2019-04-30
Add Meierijstad to beta

marliekeverweij:


marliekeverweij:
Ah, good catch! 

> Would you have any idea? Could it be the case that when an attribute is used for a sparse graph query, it's not exported directly to the AREA-file anymore?

Yes I think that's the case. We should probably add a sparse graph query for `total_land_area`

redekok:
> We should probably add a sparse graph query for `total_land_area`

Exactly my thought. :-)

PR #2026, opened 2019-05-21
Fixing errors in sankey gqueries

marliekeverweij:
@MartLubben found out that central biomass heaters and geothermal heaters are not displayed in the Sankey. I am not sure if this commit fixes those errors. Geothermal energy is not considered as primary energy, so maybe instead of the geothermal energy (which is now shown in the Sankey with this commit), the produced hydrogen/electricity/heat should be shown as input of the 'central heat'-block. Let me know what you think @michieldenhaan !

### DIFF
--- a/gqueries/output_elements/output_series/sankey/biomass_products_to_central_heat_prod_in_sankey.gql
+++ b/gqueries/output_elements/output_series/sankey/biomass_products_to_central_heat_prod_in_sankey.gql
@@ -1,4 +1,4 @@
 # Query for Sankey diagram: connection between biomass_products and chps
 
 - unit = PJ
-- query = DIVIDE(SUM(V(Q(heat_producing_converters_sankey), primary_demand_of(non_biogenic_waste, greengas , biodiesel , bio_ethanol , bio_lng , bio_oil , biogas , torrefied_biomass_pellets , wood_pellets , corn , manure , wood , biogenic_waste , woody_biomass))),BILLIONS)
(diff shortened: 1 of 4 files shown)


ChaelKruip:
After reading this PR I have a couple of suggestions:

> @MartLubben found out that central biomass heaters and geothermal heaters are not displayed in the Sankey.

For future reference: it is better to first create an issue which clearly states the issue and have any discussion there before creating a PR

> I am not sure if this commit fixes those errors

It is easier to check if your commit has the desired effect once you have a clearly defined issue (e.g., previous point 😄 )

> Geothermal energy is not considered as primary energy, so maybe instead of the geothermal energy (which is now shown in the Sankey with this commit), the produced hydrogen/electricity/heat should be shown as input of the 'central heat'-block. 

This is indeed the route that we try to follow: considering the products of renewable generation as the primary carriers. Are the choices made in this PR consistent with the rest of the model? How do we treat these carriers in the Mekko's etc?

> Let me know what you think @michieldenhaan !

I think this point warrants a bit more communication with the rest of the team (I'd like to hear the opinion of @AlexanderWirtz about this for instance) as it is quite fundamental to the general bookkeeping of the model.

AlexanderWirtz:
@ChaelKruip my opinion on what specific part? I am in favor of showing geothermal energy in the Sankey, but that does not mean we should not follow convention in how we count primary energy. 
Checking with CBS / IEA what convention now dictates (it tends to change), would be wise.

I am also unclear what this implies:
> the produced hydrogen/electricity/heat should be shown as input of the 'central heat'-block. 

What will be shown as the _source_ for the produced power or heat (how do you produce hydrogen from geothermal energy?) Nothing? That would be OK in a Mekko chart (where we show wind power) as primary electricity, but in the Sankey said power or heat would need a qualification (geothermal) as does wind power now.

Do my questions help to make clear what my opinion is?

marliekeverweij:
You're right, @ChaelKruip. We wanted to fix the sankey-issue before the session for Groningen-Drenthe last Thursday, therefore I didn't follow the proper issue-PR-route. Next time I will!

PR #2034, opened 2019-05-27
Implement changes of "curves-project"

DorinevanderVlies:


antw:
I'm not too sure about the symlinking for the derived (local) datasets. In the current `master` branch we link the **curves** directory rather than each of the individual .csv files. I think this makes sense since we expect a local dataset to use the exact same curves as the full-size NL.

There's another advantage to linking the directory instead of the individual curve files: fewer files.

```
master branch: 1 directory symlink × 495 datasets = 495 files
curves branch: 61 curve symlinks × 495 datasets = 31195 files
```

The convention *to date* has been to link directories rather than individual files (we do this also with **demands**, **fce**, and **real_estate**).

If we're going to link to the individual files then I should also update the Atlas scaler (used by ETLocal) before this is merged.

redekok:
@antw I fully agree that it makes more sense to symlink the entire directory instead of all the individual files, at least for the local datasets. Could you help out with updating the script to make this work?

antw:
> Could you help out with updating the script to make this work?

Yes certainly. I can make this a priority tomorrow.

redekok:
> Yes certainly. I can make this a priority tomorrow.

I think I've already managed to do it! Let me know if it's all right like this.

PR #2038, opened 2019-05-29
Multi-year-chart: adding gqueries for final demand and primary co2 per sector and carrier

marliekeverweij:
This PR includes the gqueries per sector and carrier. They have been generated with [these scripts](https://github.com/quintel/modeling_experiments/commit/dcb5c59191f0e2c73002de7108826e52f558dd9b) in the modelling_experiments-repository. However, not all final demand queries could be generated with the script because specific shares had to be added: for example the `sustainability_share` for the heat and electricity gqueries and the greengas-share for the gas gqueries.

The gqueries per sector, per carrier and per sector and carrier can be found in different folders. 

NOTE: final demand query for `solar_thermal` was missing, so has been added @antw

PR #2043, opened 2019-06-05
Update MYC-queries

marliekeverweij:
- add myc-prefix to all MYC-queries
- use consistent units
- link to already existing queries

https://github.com/quintel/etsource/pull/2038


antw:
I made a couple of small adjustments, mostly to the primary CO2 by-sector queries, which specified their unit as `tonnes` but were giving values in MT. Otherwise this looks to be mostly fine.

However there does seem to be a difference between the old and new primary CO2 per-sector queries, which results in a different shapes for by-sector and by-carrier:

| Be carrier | By sector |
| :---: | :--- |
| Screen Shot 2019-06-05 at 18 34 16 | Screen Shot 2019-06-05 at 18 34 31 |

As far as a I can tell, the differences are in the "by carrier" queries for "coal and derivatives", "natural gas and derivatives" and "oil and derivatives":

| Carrier | Old present value | New present value |
| :--- | --: | --: |
| Coal and derivatives | 7.2 MT | 41.7 MT |
| Gas and derivatives | 40.5 MT | 63.0 MT |
| Oil and derivatives | 50.4 MT | 51.6 MT |

I think this is because of a difference in carriers used by the old and new queries. For example [the old query](https://github.com/quintel/etsource/blob/8294bbf7d127d968cac319a4aff745e2bbdedd4a/gqueries/modules/multi_year_charts/primary_co2_chart/per_carrier/primary_co2_final_demand_of_coal_and_derivatives.gql) for coal includes coal and coal_gas, while [the new one](https://github.com/quintel/etsource/blob/8294bbf7d127d968cac319a4aff745e2bbdedd4a/gqueries/general/co2/primary_co2_of_coal_and_derivatives.gql) uses coal, coal_gas, cokes, and lignite.

At the moment I'm not sure whether it's the by-sector or by-carrier queries which are correct.

marliekeverweij:
Thanks for finding this out!

I see what the problem is. The "per carrier"-queries should calculate the `primary_co2_emission` for the carriers on _**final demand**_ level. However, the queries that are now linked to already existing queries to not look at final demand level. So in the new queries some of the `primary_co2_emission`s are double-counted, for example the emissions of coal that is used to generate electricity is now included in the `myc_primary_co2_of_coal_and_derivatives`-query and in the `myc_primary_co2_of_electricity`-query.  I'll fix this

PR #2044, opened 2019-06-06
Added RES datasets to group RES

DorinevanderVlies:
@redekok do we need to change the group of these datasets in ETLocal as well?
If so, can you open an issue for that in ETLocal?


redekok:
I _think_ the only thing that should be changed is the [export script](https://github.com/quintel/etlocal/blob/master/app/models/dataset.rb#L22-L23).

```
elsif geo_id =~ /^RG/ or geo_id =~ /^RES/
      'region'
```

should be changed to: 

```
elsif geo_id =~ /^RG/ 
      'region'
elsif geo_id =~ /^RES/
      'res'
```

This makes sure that the correct groups are assigned to the AREA-file attribute `group` in ETSource datasets.

@michieldenhaan is this correct?

michieldenhaan:
yes, and a minor translation has to be added to ETLocal as well for the pop-up
beta-local energytransitionmodel com - ETLocal 2019-06-07 09-47-17

redekok:
FYI: the relevant changes on ETLocal are requested by https://github.com/quintel/etlocal/pull/194.

DorinevanderVlies:
> LGTM! One more side question: to what extent do the duplicate regions differ in terms of graph values, etc? The recent RES region datasets are based on a summation of all underlying municipalities in ETLocal. I'm not sure if the old regional datasets (e.g. RGUT01_u16) contain more detailed or enriched data compared to the recent RES region datasets?


Good question.

I checked all duplicate datasets. Some of them differed from the new RES datasets. Most differences have been solved. Overview of current status:

image

RES West-Brabant differs from region West-Brabant because the municipality Tholen is wrongfully part of the region. The RES dataset is correct. As the RES dataset will in the future replace the regional dataset I have not changed the regional dataset.

The difference for Groningen and Drenthe is addressed in this issue:
https://github.com/quintel/etsource/issues/2045

Issue #2025, opened 2019-05-20
Primary demand of Export and Curtailment should only include net demand


ChaelKruip:
@AlexanderWirtz pointed out to me that the primary demand mekko shows a lot of oil for "Export and Curtailment":
![image](https://user-images.githubusercontent.com/1303760/58030773-6e862b00-7b1f-11e9-8e4d-eb1e8dd665ae.png)

I always imagined that this mekko displays the **net** primary demand for NL caused by exported and curtailed carriers but it seems to show **all** primary demand instead as can be seen in [this query.](https://github.com/quintel/etsource/blob/master/gqueries/general/primary_demand/primary_demand_of_oil_and_derivatives_from_export.gql)

I would think we should calculate the net primary demand by doing this:
```
SUM(
       V(G(energy_export), "primary_demand_of_crude_oil - demand")
  ) / BILLIONS
```

So we don't count the exported stuff as primary demand.

@AlexanderWirtz, does this make sense? If yes, @marliekeverweij can you change this query and similar ones?

AlexanderWirtz:
I also think the current chart is confusing. I am not an expert on how to calculate what we do want to put in there.

marliekeverweij:
Maybe we should not show the primary demand of the exported fuels at all and only show the primary demand of the sector final demands and  That is in line with the CO2-emission calculation: we also do not take into account the CO2 emissions of exported fuels.

> `SUM(V(G(energy_export), "primary_demand_of_crude_oil - demand") / BILLIONS`

I think it's not a good idea to use this query. This query would kind of show the losses of the process from primary product -> exported fuel. I think that's not very interesting.

I think we should choose to either totally include the primary demand of export fuels or exclude it. What's our opinion on this @michieldenhaan ?

marliekeverweij:
@ChaelKruip , @michieldenhaan and I just had a meeting about this issue. 

**Conclusion:**
The energetic losses related to the production of exported fuels are already included in the primary demand of the refinery sector. So the export-category should not be added to the primary demand mekko, since these losses will then be double-counted.

marliekeverweij:
> **Conclusion:**
> The energetic losses related to the production of exported fuels are already included in the primary demand of the refinery sector. So the export-category should not be added to the primary demand mekko, since these losses will then be double-counted.

So shall I:
- Remove the primary demand of export from the primary demand mekko
- Add queries that only show the primary demand of curtailment
- Rename the category from 'Export & curtailment' to 'Curtailment'

[...]

Issue #45, opened 2012-02-23
Coal costs slider (costs_coal) should also update lignite price


AlexanderWirtz:
Now it only updates coal price

PR #2047, opened 2019-06-07
Update primary demand mekko

marliekeverweij:
Fixes https://github.com/quintel/etsource/issues/2025

Instead of the 'Export & curtailment' - category there is now an 'Energy' - category added to the primary demand mekko with three queries (so no break down in carriers):

1. Export losses of fuels other than refinery fuels
2. Energy-sector own use (this was first secretly included in the industry queries)
3. Curtailment

This PR:
- Removes the total primary demand of export fuels from the primary demand mekko
- Includes a query for the total export losses of exported fuels (excl. refinery products) to the primary demand mekko and updates the export-loss-query for chart #45
- Includes a query for the total curtailment losses to the primary demand mekko
- Includes a query for the primary demand of the energy sector own use to the primary demand mekko. This is the primary demand of these three nodes: 
Screen Shot 2019-06-07 at 15 58 53
- Removes the the primary demand of energy sector own use from the industry queries

The updated primary demand mekko for NL looks like this: 
Screen Shot 2019-06-07 at 16 01 44


marliekeverweij:
The reason why I didn't split the energy sector into primary demand carriers is because I could not figure out how to do this for export losses.
For example: how could you query the primary demand of natural gas in export losses? This is not:
`V(G(energy_export),"primary_demand_of_natural_gas - input_of_natural_gas")`
Because `natural_gas` does not have to be the carrier that is exported, the exported carrier could also be `electricity` or `LNG`.

michieldenhaan:
> The reason why I didn't split the energy sector into primary demand carriers is because I could not figure out how to do this for export losses

The only nodes with non-zero export losses in the queries above are ``energy_export_oil_products``, ``energy_export_coal_gas``, ``energy_export_network_gas`` and ``energy_export_electricity``. The first is a refinery product I think, the second is complicated but should arguably not be in the ``energy_export`` group. The last two are a bit more complicated indeed. I'll plan a meeting next week about this and the more general discussions about CO2, demand and primary demand we've been having over the past month.

PR #2059, opened 2019-06-24
Removing outdated municipality datasets in Groningen-Drenthe from BETA

marliekeverweij:
The provincial and regional datasets have already been updated with new industrial data, but the municipalities have not. As far I as know these municipal datasets are not used for the SVGD, is that true @AlexanderWirtz and @MartLubben ?


marliekeverweij:
@MartLubben let's remove these outdated datasets. Could you have a look at this PR?

PR #2060, opened 2019-06-24
Fix green gas/natural gas distinction in Sankey

michieldenhaan:
Fixes: https://github.com/quintel/etmodel/issues/3062

Sankey now looks as follows:

_100% natural gas in gas network_
etmodel test - Energy Transition Model - Your free, independent, comprehensive, fact-based scenario builder  2019-06-24 11-52-02

_100% green gas in gas network_
etmodel test - Energy Transition Model - Your free, independent, comprehensive, fact-based scenario builder  2019-06-24 11-52-19

### DIFF
--- a/gqueries/output_elements/output_series/sankey/biomass_products_to_agriculture_in_sankey.gql
+++ b/gqueries/output_elements/output_series/sankey/biomass_products_to_agriculture_in_sankey.gql
@@ -4,7 +4,8 @@
 - query =
     DIVIDE(
         SUM(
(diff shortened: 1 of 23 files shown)


ChaelKruip:
Why do we import so much more electricity in the second chart? Is that because green gas is more expensive and the Merit Order changes @michieldenhaan ?

michieldenhaan:
Yes indeed. Import is cheaper so only part of gas use in the electricity sector is replaced with green gas

beta-pro energytransitionmodel com - Energietransitiemodel - Je eigen onafhankelijke, volledige en op feiten gebaseerde energie… 2019-06-24 12-37-56

beta-pro energytransitionmodel com - Energietransitiemodel - Je eigen onafhankelijke, volledige en op feiten gebaseerde energie… 2019-06-24 12-38-13

Issue #1933, opened 2019-02-22
Regional hourly profiles for demand and supply are inconsistent with demand and supply volumes.  


AlexanderWirtz:
Our national dataset for `nl` is based on 2015 data. The regional datasets are based on 2016 energy data (volume), but use 2015 profiles. This means in principle the volumes and profiles are inconsistent. 

For smaller areas this is probably not so significant, but for larger ones, like Groningen + Drenthe, I expect this will prove problematic on closer inspection of the ETM's output. Martijn Douwes asked me about this the other day. Part of the attraction for him of using the ETM for the Groningen+Drenthe project is that the ETM outputs hourly profiles as well as demands. I am thinking about:

- Temperature profiles and heat demand
- solar PV profiles and solar PV power production
- wind profiles for onshore and offshore and wind power production
- ... did I forget something?

Since at some point we will have to update the national dataset anyway (let's hope we can find some budget for this), it seems like a good idea to research or create 2016 demand and supply profiles. 

According to @michieldenhaan the 2015 national profiles are now symlinked in the regional datasets. Is there some way to create a partial 2016 national dataset and symlink to the profiles there, so we can use them. This would allow us to use consistent profiles.  

Curious what you think @ChaelKruip, @michieldenhaan @jorisberkhout, @antw

AlexanderWirtz:
Can this perhaps be done from the Stedin budget?

michieldenhaan:
I don't think so, we only have a couple of days left. But maybe as part of the time-resolved update @ChaelKruip ?

DorinevanderVlies:
Yes, it will be part of that project

marliekeverweij:
This issue is fixed during the curves project! Now all regional datasets have 2016 profiles and the NL dataset uses 2015 profiles.

There is one important exception:
Because of incomplete data for wind and solar in 2015, the NL 2015 dataset will use 2016 profiles for solar PV and wind onshore and 2017 profile for wind offshore. This is also the case for regional datasets.

AlexanderWirtz:
>Because of incomplete data for wind and solar in 2015, the NL 2015 dataset will use 2016 profiles for solar PV and wind onshore and 2017 profile for wind offshore. This is also the case for regional datasets.

@marliekeverweij I am not too happy with this fix, as this means wind data is internally inconsistent. I am sure there is a correlation with wind onshore and offshore.

[...]

PR #2061, opened 2019-06-24
Update solar and wind curves

redekok:
- Update solar and wind curves for all datasets after scripts and curves changes in https://github.com/quintel/etdataset/pull/815 
- Add `GM1966_het_hogeland` to `local_curves.yml` and symlink its curves to `nl2016` instead of `nl`


marliekeverweij:
This is a related issue https://github.com/quintel/etsource/issues/1933 
But I think we could already merge this, cause this is an improvement anyway! 

I think this PR closes these issues, right @redekok? 
https://github.com/quintel/etmodel/issues/3058
https://github.com/quintel/etmodel/issues/3056
https://github.com/quintel/etmodel/issues/3055

redekok:
> I think this PR closes these issues, right @redekok?
quintel/etmodel#3058
quintel/etmodel#3056
quintel/etmodel#3055

Yes, it does. However, it still might not be the best solution. @DorinevanderVlies has proposed a few alternatives in https://github.com/quintel/etsource/issues/1933#issuecomment-505410031. Let's stall the merge until we've decided which alternative is the best solution.

Issue #2062, opened 2019-06-24
Potential future households roof surface available for solar PV could be calculated more exact


marliekeverweij:
Currently the `residences_roof_surface_available_for_pv` is scaled with the total number of residences. We actually do have information available in the AREA-file about the roof surface available for pv **per housing type**: 

![Screen Shot 2019-06-24 at 16 06 00](https://user-images.githubusercontent.com/32833996/60025472-fe2c7580-9699-11e9-81b4-e1435a6a7207.png)

So we could make a more exact calculation of the future potential for pv in households if we want! But I do not know what the priority of this issue is.

ChaelKruip:
Good observation!

@marliekeverweij can you make this more actionable and assign someone? Or perhaps initiate a meeting to force a decision?

marliekeverweij:
We could calculate it better, but that will be kind of hassle:
We have a total roof surface for solar pv in the start year in the AREA-file (`residences_roof_surface_available_for_pv`) Everytime an extra detached house is built we could add the `typical_roof_surface_detached_house_available_for_pv` and when an apartment is broken down we could subtract `typical_roof_surface_apartment_available_for_pv`. But the `residences_roof_surface_available_for_pv` is not calculated with these roof surfaces per housing type, so we will run into problems when a lot of houses are broken down. However, it's not very realistic that that will happen in the future!

Mmmm, I think this does not have priority at the moment. I wanted to identify the issue, so people know who are calculate potential roof surface at the moment, namely: 
`future_residences_roof_surface_available_for_pv = present_residences_roof_surface_available_for_pv * (future_number_of_residences / present_number_of_residences)`
(no distinction made between housing types)

Shall I close the issue again?

ChaelKruip:
> Shall I close the issue again?

Probably best to leave this for now yes.

PR #2063, opened 2019-06-24
 Change housing type sliders to shares instead of number of units

marliekeverweij:
Screen Shot 2019-06-24 at 16 35 19

- This adjustment simplified the housing type sliders; these now only update the area-file, the useful heat demand in the nodes and insulation costs
- However, the total number of houses input statement now updates everything else (see [input statement](https://github.com/quintel/etsource/blob/d7ee58d3a432546e46b4cf1e3ecffe8ab8db6ff7/inputs/demand/households/households_misc/households_number_of_residences.ad) for more information)

Fixes https://github.com/quintel/etmodel/issues/3024

I have send this PR to the three of you, because I think it's good to test it well, but maybe not everybody! So if you have time :) I did test it locally already!

@redekok will create the scenario migration for these sliders

(Related issue: https://github.com/quintel/etsource/issues/2062)


inputs/demand/households/households_misc/households_number_of_residences.ad
```diff
@@ -35,10 +34,11 @@
...
     )
   )
+
 - priority = 0
-- max_value_gql = present:AREA(number_of_residences)
-- min_value = 0.0
-- start_value_gql = present:AREA(number_of_semi_detached_houses)
+- max_value_gql = present:AREA(number_of_residences) * 1.0e-06 * 2
+- min_value_gql = present:AREA(number_of_residences) * 1.0e-06 * 0.5
+- start_value_gql = present:AREA(number_of_residences) * 1.0e-06
```
redekok: I'm not sure if we should express the number of residences in millions. I think this works on country-level but I'm not sure if it works for local scenarios (provinces, municipalities, etc.) as well. Maybe we could express the number in thousands?

inputs/demand/households/households_misc/households_number_of_residences.ad
```diff
@@ -35,10 +34,11 @@
       )
     )
   )
+
 - priority = 0
-- max_value_gql = present:AREA(number_of_residences)
-- min_value = 0.0
-- start_value_gql = present:AREA(number_of_semi_detached_houses)
+- max_value_gql = present:AREA(number_of_residences) * 1.0e-06 * 2
```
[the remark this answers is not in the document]
antw: > I can help think about this if desired!

```
- max_value_gql = present:AREA(number_of_residences) / MILLIONS * 2
```

inputs/demand/households/households_misc/households_number_of_residences.ad
antw: If the unit is `mln`, I believe ETModel will scale it appropriately.

marliekeverweij: Yes it works for me locally! :)

[Screen Shot 2019-06-27 at 14 29 57](https://user-images.githubusercontent.com/32833996/60266298-1265b200-98e8-11e9-9f9b-f0348ca4ef82.png)

inputs/demand/households/households_misc/households_number_of_residences.ad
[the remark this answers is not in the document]
marliekeverweij: See https://github.com/quintel/etsource/pull/2063/commits/69fac0e2d0e5dc3de5a5e9e23578a596c7bee9f3

antw:
> While testing, I found out that the new inputs aren't balancing yet. 

Do the sliders sum correctly to 100% in a blank scenario? Judging by the inputs, it seems like they should (as long as `AREA(number_of_residences)` equals `SUM(AREA(number_of_apartments), AREA(number_of_corner_houses), ...)`.

I would expect that ETModel would then ensure that the group balances correctly; do you get an error back from ETEngine?

redekok:
> I would expect that ETModel would then ensure that the group balances correctly; do you get an error back from ETEngine?

False alarm! The share groups weren't defined correctly in my local admin of ETModel. The inputs are balancing well now.

PR #2072, opened 2019-07-04
Change min and max restrictions of number of residences input

redekok:
I've changed the minimum and maximum restrictions of the `households_number_of_residences` input, respectively to `0.0` and `4 * original number of residences`. Usually, we do not really impose restrictions on our users. This makes it possible to create unrealistic scenarios (for example, by "destroying" all residences). Furthermore, I wasn't sure if doubling the number of residences would be a sufficient maximum for local regions that have huge (housing) development plans.

### DIFF
--- a/inputs/demand/households/households_misc/households_number_of_residences.ad
+++ b/inputs/demand/households/households_misc/households_number_of_residences.ad
@@ -36,8 +36,8 @@
   )
 
 - priority = 0
-- max_value_gql = present:AREA(number_of_residences) * 2
-- min_value_gql = present:AREA(number_of_residences) * 0.5
+- max_value_gql = present:AREA(number_of_residences) * 4
+- min_value = 0.0
 - start_value_gql = present:AREA(number_of_residences)
 - step_value = 0.1
 - unit = #


ChaelKruip:
> I've changed the minimum and maximum restrictions of the households_number_of_residences input, respectively to 0.0

I recall that this can lead to division-by-zero problems in inputs and queries that use the `households_number_of_residences` in a division. Have you checked this @redekok ?

marliekeverweij:
I took the same restrictions as the population-slider (`min = 0.5 * original number of residences` and `max = 2 * original number of residences`). First I also put the minimum on `0.0 * original number of residences`, but then the sliders failed. I expect this is because the useful demand is scaled relative to the previous useful demand. When the previous useful demand is zero, then the future useful demand will remain zero! But maybe you can check this again?

I actually also think that it is not needed to put the minimum on `0.0 * original number of residences`. But changing the maximum to `max = 4 * original number of residences` is fine for me!

redekok:
> I recall that this can lead to division-by-zero problems in inputs and queries that use the households_number_of_residences in a division.

Good point. However, in the old situation the user was also able to set all numbers of apartments, etc. to zero, resulting in a total number of residences of zero as well.

> Have you checked this @redekok ?

> But maybe you can check this again?

I've checked this by running the model locally, which seemed to be working fine. Besides that, I couldn't find any inputs or gqueries that use the `households_number_of_residences` in a division.

ChaelKruip:
> I've checked this by running the model locally, which seemed to be working fine. Besides that, I couldn't find any inputs or gqueries that use the households_number_of_residences in a division.

👍

Still, I'm curious to know why this was a problem before and not anymore... We have had a big overhaul of the inputs for houses so perhaps this has solved a few issues with divisions through zero. I'll merge this but please test it well on BETA!

PR #2091, opened 2019-07-10
Update query for new supply slider units (MW instead of NoU)

redekok:
Related to https://github.com/quintel/etflex/issues/453


antw:
In my local ETFlex, the etflex_score query is always returning 0 no matter how I set the sliders.

redekok:
> In my local ETFlex, the etflex_score query is always returning 0 no matter how I set the sliders.
Good catch! I think I forgot to adjust the factors in the score calculation.

Issue #2085, opened 2019-07-05
(How) Should we increase power plant slider maxes?


michieldenhaan:
**Problem**
As pointed out by @MartLubben, in some local datasets the slider max for power plants is too restrictive to display installed capacity in the present. This is generally the case when a large power plant is present in a small region. E.g. one local dataset in Groningen has a 1.8 GW coal plant but the slider only goes up to 500 MW, suggesting that the area only has 500 MW installed.

**Current max**
The current slider max is based on electricity consumption in the present. For most power plants the slider max equals 0.5 times current electricity consumption.

**Proposal for new max**
Let slider max be determined by the _maximum_ of electricity consumption and production. In that way the slider max can never be too small to display installed capacity in the present. To ensure that the user can always increase installed capacity, we can set the slider max equal to 2-3 times present electricity consumption/production (whichever is bigger).

The query for this could look like this:
```present:PRODUCT(DIVIDE(MAX(Q(total_electricity_consumed)*3, Q(total_electricity_produced)*3), V(energy_power_ultra_supercritical_coal,maximum_yearly_electricity_production_per_unit)), V(energy_power_ultra_supercritical_coal, electricity_output_capacity))```

Pros:
- Slider will always correctly display installed capacity
- Slider max will probably never be too restrictive

Cons:
- Slider max may be unrealistically high

@ChaelKruip What do you think about this proposal?
@marliekeverweij Shall I pick this up before or after the deploy?

MartLubben:
I would like to let it happen before the deploy, because W&B is waiting for it with the Hogeland dataset and it is a bug.

ChaelKruip:
> Let slider max be determined by the maximum of electricity consumption and production.

As with Groningen, the local power plant may actually be much larger than the local energy consumption/production. Perhaps it is more robust to set the max to 2 typical power plants? 

>Slider max may be unrealistically high

This downside might be even worse though. 

> What do you think about this proposal?

I think it is probably an improvement although for the long term we might consider something out of the box like allowing the user to fill in a number outside of the slider range...

michieldenhaan:
> the local power plant may actually be much larger than the local energy consumption/production

What I meant is that if we allocate a 1.8 GW plant to an area, then setting the slider max equal to local production will ensure that the max will be at least 1.8 GW

> Perhaps it is more robust to set the max to 2 typical power plants?

Problem is that for larger areas/countries this is probably not enough.

> This downside might be even worse though.

[...]

PR #2093, opened 2019-07-12
Updated slider max for electricity and heat plants

michieldenhaan:
The slider max for electricity and heat plants now takes into account installed capacity of the plants in the starting year. In this way the slider max can never be too restrictive to display installed capacity in the present. 

- For electricity plants the slider max is now equal to (the installed capacity required to supply) total electricity demand in the present, or twice the installed capacity of the plant in the present (whichever is larger). 

_Example 1: Region X has a yearly electricity demand of 100 PJ in the present and no domestic production. The slider max for all power plant sliders (except sliders with 'land' potential like wind and solar) equals the capacity required to produce 100 PJ_

_Example 2: Region X has a yearly electricity demand of 10 PJ and a supply of 20 PJ because of a 200 MW coal plant in the region. The slider max for all power plant sliders except the coal plant equals the capacity required to produce 10 PJ (like in example 1). For the coal plant the slider max equals the capacity required to produce 40 PJ (twice current installed capacity), ensuring that the slider max is large enough to display current installed capacity and to allow for expansion of the current installed capacity_

- For heat plants the slider max equals half the (installed capacity required to supply) total heat production in the present (both collective heat and individual heat), or twice the installed capacity of the plant in the present  (whichever is bigger). 

Closes https://github.com/quintel/etsource/issues/2085

### DIFF
--- a/inputs/supply/electricity/coal_plants/capacity_of_energy_chp_ultra_supercritical_coal.ad
+++ b/inputs/supply/electricity/coal_plants/capacity_of_energy_chp_ultra_supercritical_coal.ad
@@ -1,17 +1,17 @@
 - query =
     EACH(
         UPDATE(
-            V(energy_chp_ultra_supercritical_coal), 
-            number_of_units, 
+            V(energy_chp_ultra_supercritical_coal),
+            number_of_units,
             USER_INPUT() / V(energy_chp_ultra_supercritical_coal, electricity_output_capacity)
         ),
         UPDATE(
-        	L(energy_chp_ultra_supercritical_coal), 
-        	preset_demand_by_electricity_production, 
+        	L(energy_chp_ultra_supercritical_coal),
+        	preset_demand_by_electricity_production,
         	V(energy_chp_ultra_supercritical_coal, production_based_on_number_of_units)
     ))
 - priority = 0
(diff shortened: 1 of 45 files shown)


ChaelKruip:
> For electricity plants the slider max is now equal to (the installed capacity required to supply) total electricity demand in the present, or twice the installed capacity of the plant in the present (whichever is larger).

This is limiting in the sense that the future electricity demand might increase and a region would like to provide this demand with a local plant. In #2085 the suggestion was made to take this approach:

> To ensure that the user can always increase installed capacity, we can set the slider max equal to 2-3 times present electricity consumption/production (whichever is bigger).

Although it may seem that I was critical of this idea, I actually prefer the wider range. 

> For heat plants the slider max equals half the (installed capacity required to supply) total heat production in the present

This is even more restrictive! I don't understand the asymmetry with respect to electricity.

@MartLubben I think that this PR was prematurely merged. As one of the assigned reviewers I would like to have the opportunity to review the changes. If there is a specific haste for this feature to make it to BETA feel free to consult me directly.

michieldenhaan:
> This is even more restrictive! I don't understand the asymmetry with respect to electricity.

My reasoning was that total heat production includes both collective and individual heat (in households, industry etc.). So the heat slider max will only be limiting in scenarios in which more than half of the future total heat demand will be collective and the user wants to supply this heat with one type of heat plant. I'm perfectly happy with increases the maxes tho.

Proposal: extend max for both heat and electricity to two times year demand?

ChaelKruip:
> Proposal: extend max for both heat and electricity to two times year demand?

Sounds good to me!

Issue #1744, opened 2018-06-18
Excess electricity overestimated?


michieldenhaan:
The amount of excess electricity available seems to go up when using flexibility options.
I've made a scenario with lots of wind. When using no flex options, the amount of electricity curtailed or export is 3271 PJ. When using flex options, ``energy_power_hv_network_electricity`` decreases by 827 PJ (see screenshot). The demand of  ``energy_production_excess_electricity``, however, goes up by 873. The difference is 5.28%, which is equal to the HV network loss.

![energytransitionmodel httpetengine test 2018-06-18 15-23-02 1](https://user-images.githubusercontent.com/32056448/41538625-0b977146-730c-11e8-92d5-8de1b045d2b4.png)

![energytransitionmodel httpetengine test 2018-06-18 15-23-25 1](https://user-images.githubusercontent.com/32056448/41538669-1d196488-730c-11e8-8d6c-6eb3af766e7d.png)

Is it an option to add an edge between ``energy_power_hv_network_electricity`` and ``energy_production_excess_electricity``?

michieldenhaan:
I think I figured out what's going on. See below for an example using no flex and using full flex:

**Scenario: NL 2016, 10.000 offshore wind turbines, no flex**

Total electricity production: **536.4 PJ**

Total used in sectors (transformer + own use): 340.16 PJ
Losses on electricity used in sectors: 18.96 PJ
Export/curtailment: 167.96 PJ
Losses on export/curtailment: 9.4 PJ
Total electricity demand: **536.4 PJ**


**Scenario: NL 2016, 10.000 offshore wind turbines, 12,279 power-to-hydrogen units**

Total electricity production: **536.4 PJ**

Total used in sectors (transformer + own use): 340.16 PJ
Losses on total used in sectors: 18.96 PJ
Extra losses reported in ETM: 9.4 PJ
Excess electricity used by flex: 177.32 PJ
Total electricity demand: **545.79 PJ**

[...]

PR #2117, opened 2019-09-27
Add new flex options for iNET

redekok:
We’ve added:
- Nodes and edges
- Input statements
- Queries for relevant charts and tables

https://github.com/quintel/etsource/commit/fed645868689874f383b8b002469511766883ffc -> We have changed the location of the `energy_flexibility_pumped_storage_electricity`. First, it was connected to the HV network, now it is connected to the excess electricity node like all other flex options. I think it doesn’t matter for the performance of the flex option, but connecting the node to excess electricity is how it should be, right @antw ? (see also this comment https://github.com/quintel/etsource/issues/1744#issuecomment-428504470)


antw:
> but connecting the node to excess electricity is how it should be, right @antw ?

I'm not an expert on this. It seems like it should be connected be connected to "excess" with everything else, but as I don't know why it was originally attached to the HV network I cannot say for sure.

nodes/households/households_flexibility_p2p_electricity.converter.ad
```diff
@@ -3,7 +3,7 @@
 - output.electricity = 0.85
 - output.loss = elastic
 - groups = [cost_flexibility, lv_net_flexibility, p2p]
-- merit_order.group = power_to_power
+- merit_order.group = household_batteries
```
antw: I'll need to write a migration in ETEngine to update existing scenarios to use this new name.

marliekeverweij: Okay, good point! Forgot about that.

marliekeverweij:
> I don't know why it was originally attached to the HV network 

I think it was connected to the HV accidentally...

Issue #2120, opened 2019-09-27
CCS-slider should also capture carbon emitted by cyclone furnace (Hisarna)


marliekeverweij:
As Berenschot correctly pointed out, it's currently not possible to capture the carbon emitted by a cyclone furnace (Hisarna) in the steel sector.

PR #2121, opened 2019-09-30
Improving CCS slider

marliekeverweij:
Fixing https://github.com/quintel/etsource/issues/2120

I first thought that the industrial CHPs generating the `steam_hot_water` for the Hisarna were not added to the input statement, but that is the case: https://github.com/quintel/etsource/blob/master/inputs/ccs/industry_ccs_in_industry.ad

Then I thought that is had something to do with these nodes (as we already thought):
Screen Shot 2019-09-30 at 10 08 22
But the `free_co2_factor` of the node `industry_final_demand_coal_gas` is already added to the input statement. 

Then I thought that the steam hot water from the central heat network still had CO2 emissions allocated, so I added the node `industry_locally_available_steam_hot_water` to the input statement. Kind of works! 

Now all CO2 emissions of industry are almost capture, there is 0.25937 MT left in a Dutch scenario with 100% hisarna. Where it this CO2 coming from?

Notifying @AlexanderWirtz


marliekeverweij:
Thanks for your feedback, I've updated the PR. You're right: it should not be possible to capture all industrial CO2, only on-the-spot.

Issue #2094, opened 2019-07-22
Households biogas CHP should not be in both must-run and dispatchable groups


ChaelKruip:
[This line](https://github.com/quintel/etsource/blob/master/nodes/households/households_collective_chp_biogas.central_producer.ad#L7) contains mutually exclusive groups

[The same goes for Agriculture](https://github.com/quintel/etsource/blob/master/nodes/agriculture/agriculture_chp_engine_biogas.central_producer.ad#L7)

marliekeverweij:
I guess the CHPs should only have the `must_run_electricity_production` group and not the `dispatchable_production` group, right?

PR #2126, opened 2019-10-07
Removing dispatchable_production group from local CHPs

marliekeverweij:
Closing https://github.com/quintel/etsource/issues/2094 : removed the `dispatchable_production` group from local CHP nodes.

Question: the node [`energy_chp_supercritical_waste_mix.central_producer`](https://github.com/quintel/etsource/blob/master/nodes/energy/energy_chp_supercritical_waste_mix.central_producer.ad) also contains both groups, should this node also only contain the `must_run_electricity_production` group?

### DIFF
--- a/nodes/agriculture/agriculture_chp_engine_biogas.central_producer.ad
+++ b/nodes/agriculture/agriculture_chp_engine_biogas.central_producer.ad
@@ -4,7 +4,7 @@
 - output.electricity = 0.42
 - output.loss = elastic
 - output.steam_hot_water = 0.35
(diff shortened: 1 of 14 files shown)


michieldenhaan:
> the node energy_chp_supercritical_waste_mix.central_producer also contains both groups, should this node also only contain the must_run_electricity_production group?

I think so, given that this node has ``merit_order.subtype = must_run``. Most energy sector power plants are dispatchable tho, maybe @AlexanderWirtz can give the final call? I'll remove the ``dispatchable`` group for now due to the deploy deadline but we can change it back later

PR #2142, opened 2019-11-08
iNET deliverable 3, 4, 5 + 6

marliekeverweij:
- weather years
- hourly gas demand, supply and storage
- hybrid heat in industry
- hydrogen in other industry


redekok:
> Is this intentional?

Yes, this was intentional, although I know it's not the optimal way to go.. The issue is that we cannot symlink the entire `curves` directory anymore (as we used to do with the former `heat` directory), since the `curves` directory includes both curves that should be symlinked to the NL ones (e.g., the `insulation_*` curves) and curves that are country specific (and shouldn't be symlinked). Due to a lack of time, I just copy-pasted the entire `curves` directory from `nl` and replaced the curves that are dataset-specific. Not sure if we can easily fix this in order to restore the symlinks where necessary?

antw:
> Not sure if we can easily fix this in order to restore the symlinks where necessary?

I've removed a lot of the duplicates and replaced them with symlinks. I think we can also remove the solar and wind curves from the `curves/` directories? I expect we're only using the versions which are now in `curves/weather/___/`?

DorinevanderVlies:
>  I've removed a lot of the duplicates and replaced them with symlinks.

These duplicate curves do indeed seem unnecessary. 

>  I expect we're only using the versions which are now in curves/weather/___/? I think we can also remove the solar and wind curves from the curves/ directories?

If the curves for countries other than NL are symlinked to the country specific curves and if for countries other than NL curves from `weather years - default` are used that seems ok!

PR #2150, opened 2019-11-18
Add new queries for total co2 without non-energy emissions

noracato:
See etmodel branch co2_sheets_toggle_other_emissions

### DIFF
--- a/gqueries/modules/co2_sheet/sector_totals/co2_sheet_agriculture_total_co2_emissions_only_energetic.gql
+++ b/gqueries/modules/co2_sheet/sector_totals/co2_sheet_agriculture_total_co2_emissions_only_energetic.gql
@@ -1,8 +1,8 @@
 - query =
-          SUM(
-              Q(co2_sheet_agriculture_heat_co2_emissions),
-              Q(co2_sheet_agriculture_power_and_light_co2_emissions)
-              )
+  SUM(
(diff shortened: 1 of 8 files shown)


gqueries/modules/co2_sheet/sector_totals/co2_sheet_agriculture_total_co2_emissions_only_energetic.gql
```diff
@@ -0,0 +1,8 @@
+- query =
+          SUM(
+              Q(co2_sheet_agriculture_heat_co2_emissions),
+              Q(co2_sheet_agriculture_power_and_light_co2_emissions)
+              )
```
antw: I'm nitpicking, and I know you've just copied the code style from the other queries so this isn't your fault at all, but the indentation is a bit off:

* After the `- query` line, the query should typically be intended with four spaces.
* The closing bracket from `SUM` should be aligned with the SUM.

```diff
- - query =
-           SUM(
-               Q(co2_sheet_agriculture_heat_co2_emissions),
-               Q(co2_sheet_agriculture_power_and_light_co2_emissions)
-               )
+ - query =
+     SUM(
+         Q(co2_sheet_agriculture_heat_co2_emissions),
+         Q(co2_sheet_agriculture_power_and_light_co2_emissions)
+     )
```

Ditto for the other queries.

PR #2165, opened 2019-12-13
WACC sliders!

marliekeverweij:
These new WACC - categories and percentages have been implemented (reasoning can be found on BC):
- huishoudens: 2%
- publieke infrastructuur: 3%
- commercieel + bewezen technologieën: 4% 
- nieuwe/niet-bewezen/onvolwassen technologieën = 7% 

Before it was:

Screen Shot 2019-12-13 at 11 32 32

The allocation of groups to nodes can be found in the Excel on DB: Projects/201909_iNet_ETM_verbeteringen/Notes/20191205_wacc_nodes

I want to add a new sidebar item to "Costs" for the WACC sliders, since it doen't fit in any of the existing sidebar items.

Screen Shot 2019-12-13 at 11 24 06


marliekeverweij:
> The inputs update the WACC based on USER_INPUT(). Do we need to divide this by 100 as the input is a number between 0-100 and the WACC is a number between 0-1? Or does ETengine do this automatically?

Whoops!!! You are right

> There's three nodes that currently have a WACC of 0:
transport_truck_using_hydrogen
transport_car_using_hydrogen
transport_car_using_electricity
Do we want to add those to one of the categories? 

Good point. Yes, I think these should be added to a WACC group. Shall we add the hydrogen car to `wacc_unproven_tech` and the electric car to `wacc_proven_tech`? 
The truck does not have investment costs and there also is no slider to add costs, so I propose to leave that.

marliekeverweij:
FYI: the total costs for current scenarios will be lower because of the lower WACC percentages. For example for NvdT Internationiaal the difference is:
BETA = 58.2 billion
Local (new WACC%) = 55.9 billion

PR #2175, opened 2020-01-09
Fixing gquery so scenario report works again

marliekeverweij:
Closes https://github.com/quintel/etmodel/issues/3211#issuecomment-572444570

I do not really understand why this error pops up now. The scenario report and failing gqueries haven't been updated recently.... Maybe this error was there already for some time, but nobody opens the Dutch scenario report?!

Anyway, this fixes the issue. `co2_sheet_non_energy_emissions_industry` does not exist, but `co2_sheet_non_energy_emissions_industry_energy` does, so I suppose this gquery should go here!

Notifying @MartLubben @antw


jorisberkhout:
Assigning @MartLubben to do the review.

PR #2177, opened 2020-01-10
Curves import

### LINKED ISSUE #2149: The import:curves and symlink_curves tasks should import and symlink the weather curves correctly
The rake tasks to import curves from ETDataset to ETSource should be updated for the new weather conditions feature. Curve sets for weather years (including `weather_properties.csv`) should be imported and symlinked correctly for all datasets on ETSource.

noracato:
closes #2149 
Please run it a few times to see if it's behaving like you etsource experts would expect

To get the etlocal export based on nl2016 or nl2017 please run the symlink_nl task to mimick a 'full' dataset and add a full.nl2016.ad file with etengine=false


noracato:
>     * How are the local datasets symlinked? Do we still need the `symlink_curves` task? If so, we might want to include conditions to check 1) whether all local datasets on ETSource are included in the `local_curves.yml` and 2) whether there are no non-existing datasets listed in this yaml.

With the renewed export task in etlocal, you can set the `country` for a regional dataset (eg `nl2017`). If you are exporting **a not yet existing** regional dataset, it will automatically get symlinked and scaled to its `country`. If your regional dataset is **already present** in `etsource`,  only the area attributes and (sparse) graph values are updated.

I can try to change this so a brand new version (including scaling and symlinking) will be created on every export from etlocal? Or maybe with an optional 'REBASE' flag to set on the export task? Then the `symlink_curves` task will become unneccesary. 

> I noticed that the antwerpen and BEGM11002_antwerpen datasets have a curves directory that is symlinked to be. However, the BEBU4402101_muide_meulestede dataset is symlinked to nl (where it should be symlinked to be as well).

I'm unsure if my change has caused this. However, the above proposition should fix this as well, as you only have to set the `country` to be in etlocal to keep it symlinked to be on future updates.

redekok:
Thanks for all the fixes @noracato, LGTM! I've pushed some minor changes (the CSV check we discussed and some updates of the yaml files). Besides that, I've also pushed the dataset changes after running the renewed import:curves task. However, this results in failing specs.. @noracato would you have time to look into this?

PR #2187, opened 2020-01-22
Must-run / dispatchable toggle for nuclear plants

marliekeverweij:
Adding input statement for must-run / dispatchable toggle for nuclear plants. Default setting is "dispatchable" for nuclear plants, however it's apparently not realistic that nuclear plants generate with a load factor <50%. Therefore the possibility is added to make them "must-run"

Screen Shot 2020-01-22 at 15 49 53
Screen Shot 2020-01-22 at 15 49 31

Notifying @ChaelKruip


ChaelKruip:
Nice work @marliekeverweij !

marliekeverweij:
Thanks! And @antw  🙌

Issue #2191, opened 2020-01-23
Output capacities are inconsistent for energy_chp_combined_cycle_network_gas


michieldenhaan:
Current specs of energy_chp_combined_cycle_network_gas:
``- output.electricity = 0.42``
``- output.steam_hot_water = 0.43``
``- electricity_output_capacity = 120.0``
``- heat_output_capacity = 117.2093``

When calculating the input capacity of this plant, we can either do:
120.0 / 0.42 = 285.7 MW
or
117.2093 / 0.43 = 272.6 MW
Which gives different results

Maybe the output capacity calculation in ETDataset is refering to the wrong cells. If we do:
120.0 / 0.43 = 297.1 MW 
117.2093 / 0.42 = 279.1 MW (swap the output efficiencies)

Might be the case for more nodes.

ChaelKruip:
Good catch @michieldenhaan! Perhaps we can make a spec for this @antw @noracato ?

marliekeverweij:
Putting this issue on hold for after the heat project merge. See: https://github.com/quintel/etsource/pull/2197#issuecomment-580219366

PR #2197, opened 2020-01-24
Update/correct CHP specs with ETDataset analysis

marliekeverweij:
Closing https://github.com/quintel/etsource/issues/2191#issuecomment-577654316

And more! I've exported the node source analyses of all CHPs and checked the efficiencies
Main things:
- efficiencies of network gas CCGT CHP (in industry & energy) were not consistent with source analysis
- industry wood pellets CHP had the same specs as `industry_chp_turbine_gas_power_fuelmix`. the node source analysis was up-to-date, so exported that to ETSource 
- some costs were not consistent with ETDataset


michieldenhaan:
I'm a bit hesitant to merge this as changing CHP efficiencies also impacts the dataset analysis (installed capacities etc.). As this seems to be a problem that has been present for a while, maybe we can put this on hold for now and pick this up later (e.g. after we deployed the heat project)?

PR #2198, opened 2020-01-28
Update network gas data export

redekok:
- Add breakdown of the metals sector into steel, aluminium and other metals to the network gas data export


michieldenhaan:
> Question: do we need to adjust queries for time-series charts or does this happen automatically?

I don't think so, the query filters on all consumer types in the industry sector: https://github.com/quintel/etsource/blob/master/gqueries/general/network_gas/demand/industry_network_gas_input_curve.gql

PR #2207, opened 2020-01-31
Remove final demand hydrogen from CCS slider

marliekeverweij:
A client pointed out that final demand hydrogen should be removed from this slider, since there's no CO2 emitted when burning hydrogen in industry and this slider only captures CO2 in industry. So CO2 emitted during hydrogen production will now not be captured with this slider!


redekok:
LGTM! Shouldn't we adjust the query [`co2_captured_by_ccs_in_industry`](https://github.com/quintel/etsource/blob/master/gqueries/general/co2/co2_captured_by_ccs_in_industry.gql) then as well?

marliekeverweij:
Yes good point! I'll adjust it

marliekeverweij:
Fix it and found a node that was missing in that query :)

redekok:
Nice one!

Issue #2214, opened 2020-02-13
Weather conditions feature is broken on BETA


michieldenhaan:
I get this error when trying to change weather curves:

![beta-pro energytransitionmodel com - Energy Transition Model - Your free, independent, comprehensive, fact-based scenario build… 2020-02-13 09-08-34](https://user-images.githubusercontent.com/32056448/74413827-753c9700-4e40-11ea-9269-b8155104f008.png)

Caused by this commit? https://github.com/quintel/etsource/commit/4333023a05c91243c6f2b98bb4527f30b4772802

redekok:
I'm on it! I'm partly fixing it manually right now so it would be nice if you could take a look at https://github.com/quintel/etsource/issues/2216 to make sure the import:curve task works well in the future @noracato. :-)

Issue #2148, opened 2019-11-15
The weather_properties.csv is not correct for default curve sets in most datasets


redekok:
Many of the curves in the default curves set for non-NL country datasets are symlinked to the `nl` ones. This is also the case for the `weather_properties.csv` file, which is not correct because the FLH should be country-specific.

marliekeverweij:
@antw , is `weather_properties.csv` used when users click on the default weather year or are the FLH then derived from the `central_producers.csv`? Maybe we could delete it from the weather/default folder, since it now contains `nl` values (2015) in all (country) datasets and that is not correct

antw:
> @antw , is weather_properties.csv used when users click on the default weather year

Yes it is. An alternative might be to wrap the whole "settings_weather_curve_set" input in an `IF` so that it isn't executed if the user reverts back to the default settings.

```ruby
IF(
  EQUALS(USER_INPUT(), "default"),
  -> {},
  -> {
      UPDATE(AREA(), weather_curve_set, USER_INPUT());

      x = WEATHER_PROPERTY(average_difference_in_air_temperature);
      heat_factor = 1.0 - 0.0893386 * x;
      cool_factor = 1.0 + 0.5669449 * x + 0.1598105 * x**2 + 0.0253755 * x**3 + 0.00163210 * x**4;

      # ...
  }
)
```

redekok:
> An alternative might be to wrap the whole "settings_weather_curve_set" input in an IF so that it isn't executed if the user reverts back to the default settings.

This solution might be prettier than having the default FLH values documented in two places (both `central_producers.csv` and `weather_properties.csv`). What do you think @marliekeverweij?

marliekeverweij:
I also like this solution!

However, I wonder how the curves and FLH will be updated to default with this `IF`. For example when users first choose 1987 and then want to return to default, I expect that with this `IF` that the FLH and curves will not be updated to default but still be 1987. Is that true @antw ?

antw:
> For example when users first choose 1987 and then want to return to default, I expect that with this IF **the FLH and curves will not be updated to default** but still be 1987. Is that true @antw ?

Inputs always change *default* values in the graph to *something else*. If we skip the input with an IF, the graph will therefore retain the defaults.

redekok:
[...]

PR #2215, opened 2020-02-13
Add weather properties for NL datasets

redekok:
Related to https://github.com/quintel/etsource/issues/2214

### DIFF
--- a/datasets/be/curves/weather/default/weather_properties.csv
+++ b/datasets/be/curves/weather/default/weather_properties.csv
@@ -1,0 +1,1 @@
+../../../../../datasets/nl/curves/weather/default/weather_properties.csv
--- a/datasets/br/curves/weather/default/weather_properties.csv
+++ b/datasets/br/curves/weather/default/weather_properties.csv
@@ -1,0 +1,1 @@
+../../../../../datasets/nl/curves/weather/default/weather_properties.csv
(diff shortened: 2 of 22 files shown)


antw:
> Merge after #2148 has been solved.

I don't think #2148 is related.

The tests are failing because the pull requests adds a weather_properties.csv for NL, but other full-size country datasets do not have one. It looks like #2177 removed the weather_properties.csv (accidentally?) for all datasets.

redekok:
> It looks like #2177 removed the weather_properties.csv (accidentally?) for all datasets.

Yes, it accidentally did.

> The tests are failing because the pull requests adds a weather_properties.csv for NL, but other full-size country datasets do not have one.

However, as discussed in https://github.com/quintel/etsource/issues/2148, I would be in favour of only having a weather_properties.csv file for weather years curve sets. For default weather curve sets (in full-size country datasets other than NL), it would be fine to read the FLH from the central producers table.

> An alternative might be to wrap the whole "settings_weather_curve_set" input in an IF so that it isn't executed if the user reverts back to the default settings.

Would this be the solution for the situation proposed above?

For now, I will add (the symlinks to) the weather_properties.csv again for country datasets other than NL in order to fix BETA.

antw:
> However, as discussed in #2148 [...]

I see! The tests assert that if NL has a curve or CSV file (in this case, default/weather_properties.csv) then the other datasets should also have a file of the same name; that's why they failed for this PR.

Indeed you are correct: if we remove the need for a default weather_properties.csv file, the other datasets will also no longer need this file.

PR #2224, opened 2020-02-21
Update solar PV potential FoodValley region

marliekeverweij:
@redekok is it correct that this line is added to the dataset?
`has_climate = false`


redekok:
> @redekok is it correct that this line is added to the dataset?
has_climate = false

The `has_climate` attribute has been replaced by the `has_weather_curves` attribute. Therefore, the `has_climate` line doesn't do anything at the moment. You could remove it from the dataset.

PR #2228, opened 2020-02-24
Set industry non_specified edges to flexible

michieldenhaan:
This is now possible thanks to https://github.com/quintel/refinery/commit/72eacf8


michieldenhaan:
No don't think so!

PR #2231, opened 2020-02-26
Heat fixes

michieldenhaan:
- Add electricity input to industry geothermal
- Update preset demand in industry and local energy CHP input statements
- Update geothermal COP based on ECN ORT data


marliekeverweij:
Looking good!

PR #2241, opened 2020-03-06
Set heat distribution losses to 0 for datasets with missing key

michieldenhaan:


marliekeverweij:
Quick fix, nice! Are we sure we're not missing more keys?

PR #2245, opened 2020-03-11
Update and add Overijssel datasets

redekok:
- Update industry demands for the Overijssel municipality, RES and province datasets, yet existing on ETSource 
- Add remaining Overijssel municipality datasets to ETSource

@michieldenhaan Is it correct that the AREA-attributes for district heating have been changed over the last couple of weeks? 

image


redekok:
> The block heating is a bug, the value should be between 0-1. I think the aggregator on ETLocal currently takes the sum, this should be a (weighted) average.

That makes sense. @michieldenhaan Do you think we can preserve the old values (for instance 0.15 for RES Twente) for this attribute? 

> I think we should go with the new data as the differences appear to be small. Agree?

Agreed!

PR #2249, opened 2020-03-20
P2H boiler and heat pump for district heating

marliekeverweij:
Screen Shot 2020-03-20 at 16 46 27


config/flexibility_order.yml
```diff
@@ -21,5 +21,7 @@
 - pumped_storage
 - power_to_gas
 - power_to_heat_industry
+- power_to_heat_district_heating_boiler
+- power_to_heat_district_heating_heatpump
```
antw: Is it definitely the case that a user will want to prioritise one of these over the other, rather than just having a "Power to heat for district heating" option? For industry we just have the one option for four technologies, and excess electricity is divided fairly between them based on the capacity of each.

marliekeverweij: Yes that's the case!

PR #2264, opened 2020-04-16
Add efficiency slider for solar PV

michieldenhaan:
Allows users to set efficiency of solar PV panels. This affects solar PV in households, buildings, energy_power and energy_hydrogen.

- For households and buildings, increasing efficiency leads to a higher electricity production of solar PV and a higher installed capacity. Number of units stays constant. Demand (= input of ``solar_radiation``) stays constant. The reasoning behind this is that users set a share of roof surface covered with PV. If PV becomes more efficient, the same number of panels produce more electricity with the same amount of radiation
- For energy_power and energy_hydrogen, increasing efficiency leads to a lower number of installed units / lower demand of solar radiation. Produced electricity and installed capacity stay constant. The reasoning behind this is that users set the amount of installed capacity. If PV becomes more efficient, less panels are needed for the same amount of production/installed capacity.

Goes with: https://github.com/quintel/etmodel/pull/3320
Closes: https://github.com/quintel/etmodel/issues/1893

SO to @antw for his help! 🚀

### DIFF
--- a/inputs/costs/technical_solar_pv_efficiency.ad
+++ b/inputs/costs/technical_solar_pv_efficiency.ad
@@ -1,17 +1,14 @@
 # This updates the efficiency of solar PV. This affects the output conversion,
-# output capacity per panel, and, for households and buildings, total electricity production.
+# output capacity per panel, and, for energy_power and energy_hydrogen, number of units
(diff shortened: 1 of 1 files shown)


ChaelKruip:
Great work @michieldenhaan and @antw! 👏

PR #2272, opened 2020-05-01
Add 2017 datasets for Zuid-Holland

redekok:
- 2017 data and profiles for municipality and RES-region datasets
- 2017 data and 2015 profiles for province dataset


DorinevanderVlies:
@michieldenhaan , for consistency with II3050 we want to use 2015 curves for SSZH.
Do you agree that we combine 2015 curves and 2017 data for the ZH-dataset?

michieldenhaan:
> Do you agree that we combine 2015 curves and 2017 data for the ZH-dataset?

Yes!

PR #2273, opened 2020-05-01
Mimick full datasets

noracato:
- Create task mimick_full_dataset to symlink all missing files and create a mimicked full.ad file for datasets with different analysis years.
- Create specs that check for changes in the original full files to keep the mimicked files up to date
- Mimick the datasets nl2016 and nl2017

Goes together with quintel/etlocal#231


redekok:
🥳

Issue #2274, opened 2020-05-01
Rake symlink_curves should be adapted


michieldenhaan:
The rake task ``symlink_curves`` is largely obsolete now that ETLocal automatically symlinks the correct curves when exporting a dataset to ETSource (https://github.com/quintel/etlocal/pull/231). This means that ``local_curves.yml`` is no longer needed (🥳).

As @redekok pointed out, the ``symlink_curves`` task can still be useful in case you want to symlink a region to another year. E.g. use 2015 curves for some 2017 regions to make sure the curves match with those used in other scenarios/studies.

To make this possible, the current rake task should be adapted in the following way:

- [ ] Make it possible to specify a region and desired 'curve' year in the terminal rather than in ``local_curves.yml``
- [ ] Remove ``local_curves.yml``

PR #2279, opened 2020-05-18
Adapt symlink_curves task

redekok:
Fixes https://github.com/quintel/etsource/issues/2274


noracato:
👍 👍 👍 
Thanks @redekok! I removed the each loop that is now redundant (we were looping over 1 item). All set to merge now 😄

redekok:
Haha, good point. Thanks! 🤓

PR #2282, opened 2020-05-26
Use 2018 parameter file to generate G2A curves

michieldenhaan:
Previous curves were based on 2021 parameter file provided by NEDU. Changed to 2018 to ensure curves are aligned with upcoming 2018 update


DorinevanderVlies:
Nice!

For reference: https://github.com/quintel/etdataset/commit/543337592f9b667f73d96caeb0b1411f88acefd5

PR #2310, opened 2020-06-26
Electricity sankey: remove solar pv curtailment from flexibility

marliekeverweij:
See https://github.com/quintel/etmodel/issues/3421


marliekeverweij:
@antw could you maybe also merge this fix with production? :)

PR #2320, opened 2020-07-10
Updated presets II3050

MartLubben:


marliekeverweij:
This is the scenario list in my local model:
Screenshot 2020-07-10 at 15 09 21
If you have suggestions for improving the titles and ordering let me know @everybody!

antw:
FYI, while preset titles cannot be localised presently, you can add locale-specific *descriptions* like so:

```
= descriptions =
  <div class="nl">
    Dutch description...
  </div>

  <div class="en">
    English description...
  </div>
```

marliekeverweij:
@antw two questions:
- I've just tried to add the links to a Refman-document, but they don't show up as a link. what's going wrong?
- the scenario title on the scenario page now says II3050 Europese CO<sub>2</sub>-sturing and not the 2 as subscript. is that possible or should we not try? ;) 

Screenshot 2020-07-10 at 15 31 16

antw:
> I've just tried to add the links to a Refman-document, but they don't show up as a link. what's going wrong?

Links to a different *host* are stripped. When deployed, the host is "enegytransitionmodel.com", but locally it will be something like "localhost" or "etmodel.dev". The links should show up when deployed.

> the scenario title on the scenario page now says II3050 Europese CO2-sturing and not the 2 as subscript. is that possible or should we not try? ;)

We allow a subset of safe HTML tags in descriptions, but titles should not contain any. HTML tags are escaped to prevent users from injecting harmful tags.

I suggest changing the title to "CO2" for now (no `<sub>`) and we can handle this in ETModel by rendering it with the subscript.

marliekeverweij:
Closing as presets are now on PRO!

PR #2342, opened 2020-08-31
Update dataset shares

redekok:
- Update shares in edge files
- Distinguish energy and molecules shares in country datasets (the molecules dir is still empty and therefore can't be pushed to git yet)
- Update path to energy shares directory in the import dataset task


michieldenhaan (changes requested):
The comma between the file name and row name is missing I think. E.g.
`SHARE("energy/industry_final_demand_for_chemical_electricity_parent_share" industry_final_demand_for_chemical_refineries_electricity)`

should be:
`SHARE("energy/industry_final_demand_for_chemical_electricity_parent_share"COMMA industry_final_demand_for_chemical_refineries_electricity)`

redekok:
Ah you're totally right. Forgot about the comma. 🙈  I'm on it!

michieldenhaan (approved):

PR #2343, opened 2020-08-31
Add conversions to molecule nodes with energy counterpart

redekok:
@michieldenhaan I'm still getting some Atlas "CannotParseError" btw, when trying to load ETEngine on my local machine:

image


michieldenhaan (changes requested):
The syntax should be carrier: co2_conversion_per_mj, not the actual carrier name. See:
https://deploy-preview-57--etm-docs.netlify.app/contrib/molecules/#molecule-flow-is-based-on-a-carrier-attribute

E.g. `- from_energy.conversion.coal = carrier: co2_conversion_per_mj`
instead of
`- from_energy.conversion.coal = coal: co2_conversion_per_mj`

michieldenhaan (approved):

PR #2363, opened 2020-10-14
Set all curves to NL2017 except buildings appliances

DorinevanderVlies:
For buildings appliances the original curve for households appliances is used as Ameland has many small tourist houses in the buildings sector.


marliekeverweij:
The Ameland dataset loads locally and below you can see the household and buildings heat demand curves:

Screenshot 2020-10-14 at 12 09 35

marliekeverweij:
There are two semaphore errors for the sum of the households_appliances and buildings_appliances curves not being exactly 1/3600. Maybe some rounding has taken place
Screenshot 2020-10-14 at 12 11 37

marliekeverweij (approved):

PR #2373, opened 2020-11-03
Update emission factors UK and NI

marliekeverweij:


michieldenhaan:
I think we're good to go. Remove the old, unused FCE sliders. These sliders referred to countries like South Africa, Russia etc. causing the model to stop loading when these countries aren't explicitly specified in the FCE files.

Did a global search and only found some references to these sliders in ETModel, but they all seem to be hidden/disabled. I'll send a separate PR to remove these references.

michieldenhaan (approved):

PR #2376, opened 2020-11-10
Electricity storage technologies update

mabijkerk:
Updated etsource files of the five electricity storage technologies in the model: household battery, ev battery, large-scale battery, reservoirs, underground pumped storage. This was done to match the updates in the corresponding datasets that were performed in the etdataset#856 pull request.


michieldenhaan:
Question for @ChaelKruip: @mabijkerk researched new households P2P specs as the old specs were outdated and not fully documented. One striking difference is that the old volume per unit was quite high: 20 kWh compared to 10 kWh in the new specs. This was perhaps caused by a misinterpretation of the old source data. We think 10 kWh is a more realistic value for one household (at least in the foreseeable future). This change has an impact on existing scenarios: the user (implicitly) sets the installed number of units, so with the new value the available volume in scenarios effectively halves. We have roughly three options:
- Migrate scenarios, i.e. doubling the household battery share. For scenarios with a current slider setting of more than 50%, this can only be done to some degree as the slider cannot go above 100%.
- Alter the new specs such that one battery unit has 20 kWh volume instead of 10 kWh (i.e. double the volume, capacity, costs etc.). This means the energetic impact on scenarios is close to zero. 
- Do nothing (to a certain extent this can be seen as a 'bug fix')

Do you have a preference for one of these options?

ChaelKruip:
> Migrate scenarios, i.e. doubling the household battery share. For scenarios with a current slider setting of more than 50%, this can only be done to some degree as the slider cannot go above 100%.

Thanks for this clear overview @michieldenhaan! 

I prefer the first option as it tries to accommodate as much as possible the vision of a scenario which I think will be likely based on storage volume rather than capacity. This is a bit of a curtesy to the user in comparison to option three. Option two is interesting and tempting but I feel that it is hard to explain that we use a 'double' battery as the default. So, option one still has my preference.

PR #2382, opened 2020-11-13
Remove co2_conversion_per_mj from network gas carrier and use hard-coded factor for from_energy.conversion.network_gas

michieldenhaan:
Setting co2_conversion_per_mj on the network_gas carrier resulted in unintended side effects, e.g. increasing green gas no longer decreases the weighted_carrier_co2_per_mj of network gas. This again affects the amount of CO2 credits gas power plants have to pay and hence their position in the merit order.

This commit removes the co2_conversion_per_mj attribute from the network_gas carrier file. Molecule nodes using `from_energy.conversion.network_gas = carrier: co2_conversion_per_mj` now use a hard-coded conversion factor of methane (0.0564).

In the future, we may want to explore adding separate methods for on-the-spot emissions (including bio emissions) and IPCC emissions (relevant for CO2 credits).

Notifying @Charlottevm

### DIFF
--- a/carriers/network_gas.ad
+++ b/carriers/network_gas.ad
@@ -1,3 +1,2 @@
 - sustainable = 0.0
 - infinite = false
-- co2_conversion_per_mj = 0.0564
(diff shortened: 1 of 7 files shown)


antw (approved):
LGTM!

The `weighted_carrier_co2_per_mj` of energy_power_combined_cycle_ccs_network_gas now matches the value on production.

* Before: 0.0564
* After: 0.05630567652233583

PR #2388, opened 2020-12-10
Use weighted average instead of average

michieldenhaan:
Previously, this query returned a simple average. As a result, the emission slider of each interconnector affected the average emissions even for interconnectors with 0 capacity. This new query circumvents this by taking a weighted average, weighted by the actual import volume of each interconnector.


marliekeverweij:
Where is this query used? I do agree with the update :)

marliekeverweij (approved):

michieldenhaan:
In the 'domestic' CO2 dashboard item to subtract import emissions!

PR #2400, opened 2021-01-19
Update presentation_groups for solar and electricity production CCS

marliekeverweij:
Goes with: https://github.com/quintel/etengine/pull/1149


michieldenhaan (commented):
A few comments:
For some presentation_groups (like space heaters) it isn't clear what the unit of the costs is. Is it per unit, per MW, total costs? If possible I would make that explicit.
image

For solar PV the capacity per unit isn't showing correctly:
image

For wind turbines we show an electrical efficiency of 97%, what is this exactly? If possible I would leave this out.
image

For solar PV farms the land use per MW is not correct. It is a factor 1000 off? Also, I'm not a big fan of showing the capacity per unit as this is an arbitrary number for PV farms.
image

For CCS capacities, is it possible to change the unit to T CO2/hour instead of just T/hour? I think that's more clear.
image

Also for CCS, I'm not a big fan of showing attributes _per unit_, like solar farms 'units' are not very informative here. Maybe we can show this in terms of an hourly or annual capacity?
image

marliekeverweij:
Thanks Michiel!

marliekeverweij:
> For some presentation_groups (like space heaters) it isn't clear what the unit of the costs is. Is it per unit, per MW, total costs? If possible I would make that explicit.

I've added 'per unit' 

> For solar PV the capacity per unit isn't showing correctly

@antw do you know what's going on here? there is a `electricity_output_capacity` given in the node. Maybe it's to small to show?

> For wind turbines we show an electrical efficiency of 97%, what is this exactly? If possible I would leave this out.

I've removed this :)

> For solar PV farms the land use per MW is not correct. It is a factor 1000 off? 

Good one, @antw will have a look

> Also, I'm not a big fan of showing the capacity per unit as this is an arbitrary number for PV farms.

Me too, I removed it.

> For CCS capacities, is it possible to change the unit to T CO2/hour instead of just T/hour? I think that's more clear.

@antw is that possible? Otherwise we could add 'CO2' to the translation in the left column.

> Also for CCS, I'm not a big fan of showing attributes per unit, like solar farms 'units' are not very informative here. Maybe we can show this in terms of an hourly or annual capacity?

I've changed these to 'yearly storage/transport/capture capacity'

antw:
> For solar PV the capacity per unit isn't showing correctly:

I've fixed this by removing all number rounding in the ETEngine node serializer data. ETModel formats the numbers now, so there's no point in rounding numbers in ETEngine (and results in small numbers like this being rounded to zero).

Screenshot 2021-01-19 at 14 49 50

---

> For solar PV farms the land use per MW is not correct. It is a factor 1000 off? 

Fixed by changing the scaling of m2 to km2. Unfortunately the scaler still assumes that each increase in magnitude is 1000 times larger so it has a hard time scaling km2 down to m2. The result is that in this case it will show "0.015 km2" rather than "15,000 m2".

Screenshot 2021-01-19 at 14 51 34

---

> For CCS capacities, is it possible to change the unit to T CO2/hour instead of just T/hour? I think that's more clear.

Done.

Screenshot 2021-01-19 at 15 11 51

"tonne CO2" scales to "KT CO2" scales to "MT CO2", etc. Perhaps you'd prefer "Ktonne" and "Mtonne" (though I dislike those myself)?

@marliekeverweij Would you be open to doing something similar with "tonne" and "T"? We'd use "tonne" everywhere that the number is less than 1000, and kT, MT, etc when greater. Or would this be inconsistent?

marliekeverweij:
Great @antw!!

> @marliekeverweij Would you be open to doing something similar with "tonne" and "T"? We'd use "tonne" everywhere that the number is less than 1000, and kT, MT, etc when greater. Or would this be inconsistent?

Yes, that's perfect! (Now that "CO2" has been added, we could also to "T CO2" instead of "tonne CO2" if you like that better)

michieldenhaan (approved):

antw:
> Yes, that's perfect! 

Just so I'm sure we're on the same page, we'll change...

* ... the units in the serializer data back to "T" ("T / year", etc) *again*.
* ... the translations in ETModel so that T is shown as "tonne". KT, MT, GT, ..., will remain unaffected.

> (Now that "CO2" has been added, we could also to "T CO2" instead of "tonne CO2" if you like that better)

I think if we show "T" as "tonne", then "T CO2" should be "tonne CO2".

marliekeverweij:
Yes we're on the same page 👍

PR #2405, opened 2021-01-27
User curve fixes

michieldenhaan:
* Add custom FLH inputs for solar and wind that can be set regardless of the selected weather year. For solar, the `solar_pv_profile_1_share` is set to 1 to prevent interpolation
Question: do we have to make sure that the custom solar FLH input is applied after the regular one? To make sure `solar_pv_profile_1_share` is always set to 1?

* Update the curve config file to set both the original FLH sliders for wind and solar and the new custom inputs. The former is included in order to disable it, the latter to correctly update the FLHs/prevent interpolation

* Use `solar_pv_profile_1` rather than `solar_pv`. Idem for `wind_inland_baseline` vs. `wind_inland`. Same for coastal and offshore. I think this is the reason these curves were not updated correctly.

### DIFF
--- a/config/user_curves.yml
+++ b/config/user_curves.yml
@@ -180,13 +180,14 @@ network_gas_import:
   type: profile
   display_group: gas_import_export
 
-solar_pv:
+weather/solar_pv_profile_1:
   type: profile
   display_group: electricity_production
   reduce:
     as: full_load_hours
     sets:
       - flh_of_solar_pv_solar_radiation
+      - flh_of_solar_pv_solar_radiation_user_curve
 
 weather/solar_thermal:
   type: profile
@@ -212,34 +213,29 @@ river:
     sets:
       - flh_of_hydro_river
 
-wind_offshore:
+weather/wind_offshore_baseline:
   type: profile
   display_group: electricity_production
   reduce:
     as: full_load_hours
     sets:
       - flh_of_energy_power_wind_turbine_offshore
+      - flh_of_energy_power_wind_turbine_offshore_user_curve
 
-wind_coastal:
+weather/wind_coastal_baseline:
   type: profile
   display_group: electricity_production
   reduce:
     as: full_load_hours
     sets:
       - flh_of_energy_power_wind_turbine_coastal
+      - flh_of_energy_power_wind_turbine_coastal_user_curve
 
-wind_inland:
+weather/wind_inland_baseline:
   type: profile
   display_group: electricity_production
   reduce:
     as: full_load_hours
     sets:
       - flh_of_energy_power_wind_turbine_inland
-
-weather/air_temperature:
-  type: temperature
-  display_group: weather
-  reduce:
-    as: temperature
-    sets:
-      - flexibility_outdoor_temperature
+      - flh_of_energy_power_wind_turbine_inland_user_curve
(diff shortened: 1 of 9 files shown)


antw:
> Question: do we have to make sure that the custom solar FLH input is applied after the regular one? To make sure solar_pv_profile_1_share is always set to 1?

I think so, if only to be certain that the value from the new input is preferred over the existing input (which the user may have set prior to uploading a curve). Currently for the priorities for setting solar FLH are:

* [settings_weather_curve_set](https://github.com/quintel/etsource/blob/master/inputs/misc/settings_weather_curve_set.ad) : priority 2
* [flh_of_solar_pv_solar_radiation](https://github.com/quintel/etsource/blob/master/inputs/full_load_hours/flh_of_solar_pv_solar_radiation.ad) : priority 1

Highest priority inputs are executed _first_. This means that if values are set for both inputs, the value for flh_of_solar_pv_solar_radiation will overwrite the value from the weather year.

_This is the opposite of what I think should happen (the weather curve data should be used in preference over the slider setting, so the priorities should be swapped), but it doesn't matter: the query for flh_of_solar_pv_solar_radiation [is skipped](https://github.com/quintel/etsource/blob/313f99a1aecba63394f0070f66a837dd4bf3d0c2/inputs/full_load_hours/flh_of_solar_pv_solar_radiation.ad#L7-L8) when using a weather year._

I suspect the actual order should be:

* flh_of_solar_pv_solar_radiation : priority 3
* settings_weather_curve_set : priority 2
* flh_of_solar_pv_solar_radiation_user_curve : priority 1

This ensures the new FLH input is run last, so its value will override any set in the previous two inputs. If the new input and existing FLH input have the same priority, the tie-breaking is the natural order of their keys: a key beginning with "energy" would be run before one beginning with "flh". In this case we'd still end up with the desired order (because "_user_curve" sorts after ""), but I don't think we should rely on happy accidents.

antw (approved):
Looks good to me. :+1: Approved, pending your thoughts on the priorities.

michieldenhaan:
I'm a little bit hesitant because of [this](https://github.com/quintel/etsource/commit/6dd90b43285d32d670070466cf83061c5a82eb41) commit which seems to imply that the weather curve priority should be higher than the FLHs.

If that's still valid, maybe it should be something like this?
* settings_weather_curve_set : priority 2
* flh_of_solar_pv_solar_radiation : priority 1
* flh_of_solar_pv_solar_radiation_user_curve : priority 0

antw:
> because of this commit which seems to imply that the weather curve priority should be higher than the FLHs.

I *think* I was mistaken to increase the priority of the settings_weather_curve_set input in that commit. I should have lowered it, or increased the priority of the FLH inputs. In practice, I don't think that change mattered _at all_ because of the `IF`s added to the FLH inputs in that commit.

Today the weather year input [has its own `IF`](https://github.com/quintel/etsource/blob/313f99a1aecba63394f0070f66a837dd4bf3d0c2/inputs/misc/settings_weather_curve_set.ad#L13-L25) which ensures it runs only when a weather year has been chosen. The FLH input queries are skipped when using a weather year, so their relative priorities end up being irrelevant:

* If a weather year is set: the weather input is executed and the FLH inputs are skipped.
* If no weather year is set, or the year is "default": the weather input is skipped and the FLH inputs may execute.

tl;dr: I don't think the priorities of the existing FLH inputs and the weather year input matter, but the *_user_curve inputs should have a lower priority than both of those.

PR #2410, opened 2021-01-29
Fix bug where storage was not imported in import:node task

### LINKED ISSUE #2378: import:node should also update storage attributes
Using the rake task to read updated specifications from ETDataset file to corresponding ETSource file does not function properly. Not all attributes that should be updated are updated, and some attributes that should be deleted are not deleted. Additionally, the order of attributes changes, so it is difficult to verify if all updates are read correctly by the rake task.

To give an example: the dataset containing the specifications for households batteries (households_flexibility_p2p_electricity) has been updated. Below I will give the changes that should be made for a few exemplary attributes, and describe what actually happens when the rake task is used.

output.electricity
- old value in ETSource: 0.85
- new value in ETDataset: **0.90**
- new value in ETSource after rake: 0.90
- This attribute is updated correctly

storage.decay
- old value in ETSource: 4.11e-05
- new value in ETDataset: **2.57e-03**
- new value in ETSource after rake: 4.11e-05
- This attribute is not updated correctly

hours_maint_nl
- old value in ETSource: 0.0
- new value in ETDataset: **None** (removed based on etdataset issue #852)
- new value in ETSource after rake: 0.0
- This attribute is not removed, though it should be

In the end I have updated the ETSource file manually. This is however prone to error, which is why the preferred option would be to have a properly functioning rake task.

noracato:
Closes  #2378

### DIFF
--- a/tasks/etdataset/import_node.rb
+++ b/tasks/etdataset/import_node.rb
@@ -11,6 +11,8 @@ namespace :import do
 
   DESC
   task node: :environment do
+    include ImportHelper
+
     if ENV['NODE'].blank?
       raise "Please provide a node name. For example: bundle exec rake import:node NODE=my_node"
     end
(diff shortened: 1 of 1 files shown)


mabijkerk (approved):
Tested on mv batteries for storage volume and storage decay attributes. Values are updated properly.

PR #2427, opened 2021-03-10
Update LNG demand NI to 15.1% of gas network demand

marliekeverweij:


michieldenhaan (changes requested):
This requires whitelisting the LNG edge on Production as well

michieldenhaan (approved):

PR #2430, opened 2021-03-24
Gelderland update

redekok:
- Update roof surface available for solar PV for Gelderland municipality and province datasets
- Add "Gelderland+" dataset (including all municipalities of the following RES-regions: FoodValley, Achterhoek, Stedendriehoek/Cleantechregio, Arnhem-Nijmegen, Noord Veluwe, Rivierenland)


michieldenhaan (approved):
👍 LGTM! One point of concern is that the solar potential update has quite a bit of fall-out for existing Gelderland (municipality) scenarios. I suppose this has been discussed with the province/OM? If not do we need a migration for this?

PR #2435, opened 2021-03-29
Remove duplicate lignite query from co2 of electricity production

mabijkerk:
I was notified by a user that the emissions shown in the co2 emissions of electricity production chart seemed rather high compared to official data. The chart (see screenshot below) indicated that the total co2 emissions of electricity production in Germany in 2015 were 448 Mton, while official [data](https://www.umweltbundesamt.de/sites/default/files/medien/384/bilder/dateien/6_datentabelle-zur-abb_entw-co2-emi_2020-03-11.pdf) reported 304 Mton.

Screenshot 2021-03-29 at 14 38 20

Looking into the underlying queries it turned out that the emissions of coal counted those of both coal and lignite power plants, even though the emissions of lignite were shown separately in the chart. Lignite has therefore been removed from the query `electricity_producing_coal_converters` and affected queries have been adjusted. The resulting emissions for Germany in the updated chart appear to be much more in line with the official data (see screenshot below).

Screenshot 2021-03-29 at 14 39 43

Additionally, the queries `electricity_producing_waste_converters`, `electricity_producing_lignite_converters` and `electricity_producing_torrefied_biomass_pellets_converters` had a query structure which resulted in nodes being selected that could **potentially** use waste/lignite/torrefied wood pellets to produce electricity, instead of nodes that **actually produce** electricity in a certain scenario - which is the case for the other electricity producing converter queries. These queries have been updated to match the other electricity producing converter queries and affected queries have been adjusted.

### DIFF
--- a/gqueries/general/converter_groups/electricity_producing_coal_converters.gql
+++ b/gqueries/general/converter_groups/electricity_producing_coal_converters.gql
@@ -1,7 +1,7 @@
 - query =
     FILTER(
       G(electricity_production),
-      "coal_input_conversion > 0 || lignite_input_conversion > 0"
+      "coal_input_conversion > 0"
     )
 - unit = nodes
 - deprecated_key = coal_plants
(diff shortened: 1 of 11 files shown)


michieldenhaan (approved):
🙇‍♂️

PR #2436, opened 2021-03-31
Queries for imported and renewable heat in primary demand

mabijkerk:
In etmodel [#3668](https://github.com/quintel/etmodel/issues/3668) it is addressed that solar thermal and imported heat were missing from the primary energy demand mekko. This pull request therefore adds queries for:

- Renewable heat: this includes solar thermal and steam hot water. Steam hot water is added because the primary demand query uses the output energy carrier and the node `energy_heat_solar_thermal`, which represents the use of solar thermal energy in district heating, has steam hot water as output energy carrier. `energy_heat_solar_thermal` is the only primary demand node with steam hot water as output energy carrier. Referring to renewable heat enables us to potentially add geothermal heat to these queries in a later stage for etmodel [#3641](https://github.com/quintel/etmodel/issues/3641).
- Imported heat

Minor fixes for the queries `primary_demand_of_hydrogen_from_export` and `imported_hydrogen_export_and_curtailment_in_mekko_of_primary_demand` are included in this pull request.

### DIFF
--- a/gqueries/general/primary_demand/primary_demand_of_imported_heat_from_agriculture.gql
+++ b/gqueries/general/primary_demand/primary_demand_of_imported_heat_from_agriculture.gql
@@ -1,0 +1,7 @@
+- unit = PJ
+- query =
+    DIVIDE(
+      V(INTERSECTION(SECTOR(agriculture),
+      G(final_demand_group)),primary_demand_of_imported_heat).sum,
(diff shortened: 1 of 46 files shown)


marliekeverweij (approved):
Very nice!

marliekeverweij (approved):

PR #2438, opened 2021-04-01
Import new SMR specs from ETDataset

michieldenhaan:


marliekeverweij:
We have a green light from our clients from Northern-Ireland to merge this :)

mabijkerk:
@michieldenhaan do we still need to check scenario migration for this update?

michieldenhaan:
Do you have an opinion on scenario migrations @ChaelKruip @marliekeverweij? We updated the SMR research, resulting in a higher hydrogen output efficiency (4-5 percentage points) and a lower carbon capture rate (13 percentage points). We could do a migration by reducing installed SMR capacity (to offset the first effect) and increasing the CCS slider (to offset the second effect). 

I'm inclined not to do a migration given that the new data is better and SMR typically plays a minor role in scenarios. Let me know if you disagree!

marliekeverweij:
> I'm inclined not to do a migration given that the new data is better and SMR typically plays a minor role in scenarios. Let me know if you disagree!

I agree!

mabijkerk (approved):
LGTM

PR #2448, opened 2021-04-26
Update PV26 Utrecht to 2018 data

mabijkerk:
Based on update of Utrecht in the ETM Dataset Manager https://github.com/quintel/etlocal/pull/302.


datasets/PV26_utrecht/PV26_utrecht.derived.ad
```diff
@@ -126,37 +104,59 @@
...
-- other_emission_transport = 0.113467
-- other_emission_industry_energy = 0.179275
-- number_of_detached_houses = 40056.719234
-- number_of_apartments = 198006.1825
-- number_of_semi_detached_houses = 41169.938768
-- number_of_corner_houses = 72187.710028
-- number_of_terraced_houses = 196693.44947
-- base_dataset = nl
+- total_land_area = 1487.52
+- number_of_buildings = 37084.0
```
redekok, line 142: Wow, it seemed as if the number of buildings has dropped by half. Maybe we should double check that number based on Klimaatmonitor/CBS data for the province?

mabijkerk:
According to the Dataset Manager it has actually increased (though the previous change in the Dataset Manager does not appear to have been committed to ETSource). I will look into the underlying data.

Screenshot 2021-04-26 at 11 25 46

mabijkerk:
According to the previous commit, buildings with "overig" and "industrie" are not taken into account. Looking at the 2018 data from [CBS](https://opendata.cbs.nl/statline/#/CBS/nl/dataset/81955NED/table?dl=51909) (see screenshot below), subtracting the total number of non-residence building with the "overig" and "industrie" buildings gives 32719 buildings.

Screenshot 2021-04-26 at 12 12 44

redekok:
Thanks for checking @mabijkerk! And I agree, let's go with the 37k value.

redekok (approved):

PR #2449, opened 2021-04-29
Update all Utrecht municipality datasets to 2018

redekok:
And keep the older 2016 dataset for Amersfoort in order to maintain the (published) original scenario results


datasets/GM0307_amersfoort/GM0307_amersfoort.derived.ad
```diff
@@ -126,37 +104,59 @@
...
-- number_of_buildings = 9623.0
-- number_of_cars = 98852.0
-- number_of_residences = 65289.0
-- number_of_inhabitants = 153602.0
-- offshore_suitable_for_wind = 448.608030215936
+- number_of_buildings = 4932.0
+- number_of_cars = 102990.0
+- number_of_residences = 66530.0
+- number_of_inhabitants = 155226.0
+- offshore_suitable_for_wind = 12500.0
```
mabijkerk, line 146: The offshore area suitable for wind seems to have increased strongly. Is this as expected?

mabijkerk: It is now the same value for all municipalities: 12500 square kms.

redekok: > It is now the same value for all municipalities: 12500 square kms.

Interesting.. This seems to be the case for all 2018 datasets in the dataset manager. @michieldenhaan would you have any idea what's going on here?

michieldenhaan: If I remember correctly we first scaled the potential based on number of residences but after receiving complaints that this was too strict in some scenarios we decided to allocate the full NL potential to each municipality (to ensure the slider max is always sufficient)

redekok: That explains, thanks!

mabijkerk (approved):

PR #2451, opened 2021-05-05
Add detailed greenhouse gas emissions

michieldenhaan:
Goes with:
[ETModel](https://github.com/quintel/etmodel/pull/3718)
[ETEngine](https://github.com/quintel/etengine/pull/1167)
[Atlas](https://github.com/quintel/atlas/pull/146)


antw:
Looks good to me. It can't be merged due to a conflict: some datasets have been renamed or removed. If I understand correctly, the "no_highways" and "no_steel" versions of datasets were removed entirely? There another later commit also with conflicts – due to both `master` and this branch having changes to the same dataset attributes – that I don't know enough about to be comfortable resolving.

@michieldenhaan Would you mind taking a look when you have some time available?

michieldenhaan:
Thanks @antw! We're good to go now.

PR #2455, opened 2021-05-11
Dataset Breda BiZ Steenakker

DorinevanderVlies:


DorinevanderVlies:
@noracato I changed the group to region. I cannot see this specified on ETLocal should we make changes there as well?

noracato:
Good catch!

>  I cannot see this specified on ETLocal should we make changes there as well?

No, it's a remnant of exporting the dataset when it still had its old `geo_id`. It defaults to province when the `geo_id` does not start with any recognisable code. Since the new one starts with `RG` on future exports to `etsource` it will be interpreted as a region.

PR #2464, opened 2021-05-25
Update NI emission attributes based on DfE data

michieldenhaan:


marliekeverweij:
Looks good to me!

The only thing that I doubt about is the change of the `scaling.value`. Is that intended?

Screenshot 2021-05-25 at 13 00 19

michieldenhaan:
The scaling value automatically equals the number of residents. For some reason that wasn't the case for NI, I suppose that wasn't intentional? I think the impact is small (we override many of the scaled values, apart maybe from some area attributes?).

marliekeverweij (approved):

PR #2473, opened 2021-06-18
Replace generic coal transformation loss with updated energy flows for cokes

mabijkerk:
This pull request removes the need for the `industry_transformation_generic_coal` loss. Instead, the energy flows for cokes are remodelled in such a way that they better represent the energy balance.

**Old solution:**
The steel sector demand for cokes and the total demand for cokes are not necessarily the same on the energy balance. The difference can be exported/imported or used for final consumption. In the model however, only cokes use by the steel sector is modelled. Because cokes are produced with coal, the difference in modelled cokes demand and actual cokes demand therefore also results in a difference in primary coal demand. This difference is corrected in the `industry_transformation_generic_coal` node.

**New solution:**
The export and import of cokes are made explicit and the `energy_distribution_cokes` node is now also connected to the `industry_final_demand_coal` and `industry_final_demand_coal_non_energetic` nodes. The result is that the cokes production in the model now aligns with the energy balance, which means that there is no difference in modelled primary coal demand from cokes production and the `industry_transformation_generic_coal` loss is set to 0.

@michieldenhaan could you verify this solution and check whether all the necessary steps were taken (e.g. export the correct local datasets). I believe you also wanted to check something for the molecule graph?


graphs/energy/nodes/industry/industry_transformation_generic_coal.ad
```diff
@@ -5,4 +5,4 @@
 - groups = []
 - free_co2_factor = 0.0
 
-~ output.coal = EFFICIENCY(industry_transformation_generic_coal, output, coal)
+~ output.coal = 1.0
```
michieldenhaan, line 8: Better to change to `- output.coal = 1.0` and move to line 3

datasets/RES17_noord_holland_zuid/RES17_noord_holland_zuid.derived.ad
```diff
@@ -140,38 +139,23 @@
...
-- non_energetic_emissions_co2_chemical_industry = 0.0
-- non_energetic_emissions_co2_other_industry = 0.0
-- non_energetic_emissions_co2_waste_management = 0.0
-- non_energetic_emissions_other_ghg_chemical_industry = 0.0
-- non_energetic_emissions_other_ghg_other_industry = 0.561904
-- non_energetic_emissions_other_ghg_waste_management = 0.0
-- energetic_emissions_other_ghg_industry = 0.0
-- energetic_emissions_other_ghg_energy = 0.0
-- indirect_emissions_co2 = 0.0
+- other_emission_agriculture = 0.416009
```
michieldenhaan, line 149: These attributes are outdated. Exported using an old version of the data manager / data manager database?

michieldenhaan (changes requested):
Excellent work @mabijkerk! I've added two small comments but apart from that we're good to go!

graphs/energy/nodes/industry/industry_transformation_generic_coal.ad
mabijkerk: Changed it accordingly.

datasets/RES17_noord_holland_zuid/RES17_noord_holland_zuid.derived.ad
mabijkerk: Exported with updated dataset manager and database.

michieldenhaan (approved):

PR #2481, opened 2021-07-05
Query for percentage energetic emissions CO2 sheet

Charlottevm:


gqueries/modules/co2_sheet/indicators/co2_sheet_energetic_emissions.gql
```diff
@@ -0,0 +1,47 @@
...
+        Q(emissions_non_energetic_agriculture_soil_cultivation_co2),
+        Q(industry_chemical_non_energetic_co2_emissions),
+        Q(emissions_non_energetic_industry_chemical_other_ghg),
+        Q(emissions_non_energetic_other_industry_other_ghg),
+        Q(emissions_non_energetic_industry_waste_management_other_ghg),
+        Q(emissions_non_energetic_agriculture_fermentation_other_ghg),
+        Q(emissions_non_energetic_agriculture_manure_other_ghg),
+        Q(emissions_non_energetic_agriculture_soil_cultivation_other_ghg),
+        Q(emissions_non_energetic_agriculture_other_other_ghg),
+        Q(emissions_energetic_agriculture_other_ghg),
```
michieldenhaan, line 39: The energetic other GHG queries are included twice in the denominator

gqueries/modules/co2_sheet/indicators/co2_sheet_energetic_emissions.gql
```diff
@@ -0,0 +1,47 @@
...
+        Q(emissions_energetic_buildings_other_ghg),
+        Q(emissions_energetic_households_other_ghg),
+        Q(emissions_energetic_energy_other_ghg),
+        Q(emissions_energetic_industry_other_ghg),
+        Q(emissions_energetic_transport_other_ghg)
+      )
```
michieldenhaan, line 45: Delayed and indirect emissions are missing

michieldenhaan (changes requested):
Based on NL2015 data, I would expect the % energetic to be 85.136%. The current query returns 83.1%.

For the denominator: I would use the same queries used in the CO2 sheet. Would make this query much simpler and less error prone.

So SUM(
Q(co2_sheet_agriculture_total_emissions),
Q(co2_sheet_industry_energy_total_emissions),
Q(co2_sheet_transport_total_emissions),
Q(co2_sheet_buildings_households_total_emissions),
Q(co2_sheet_indirect_delayed_emissions))

michieldenhaan:
Another minor thing is to make the query name more descriptive. E.g. `co2_sheet_energetic_emissions_share` instead of `co2_sheet_energetic_emissions`

Charlottevm:
Thanks for your comments! How did you produce the 85.136% of energetic emissions in the NL dataset 2015? If I calculate the percentage based on the ETM emissions overview graph/table I get 83.1%.

michieldenhaan:
Sorry my bad, I summed all the non-energetic emissions in the area attributes and divided by the total. But this neglects the delayed emissions as these are calculated by the ETM, not via area attributes. So your number is correct! 🎈 

The current query returns the right number for nl2015, even though energetic GHG is double counted and delayed/indirect emissions are missing. Turns out those are almost equal to each other for nl2015 🤯

Charlottevm:
Wow sick! I guess I made the perfect mistake :sweat_smile: But due to your mistake we have prevented a serious error, so I guess the universe giving back :earth_africa: Thanks for your sharp review!

PR #2499, opened 2021-07-30
Update Heerhugowaard dataset to 2018

Charlottevm:


redekok (changes requested):
See https://github.com/quintel/etsource/pull/2497#pullrequestreview-719123112

PR #2508, opened 2021-08-05
Must-run producer changes for security of supply chart etmodel issue #3778

mabijkerk:
Changes made for https://github.com/quintel/etmodel/issues/3778:

- This adds a query for must-run production in the security of supply chart.


redekok:
LGTM! Minor thing, the description of this PR is still outdated since it says that the waste incinerator is changed into a must-run producer. Let's change that before merging this, agreed? :-)

PR #2513, opened 2021-08-17
Add air heatpumps buildings to investment table

Charlottevm:


redekok (commented):
I think the gqueries for heatpumps used for cooling aren't necessary though; these will cause a double counting of costs in the investment table.

redekok:
@Charlottevm I don't see any changes to the node files. Just to make sure, were the efficiencies for the ASHP changed after all?

redekok (approved):

PR #2517, opened 2021-08-19
Add sustainability_share to primary demand nodes

antw:
Some primary demand nodes which are sustainable had no `sustainability_share` assigned, and relied on ETEngine treating nodes with "infinite" carriers differently when calculating the share.

This is the ETSource part of the implementation for [the second option discussed etmodel#3795](https://github.com/quintel/etmodel/issues/3795#issuecomment-897645801) whereby we'll raise an error if we cannot find a `sustainability_share` – or carriers with a `sustainable` value – when reaching "primary_energy_demand" nodes.

### DIFF
--- a/graphs/energy/nodes/energy/energy_heat_solar_thermal.converter.ad
+++ b/graphs/energy/nodes/energy/energy_heat_solar_thermal.converter.ad
@@ -26,6 +26,7 @@
 - construction_time = 2.0
 - technical_lifetime = 25.0
 - wacc = 0.04
+- sustainability_share = 1.0
 
 ~ demand = 0.0
 
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ gem 'rake'
 
 group :development, :test do
   gem 'roo'
-  gem 'atlas',    ref: '0546b99', github: 'quintel/atlas'
+  gem 'atlas',    ref: 'd80297d', github: 'quintel/atlas'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GIT
   remote: https://github.com/quintel/atlas.git
-  revision: 0546b99596e7b5234113c49eba307f389e0b311d
-  ref: 0546b99
+  revision: d80297d705bca0e81733d1a04c74acf090372d30
+  ref: d80297d
   specs:
     atlas (1.0.0)
       activemodel (>= 4.1.14.1)
@@ -29,9 +29,9 @@ GIT
 GEM
   remote: https://rubygems.org/
   specs:
-    activemodel (6.1.4)
-      activesupport (= 6.1.4)
-    activesupport (6.1.4)
+    activemodel (6.1.4.1)
+      activesupport (= 6.1.4.1)
+    activesupport (6.1.4.1)
       concurrent-ruby (~> 1.0, >= 1.0.2)
       i18n (>= 1.6, < 2)
       minitest (>= 5.1)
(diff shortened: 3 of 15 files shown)


graphs/energy/nodes/energy/energy_heat_solar_thermal.converter.ad
```diff
@@ -26,6 +26,7 @@
 - construction_time = 2.0
 - technical_lifetime = 25.0
 - wacc = 0.04
+- sustainability_share = 1.0
```
redekok, line 29: @antw, thanks for setting up this PR! Just a quick question: I noticed that a `sustainability_share` was added to the `energy_heat_solar_thermal` node but not to the `households_final_demand_solar_thermal` and `buildings_final_demand_solar_thermal` nodes. The latter are also in the `primary_energy_demand` group. Is there a reason why these didn't get a `sustainability_share`?

antw:
> The latter are also in the primary_energy_demand group. Is there a reason why these didn't get a sustainability_share?

The heuristic I used for adding a `sustainability_share` was to find primary_energy_demand nodes where the following two conditions were true:

* The node has no `sustainability_share` already.
* The node's output carriers don't define a `sustainable` value.

When both are true, ETEngine cannot calculate a `sustainability_share` for the node.

In the case of "households_final_demand_solar_thermal" and "buildings_final_demand_solar_thermal", both have solar_thermal as their output carrier. This carrier [defines a value for `sustainable`](https://github.com/quintel/etsource/blob/d77f954a6abf6323a460c512b85c9ca221da2993/carriers/solar_thermal.ad#L1) from which ETEngine can calculate a `sustainability_share` for the nodes.

"energy_heat_solar_thermal", on the other hand, has steam_hot_water as its main output carrier. This carrier does not define a `sustainable` value.

If you don't like the inconsistency, an alternative might be to remove all three nodes from the primary_energy_demand group, and replace them with "[energy_production_solar_thermal](https://engine.energytransitionmodel.com/inspect/858990/graphs/energy/nodes/households_final_demand_solar_thermal)". There are no conversion losses between this node and its outputs so the results should be identical, with the bonus that ETEngine can calculate the `sustainbility_share` from its solar_thermal output.

graphs/energy/nodes/energy/energy_heat_solar_thermal.converter.ad
antw: > The latter are also in the primary_energy_demand group. Is there a reason why these didn't get a sustainability_share?

The heuristic I used for adding a `sustainability_share` was to find primary_energy_demand nodes where the following two conditions were true:

* The node has no `sustainability_share` already.
* The node's output carriers don't define a `sustainable` value.

When both are true, ETEngine cannot calculate a `sustainability_share` for the node.

In the case of "households_final_demand_solar_thermal" and "buildings_final_demand_solar_thermal", both have solar_thermal as their output carrier. This carrier [defines a value for `sustainable`](https://github.com/quintel/etsource/blob/d77f954a6abf6323a460c512b85c9ca221da2993/carriers/solar_thermal.ad#L1) from which ETEngine can calculate a `sustainability_share` for the nodes.

"energy_heat_solar_thermal", on the other hand, has steam_hot_water as its main output carrier. This carrier does not define a `sustainable` value.

If you don't like the inconsistency, an alternative might be to remove all three nodes from the primary_energy_demand group, and replace them with "[energy_production_solar_thermal](https://engine.energytransitionmodel.com/inspect/858990/graphs/energy/nodes/households_final_demand_solar_thermal)". There are no conversion losses between this node and its outputs so the results should be identical, with the bonus that ETEngine can calculate the `sustainbility_share` from its solar_thermal output.

michieldenhaan: I think that won't work because of the losses on this node:
image

Ideally we should revise the households/buildings implementation of solar thermal at some point so that it is more in line with the district heating node. But I guess that's something for the long list. I'm happy with your current implementation of `sustainability_share` and the two conditions you describe ✌️

michieldenhaan (approved):

redekok (approved):

PR #2526, opened 2021-08-27
Add efficiencies for input carriers in oil mix for agriculture

noracato:
Reverts the agriculture final demand crude oil node only accepting crude oil. All oil mix carriers (diesel, biodiesel, kerosene, bio_kerosene, lpg, bio_oil and crude_oil) are now input carriers for this node again. The issue of the missing agriculture sector of Germany has been circumvented by adding input efficiencies for all countries, and setting the German inputs to be 100% crude oil from its efficiency file.


redekok (approved):
LGTM!

PR #2528, opened 2021-09-01
chemical other useful demand extra input statements

karel97:
Add input statements for electricity, useable heat, and non energetic shares in chemical other useful demand.

At the moment the ratio between the electricity, useable heat, and non energetic shares of the total industry_useful_demand_for_chemical_other is fixed when one changes the total chemical other useful demand. For the coupling with the CTM we would like to vary this ratio and therefore communicate directly with these nodes through the API. For this reason we added three new input statements, which are invisible to the user through ETModel and do not affect the model when left at their default value of 100%. 

@noracato @redekok

### DIFF
--- a/inputs/demand/industry/industry_useful_demand_for_chemical_other_electricity.ad
+++ b/inputs/demand/industry/industry_useful_demand_for_chemical_other_electricity.ad
@@ -1,0 +1,11 @@
+# adds an input statement for updating the industry_useful_demand_for_chemical_other_electricity
+
+- query = UPDATE_WITH_FACTOR(V(industry_useful_demand_for_chemical_other_electricity), preset_demand, USER_INPUT())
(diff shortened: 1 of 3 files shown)


noracato (approved):
Looks good @karel97! Feel free to merge, or to look at my nitpicky suggestion first 😉.  I assumed you chose the max of 10,000 and step size of 0.001 because you need such precision for your coupling-project.

karel97:
Thanks for the review @noracato ! We indeed purposefully changed the max values and step size for the coupling with the CTM. This shouldn't be necessary in the long run but for now it is a workable solution. I've removed the comments from the input statements, I agree they are not needed. It seems like I don't have permission to merge the pull request? Please let me know if I'm mistaken, but otherwise feel free to merge it yourself.

PR #2531, opened 2021-09-02
Add CCS demands for waste incineration to the energy use for CCUS chart

MartLubben:
Closes https://github.com/quintel/etmodel/issues/3818


graphs/energy/nodes/energy/energy_power_supercritical_ccs_waste_mix.ad
```diff
@@ -37,4 +37,4 @@
 
 ~ demand = 0
 ~ full_load_hours = CENTRAL_PRODUCTION(energy_power_supercritical_waste_mix, full_load_hours)
-~ output.electricity = CENTRAL_PRODUCTION(energy_power_supercritical_waste_mix, electricity_output_conversion)
+~ output.electricity = 0.20
```
redekok, line 40: Just out of curiosity: why are we setting this value hard-coded instead of querying it from the `central_producers.csv`?

redekok (approved):

PR #2542, opened 2021-09-21
Update chart 'Total Cost of Electricity Production'

lottevanvlimmeren:
One of the queries needed for the calculation of the costs of biomass CHP use included a node that was missing in the energy graph. I have changed the node in the query to an existing node.


lottevanvlimmeren:
Closes #3842

mabijkerk (commented):
Looks good! Can be merged after removing obsolete queries that still include the agriculture chp node.

PR #2545, opened 2021-09-23
Remove outdated inputs

### LINKED ISSUE #2524: default value for number_of_industry_chp_combined_cycle_gas_power_fuelmix_present greater than maximum value
Not sure where to submit this issue, but the default value for number_of_industry_chp_combined_cycle_gas_power_fuelmix_present seems to be greater than the specified maximum value.

michieldenhaan:
Removes outdated inputs for adjust_scaling, targets and employment which are no longer in use. Avoids confusing for API users requesting the list of available inputs.

Closes #2524

@antw Does this require a migration? IIRC there is no harm in scenarios containing outdated user_values but may be nicer to remove?


antw (approved):
Thanks! :+1:

antw:
> @antw Does this require a migration? IIRC there is no harm in scenarios containing outdated user_values but may be nicer to remove?

Not _really_. User values without a matching input are ignored. The only value to removing them from a scenario is if you think we might reuse an input name again in the future.

I can put together a migration if you'd like one.

michieldenhaan:
Thanks! I don't think that will be the case so happy to leave it

PR #2554, opened 2021-10-04
Add heat demand curves for the DE and BE NUTS1 regions

noracato:
Symlinked the whole curves folder to the parent set, with the only exception being the regions own heat demand curves. These heat demand curves were imported from `etdataset`.

@antw, @michieldenhaan , from Atlas I derived that this is how it should be done: symlinking the whole `curves` directory, when you want to change/add a single curve. But I haven't seen it done for other regional datasets. So a small check with either of you, if what I've done won't break anything?


antw:
> symlinking the whole curves directory

This shouldn't be necessary anymore. Create a `./curves` directory for the dataset and place into it any custom CSVs. There should be there need to symlink curves from the parent dataset; Atlas will do the right thing.

Charlottevm (approved):

PR #2556, opened 2021-10-05
Fix the DE-regions heat demand curves to be of length 8760

noracato:
Fixed the buildings heat demand, and thrown in a test for this not to happen again.

Let's do this on pro as well.


antw (approved):
Great, thanks! Also, thank you for adding the test. :tada:

PR #2592, opened 2022-01-04
Innovations flex pvt

lottevanvlimmeren:
Addition of heat pump+pvt and the addition of the roof area chart for households


mabijkerk (approved):
I included `pvt` in the household final demand space heating name. Other than that it looks good.

Issue #2585, opened 2021-12-15
EU countries: sliders for production plants should not be disabled when present capacity is 0


DorinevanderVlies:
For some power plants the slider is disables when the present capacity is 0.

![image](https://user-images.githubusercontent.com/19907658/146208480-deb76992-f766-423e-af84-658d413e8eca.png)

PR #2597, opened 2022-01-10
Update EU datasets from Dataset Manager

noracato:
Re-exported all EU datasets from the Dataset Manager.


marliekeverweij:
https://github.com/quintel/etsource/issues/2585

PR #2613, opened 2022-02-02
Export Northern-Ireland dataset again with new basedataset EU

marliekeverweij:
A carriers.csv containing the emission factors specific for Northern-Ireland is added. This file was not included before, because before the base_dataset was `uk` and the Northern-Ireland dataset used the emission factors of that dataset. Now that we have changed to `eu` as dataset, the Northern-Ireland datasets needs it's own carriers.csv file.


datasets/UKNI01_northern_ireland/UKNI01_northern_ireland.derived.ad
```diff
@@ -18,26 +18,25 @@
 - has_mountains = true
 - has_old_technologies = true
 - has_other = true
-- has_solar_csp = false
-- has_offshore_solar = false
```
mabijkerk, line 22: I don't think that the line `has_offshore_solar = false` should be deleted.

datasets/UKNI01_northern_ireland/UKNI01_northern_ireland.derived.ad
```diff
@@ -18,26 +18,25 @@
...
+- has_solar_csp = true
```
mabijkerk, line 21: Why is `has_solar_csp` true for this dataset? NL doesn't have CSP either and as far as I know the weather in the UK/NI is not that much better than in NL.

datasets/UKNI01_northern_ireland/UKNI01_northern_ireland.derived.ad
```diff
@@ -18,26 +18,25 @@
...
 - co2_emission_1990_aviation_bunkers = 15.37818186
 - co2_emission_1990_marine_bunkers = 8.69679035
-- co2_percentage_free = 0.85
-- co2_price = 0.005
-- offshore_ccs_potential_mt_per_year = 42.5
-- economic_multiplier = 0.0
-- employment_fraction_production = 0.0
-- employment_local_fraction = 0.0
+- co2_percentage_free = 0.4
+- co2_price = 0.02483
```
mabijkerk, line 35: This CO2 price is linked to the UK 2019 dataset, while the NI dataset is for 2018. I suppose this applies to other values as well, is this an issue or is this as good as it gets, because we don't have a UK 2018 dataset?

datasets/UKNI01_northern_ireland/graph_values.yml
```diff
@@ -170,36 +174,50 @@ energy_heat_burner_wood_pellets:
...
     :steam_hot_water: 0.9
+  full_load_hours: 2190.0
 energy_chp_supercritical_waste_mix:
   demand: 885.1851851851851
   full_load_hours: 4155.0
   output:
     :steam_hot_water: 0.15
     :electricity: 0.27
 energy_chp_ultra_supercritical_lignite:
-  demand: 0.0
```
mabijkerk, line 180: I don't think this line should be deleted: this would lead to an earlier issue where non-existent lignite power plants were added to the model.

mabijkerk:
I have reviewed the commits. Some smaller comments which I could address myself if you like @marliekeverweij? A few require some discussion.

Why is the graph_values.yml exported again as well? Do changes from the scaling affect these values?

datasets/UKNI01_northern_ireland/UKNI01_northern_ireland.derived.ad
marliekeverweij: Yes, true. But by default this attribute is set to false by @noracato, so luckily this is no issue

datasets/UKNI01_northern_ireland/UKNI01_northern_ireland.derived.ad
marliekeverweij: Sharp. The UKNI is based on the `eu` dataset and there solar csp is true. I think not a big issue if this technology is available, it is actually available for all EU countries now

datasets/UKNI01_northern_ireland/UKNI01_northern_ireland.derived.ad
marliekeverweij: This is the CO2 price and percentage of the `eu` dataset and this is the same as the nl2019 dataset. So indeed the year doesn't correspond exactly, but we indeed do not have a 2018 UK dataset

datasets/UKNI01_northern_ireland/graph_values.yml
marliekeverweij: Good points (this and the next!). @antw do you know why this demand is deleted? We have added `demand` to the `graph_methods` in the nodes ànd added a sparse graph query setting this demand

antw: The demand values aren't deleted: they're just moved a few lines lower.

Screenshot 2022-02-02 at 14 30 54

marliekeverweij: Woooooops! Sorry

mabijkerk: That one is on me, sorry!

PR #2614, opened 2022-02-02
Export Northern-Ireland dataset again with new basedataset EU

marliekeverweij:


mabijkerk:
Changes in Electricity production are significant. These can however be explained by the CO2 price and free allocation, which now have the proper values. Still may be worth investigating what the merit order actually was in 2018: was coal before gas and gas before import? Or was that ordering also based on an incorrect CO2 price and free allocation?

mabijkerk (approved):

PR #2618, opened 2022-02-04
WIP - Updated Turnhout with export  script from dataset manager

MartLubben:
Have to check first, don't merge.


MartLubben:
Checked LIVE with BETA. 
Two desired changes:

- More electricity produced by solar on buildings, which leads to less import and more CO2-reduction in the start year.
- A bug removed in the scaling of the costs of the gas network. Leads to lower costs of the gas network in the start year.

marliekeverweij (approved):

PR #2630, opened 2022-02-10
External coupling of industry sectors

marliekeverweij:
This PR includes the external coupling for refineries, fertilizers & chemical other sector.

For the fertilizer industry we have added two coupling nodes to the graph (non-energetic & energetic), for the other two sectors we have added external coupling inputs only.


mabijkerk:
The WACC of metal industry now has an input statement that uses a factor e.g., 0.07, while for the input statement itself says that the unit is a %. For fertilizers I have made the input statement accept a percentage for the WACC e.g., 7.0. For alignment I would suggest changing the WACC input statement of the metal industry to accept a percentage instead of a factor.

mabijkerk (approved):

Issue #2568, opened 2021-11-01
Final demand biodiesel for transport correction


lottevanvlimmeren:
The demand for node `transport_final_demand_biodiesel` is determined directly from the energy balance as following:

![image](https://user-images.githubusercontent.com/89903998/139683272-3499ba8b-7dda-4173-820f-35e269867bf9.png)

For the "transport" flow, the products "biodiesels", "other liquefied biofuels", and "non-specified primary biofuels and waste" are being used. For the "non-specified transport" and "pipeline transport" flows, the product "biodiesels" is determined from the energy balance three times. I assume this is a copy-paste mistake, and that the two of the three "biodiesels" could be changed into "other liquefied biofuels" and "non-specified primary biofuels". @mabijkerk do you agree?

mabijkerk:
I agree that it is probably a copy-paste mistake. The `8_transport_analysis` clearly shows that we assume the ETM carrier 'biodiesel' to consist of 'biodiesels', 'other liquid biofuels' and 'non-specified primary biofuels and waste'.

Could you check other transport nodes for similar mistakes and update them to the correct products? I checked `transport_final_demand_gasoline.final_demand.ad` and it contains a similar mistake for 'motor gasoline'.

lottevanvlimmeren:
The demand for node `transport_final_demand_biodiesel` seems to have been rewritten already. There were two other nodes that had the same mistake: `transport_final_demand_coal` and `transport_final_demand_gasoline`.

PR #2632, opened 2022-02-15
Minor fix of energy balance carriers in transport final demand nodes

lottevanvlimmeren:
Some energy carriers were read from the energy balance twice within a node, while other energy carriers were missing. This is fixed now for the final demand of gasoline and coal in the transport sector.


lottevanvlimmeren:
Solves #2568

mabijkerk (approved):
Looks good!

PR #2656, opened 2022-03-22
Add hydrogen, solar thermal and solar pvt heating and hot water technologies to final demand chart

mabijkerk:
This PR adds missing heating technologies to the households final demand per application chart:

Screenshot 2022-03-22 at 10 09 09

@Charlottevm could you check whether the chart is consistent now?


Charlottevm (approved):
Looks good to me!

PR #2663, opened 2022-03-23
Demand side response - load shifting in industry with dataset update

lottevanvlimmeren:
Addition of demand side response in industry: load shifting of electricity is now available for the industrial sectors metal, chemical, and other. The datasets have been updated from ETLocal


mabijkerk (approved):
🚀

Issue #2649, opened 2022-03-11
Add external_inputs for CCUS sliders and make it possible to set % of coal gas


marliekeverweij:
1. Please add external_inputs for these CCUS sliders:
https://pro.energytransitionmodel.com/scenario/emissions/ccus/utilisation-and-storage-of-co2
![Screenshot 2022-03-11 at 15 59 25](https://user-images.githubusercontent.com/32833996/157893191-5a3ca175-4f07-449e-9164-d94cebc30116.png)
https://pro.energytransitionmodel.com/scenario/emissions/ccus/coal-gases-from-blast-furnaces-steel
![Screenshot 2022-03-11 at 16 04 16](https://user-images.githubusercontent.com/32833996/157893179-ea0561ed-523d-4d09-ab55-71476b34acae.png)

2. Additionally, it is desired that it's possible to adjust the percentage of coal gas in this node:
https://engine.energytransitionmodel.com/inspect/902955/graphs/energy/nodes/energy_distribution_coal_gas
![Screenshot 2022-03-11 at 16 07 32](https://user-images.githubusercontent.com/32833996/157893756-7792265e-a994-4693-bca4-2099672771af.png)

mabijkerk:
For the share of coal gas in the energy distribution coal gas node it is important that the total output sums to 100%. The input statements should therefore also update the reversed output shares, not just the final demand for coal gas share.

PR #2669, opened 2022-03-28
Add input statements and queries for external coupling

mabijkerk:
This PR adds queries for the biogenic emissions of hydrogen and electricity. Additionally, this PR adds an input statement to change the share of final coal gas demand in the energy distribution coal gas node through external coupling (see #2649).


marliekeverweij (commented):
Looks good! One comment:
I think the group in this query has to be `hydrogen_production`
gqueries/modules/external_coupling/fossil_co2_emissions_of_produced_hydrogen.gql

mabijkerk:
Oops, I updated it!

Issue #2662, opened 2022-03-23
Gas network costs of ETLocal datasets are unrealistic small 


marliekeverweij:
In [this pull request ](https://github.com/quintel/etsource/pull/2661)we have updated all ETSource datasets with the current data on ETLocal 🎉

During this update the gas network costs of all datasets decreased with a factor 1.000.000. This has to be updated before next deploy!

https://github.com/quintel/etsource/pull/2661/files#diff-45886ffb82aaecb57fb53239c521dd307f1e4f4bd8bbec0012cf325e0a684ee7L137-R137

marliekeverweij:
I haven't had the time to look into this issue earlier. The reason that this issue pops up is that we haven't set a value in ETLocal and when there is no value, ETLocal automatically takes the value in the area-file for the ETLocal key `annual_infrastructure_cost_gas` and uses the SGQ `area+annual_infrastructure_cost_gas.ad` which divides it by one million. That's why this area-attribute decreases with a factor 1.000.000 every time we export a dataset.

How can we solve this? We should set a value for each dataset on ETLocal.

@antw / @noracato do you agree with this? Would you have time before the deploy to write a dataset migration for all ETLocal dataset, which scales the Dutch gas network costs with the total final demand gas? The Dutch value is 1,452,000,000 euro and can be found here https://github.com/quintel/etsource/blob/master/datasets/nl2019/nl2019.full.ad#L145

I could provide a list of the total final demand gas per dataset on Monday

antw:
> Would you have time before the deploy to write a dataset migration for all ETLocal dataset

Are you sure that it's right to set a value for each dataset? Would it not be preferred to scale the value from the NL dataset each time, so that if NL were ever updated the regional datasets would automatically reflect this change?

Why does the SQL divide by one million?

marliekeverweij:
Thank for your response @antw, interesting questions

IMO we want to move to the direction that all data is set through the Dataset Manager. Then the Dataset Manager will give a complete overview of the data (very transparent) and no non-transparant scaling (and undesired?) takes place. The datasets will not be "derived" anymore, but stand-alone. We are not there yet and I would like to investigate what we need to do to get there. 

In the past entso-project we have added the key for gas network costs to the Dataset Manager. This way users can see the assumption that has been made for their datasets. For gas network costs they can see that this is not a well researched value, but an estimation based on the Dutch network costs.

[...]

PR #2675, opened 2022-04-04
Final adjustments for entso project

marliekeverweij:
See commit messages :)


datasets/BEGM11002_antwerpen/BEGM11002_antwerpen.derived.ad
```diff
@@ -134,7 +134,7 @@
 - electric_vehicle_profile_5_share = 0.0
 - solar_pv_profile_1_share = 1.0
 - solar_pv_profile_2_share = 0.0
-- annual_infrastructure_cost_gas = 4.988546861845945e-29
+- annual_infrastructure_cost_gas = 29099857.79
```
mabijkerk, line 137: Based on #2662 I thought that gas costs had only been decreased by 1e6. Instead I'm seeing that the values have been changed altogether in most datasets. Why is this the case?

marliekeverweij: First, the values were based on the EU gas network costs scaled by number of residences (which were also based on the NL costs) and now they are directly based on the NL costs scaled by the total gas demand. So they have improved :)

PR #2677, opened 2022-04-06
Noord holland update 2019

Charlottevm:
Updating Noord Holland to 2019, steel sector is similar to steel sector NL2019


MartLubben (approved):
LGTM!

PR #2678, opened 2022-04-06
Vans should be included in the chart 'emissions_transport'

Charlottevm:
Vans were not included in the previous query


MartLubben:
If I understand it correctly. The former query did not include vans in the CO2 emissions, and this commit adds the vans to the query. 
@Charlottevm

Charlottevm:
Yes, it does indeed! @MartLubben Thanks for the clarification

PR #2692, opened 2022-05-04
European datasets - improve CHP capacities with data from Eurostat

marliekeverweij:
Update European datasets:
- improve CHP capacities with data from Eurostat
- fix PP sliders; setting the full load hours for power plants that are not running to a default value, to enable the sliders in the ETM front-end
- remove redundant SGQ


mabijkerk (approved):
🇪🇺

PR #2693, opened 2022-05-13
Add EU27 2019 dataset

marliekeverweij:
+ several updates on EU country datasets (refinery output, housing types, inhabitants)


mabijkerk:
I have the following remarks/questions:

- A lot of values have been changed for GM1895_oldambt and it suddenly has country as its group; this seems unintentional and should probably be reverted to municipality as it now shows up in the country dropdown on ETModel:

Screenshot 2022-05-17 at 09 39 42

- The EU27 dataset name seems rather long; for readability EU27_european_union would be better.
- The EU27 dataset now shows up twice in the country dropdown on ETModel, the old one should be removed and a flag should be added to the new one:

Screenshot 2022-05-17 at 09 42 25

- When I open a blank scenario for EU27, the final demand for households in 2050 is higher than in 2019, which shouldn't be possible, since I haven't made any changes. The difference seems to be in hot water and heat. The issue also occurs for other datasets on beta and pro, so I'll open a separate issue for this:

Screenshot 2022-05-17 at 09 59 24

- Where do the fuel prices and interconnector price come from?

- There are no terraced house and no corner houses in EU27 according to our dataset; is this correct?

marliekeverweij:
Thanks for your feedback!

> A lot of values have been changed for GM1895_oldambt and it suddenly has country as its group; this seems unintentional and should probably be reverted to municipality as it now shows up in the country dropdown on ETModel

There is already an[ issue ](https://github.com/quintel/etlocal/issues/394)for this and I manually changed it on master for now: https://github.com/quintel/etsource/commit/f7d935dd5528bb10a5bece2442b1ecf047fda906

> The EU27 dataset now shows up twice in the country dropdown on ETModel, the old one should be removed and a flag should be added to the new one

The old one is disable in this commit: https://github.com/quintel/etsource/pull/2693/commits/003898cde423bb5998c89581af9568918bef51d4 Interesting that you still see it. I have already added the flag on the ETModel branch

> Where do the fuel prices and interconnector price come from?

Good question, these are all derived from the old eu dataset..... I think we should update them. @AlexanderWirtz @MartLubben what do you think?

> There are no terraced house and no corner houses in EU27 according to our dataset; is this correct?

Yes Eurostat does not report about those housing types

MartLubben:
>> Where do the fuel prices and interconnector price come from?

>Good question, these are all derived from the old eu dataset..... I think we should update them. @AlexanderWirtz @MartLubben what do you think?
I agree with updating. 

I think we should be very careful with all data used from the old eu dataset, because it is very old. Do we use the old eu dataset a lot?  And if so, for which assumptions?

mabijkerk:
> The old one is disable in this commit:

It is enabled again in this commit: https://github.com/quintel/etsource/pull/2693/commits/12e461b963f57f7a2dca717c063d8b3474b701f7

marliekeverweij:
> @MartLubben I think we should be very careful with all data used from the old eu dataset, because it is very old. Do we use the old eu dataset a lot? And if so, for which assumptions?

Most of the data is now dataset-specific, we have ensured this during the project when we added the EU datasets. Only fuel prices and import prices cannot be set in ETLocal yet, that's why we use the old eu dataset values. These differ from the prices we have in `nl2019`. I would propose to use the same values as in `nl2019` for all eu dataset and this new eu27 dataset. What do you think @MartLubben @mabijkerk @AlexanderWirtz ?

PR #2701, opened 2022-06-02
EU countries: update energy flows renewable plants

marliekeverweij:
- Improved Excel analysis of renewable power plants (see: 20220511_EU_power_plants.xlsx on ETDataset)
- Updated Eurostat table with renewable power plant capacities [NRG_INF_EPCRW__custom_1553298]


Charlottevm (approved):

mabijkerk (changes requested):
I think there still is a mistake in the renewables calculation. Looking at Germany for example, there suddenly is no offshore wind power anymore, which does not seem right.

Looking at the Excel source file in the sheet `source renewable - Eurostat` each country in the rows `Summing for power plants in the ETM` seems to link to the wrong column in the rows `ETM categories`. Denmark refers to Germany, Germany to Estonia etc. Each country is draws it source data from the country on the left.

Screenshot 2022-06-03 at 10 04 55

marliekeverweij:
Sharp @mabijkerk ! I only do not have time to fix this before the deploy today

mabijkerk (approved):

PR #2714, opened 2022-07-15
Updated slider max for electricity and heat plants

DorinevanderVlies:
For electricity plants the slider max is now equal to twice (the installed capacity required to supply) total electricity demand in the present. For heat plants the slider max equals twice the (installed capacity required to supply) total heat production in the present.


mabijkerk (approved):
I fixed a minor syntax error for hydro mountain and updated the biomass power plant inputs to align with the other power plants.

PR #2718, opened 2022-07-26
Updating foodvalley to 2019

Charlottevm:


Charlottevm:
The old foodvalley dataset seems to have incorrect numbers for inhabitants / buildings etc. (after a small check online). Therefore I am assuming most of the data is in some way corrupted and therefore we can merge this version.

PR #2719, opened 2022-07-26
Foodvally update to 2019

Charlottevm:
The data is very different, it seems the older dataset is faulty as e.g. the number of inhabitants is correct in the updated version.


MartLubben:
I think something went wrong in creating this dataset. I see quite some major changes compared to the former dataset. I check the number of inhabitants with Klimaatmonitor and they differ quite a lot (both 2019): 

image
image

MartLubben (approved):

PR #2728, opened 2022-08-08
Update Northern Ireland dataset

marliekeverweij:
This PR updates:
- oil & coal demands in all sectors
- update of shares in transport & households

Both due to update of Department for Business, Energy & Strategy (BEIS)


mabijkerk:
When I try to run the model while on this branch for ETSource, I get the following error:

Screenshot 2022-08-08 at 09 54 42

PR #2729, opened 2022-08-08
Northern-Ireland update

marliekeverweij:
This PR updates:

- oil & coal demands in all sectors
- update of shares in transport & households

Both due to update of Department for Business, Energy & Strategy (BEIS)


mabijkerk (approved):
✨

PR #2730, opened 2022-08-11
Addition of load shifting DSR for the central ICT

mabijkerk:
This PR adds load shifting for DSR for the central ICT industry:

- Addition of nodes to backend
- Input sliders
- Addition of load shifting to relevant charts and tables

Corresponds with https://github.com/quintel/etmodel/pull/3994


marliekeverweij:
The load shifting of ICT is very small. I had to increase the ICT sector by 3000% to see something significant (see screenshot below).

@ChaelKruip do you have time this week to also test it? I think it is good to double-check if this is logical.

Screenshot 2022-08-17 at 09 36 36
Screenshot 2022-08-17 at 09 37 54

ChaelKruip:
> @ChaelKruip do you have time this week to also test it? I think it is good to double-check if this is logical.

Sorry, my week is already too full. Perhaps some other testers can be found?

mabijkerk:
@marliekeverweij I think it is as expected: if you look at the hourly ICT demand in a blank nl2019 scenario, the peak capacity is somewhere around 500 MW (see screenshot below). The maximum capacity you can therefore allocate to DSR load shifting amounts to 50% * 500 MW = 250 MW. When you look at your chart showing the the flexible electricity demand per hour, the value typically range between GWs of capacity. It is therefore not strange that a 250 MW demand is hardly visible.

Screenshot 2022-08-22 at 10 31 48

At a size of 3000% and a yearly efficiency improvement of 3%, the peak capacity is around 6 GW. I would therefore expect the capacity available for DSR load shifting in your scenario to be around 3 GW.

marliekeverweij:
Yes that's the case! 

Flexible electricity demand per hour 1654766

marliekeverweij (approved):

PR #2736, opened 2022-08-23
Dataset Almaty with data provided by Witteveen + Bos

Charlottevm:
The dataset is conceptual / lacking on multiple levels:
- Electricity production is significantly higher than electricity demand (demand and production are not correctly balanced)
- Only demand is filled in, other data is provided for by the Polish dataset
- Some values are set to 0.0 (such as the 1990 emissions and interconnection capacity) 

The validity of the data has also not been checked by Quintel. For that reason the dataset is not visible in the ETM and should stay hidden on ETModel. If the dataset should be removed contact Laurens Frowijn from Witteveen + Bos (laurens.frowijn@witteveenbos.com).


noracato (approved):
:+1:

PR #2740, opened 2022-08-25
Fix incorrect flh for wind turbines in Sweden and UK

marliekeverweij:
Closes https://github.com/quintel/etmodel/issues/4000


mabijkerk (approved):
Thanks for spotting the SGQ for DSR ICT. LGTM!

PR #2744, opened 2022-08-31
Update specifications of nuclear power plants

mabijkerk:
This loads the updates from https://github.com/quintel/etdataset/pull/939 for the 3rd generation and SMR nuclear power plants.

It also removes outdated attributes from the 2nd generation nuclear power plant.


AlexanderWirtz:
@mabijkerk will these updates not result in changes to existing scenarios? Among others because the availability of the 3rd Gen nuclear plants is now slightly lower? If so, is there a migration to fix this?

mabijkerk:
@AlexanderWirtz that's right. I think we can write a migration that changes the installed capacity (to compensate for the lower availability) and the costs.

marliekeverweij:
Checking the efficiencies:
- For the Netherlands the efficiency of the 2nd & 3rd gen are both 37%, is that correct?
- For EU countries we calculate the efficiency in ETLocal and we use the following default values: 32% for 2nd generation and 36% for 3rd generation.
```
      - key: input_energy_power_nuclear_gen2_uranium_oxide_electricity_output_conversion
        unit: '%'
        entso: |
          IF(EB("Nuclear conventional", "Nuclear heat") > 0.0,
          -> { EB("Nuclear conventional", "Electricity") / EB("Nuclear conventional", "Nuclear heat") * 100 },
          -> { 32.0 })
      - key: input_energy_power_nuclear_gen3_uranium_oxide_electricity_output_conversion
        unit: '%'
        entso: 36.0
```

mabijkerk:
For the nl2019 dataset the numbers are based on Borssele which is a refurbished 2nd generation nuclear power plant:
Screenshot 2022-09-01 at 11 37 08

For the EU I am a bit surprised to find hard-coded numbers in the ETLocal final. These efficiencies seem to be taken from the nl2015 dataset. I don't think it is very consistent to use a 32% efficiency for the 2nd generation since we note a 33% in the [source file](https://github.com/quintel/etdataset-public/blob/master/eu_datasets/power_plants/20220609_EU_power_plants.xlsx). The technology factsheet for the 3rd generation notes a thermal efficiency of 37%.

**Conclusion**
- We could update the efficiencies from 32% and 36% to 33% and 37% but I'm not sure if it is worth the effort
- Additionally, I feel like at some point the documentation of these efficiencies should be improved, since the comment in the 'version history' on ETLocal makes no mention of them coming from the nl2015 dataset

What do you think @AlexanderWirtz and @marliekeverweij

AlexanderWirtz:
> For the nl2019 dataset the numbers are based on Borssele which is a refurbished 2nd generation nuclear power plant:

This is a very easy tweak: the power plant analysis simply compares the electricity produced with a certain installed capacity to the actual numbers from Eurostat. The explanation for the higher efficiency is indeed that it is a refurbished 2nd Gen plant.

> - We could update the efficiencies from 32% and 36% to 33% and 37% but I'm not sure if it is worth the effort

Let's wait for someone to notice and pay for it. 

> - Additionally, I feel like at some point the documentation of these efficiencies should be improved, since the comment in the 'version history' on ETLocal makes no mention of them coming from the nl2015 dataset

Agree.

marliekeverweij (approved):

PR #2750, opened 2022-09-08
Re export of flh for wind in SE and UK

mabijkerk:


marliekeverweij:
Accidentally the old flh were deployed. Here we implement [this fix](https://github.com/quintel/etsource/pull/2740) again!

marliekeverweij (approved):

PR #2760, opened 2022-09-30
Enable more curves to be adjusted through API'

marliekeverweij:
Cooling curves
Appliances curves
Agriculture electricity
Households hot water

### DIFF
--- a/config/user_curves.yml
+++ b/config/user_curves.yml
@@ -318,6 +318,18 @@ interconnector_12_export_availability:
   disables:
     - electricity_interconnector_12_export_availability
 
+agriculture_electricity:
+  type: profile
+  display_group: agriculture
+
+buildings_appliances:
+  type: profile
+  display_group: buildings
+
+buildings_cooling:
+  type: profile
+  display_group: buildings
+
 weather/buildings_heating:
   type: profile
   display_group: buildings
@@ -327,6 +339,10 @@ electric_buses:
   display_group: transport
 
 electric_trucks:
+  type: profile
+  display_group: transport
+
+electric_vans:
   type: profile
   display_group: transport
 
@@ -349,6 +365,18 @@ electric_vehicle_profile_4:
 electric_vehicle_profile_5:
   type: profile
   display_group: transport
+
+households_appliances:
+  type: profile
+  display_group: households
+
+households_cooling:
+  type: profile
+  display_group: households
+
+households_hot_water:
+  type: profile
+  display_group: households
 
 industry_chemicals_electricity:
   type: profile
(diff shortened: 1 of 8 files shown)


antw (approved):

antw:
@marliekeverweij I've reverted and reopened this. I think the queries in this PR are why the chart in ETModel does not work. I don't understand what is meant by this line in the queries:

```ruby
"electricity_input_curve + hydrogen_input_curve + network_gas_input_curve + steam_hot_water_input_curve"
```

This is going to take the electricity curve, H2 curve, NG curve, and hot water curve, then concatenate them together which is probably not what you intended. For example, if you have an electricity curve `1, 2, 3` and NG curve of `4, 5, 6` you'll end up with `1, 2, 3, 4, 5, 6`.

Can you explain what you're trying to do here?

marliekeverweij:
Okay! Thanks for looking into this.

I just copy&pasted the curve queries that were already there, I assumed they would be correct. The [`buildings_heating_curve.gql`](https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/user_uploaded_profiles/buildings_heating_curve.gql) was already there and was working, so I expected the others to work aswell.

I want to show all curves in that chart. It is however not possible to query the original curve from the dataset. Therefore, these queries take the hourly input curves of all technologies, in case of `buildings_heating_curve.gql` it takes all input curves of all heating technologies in the buildings sector.

antw:
Thanks for explaining. I was expecting them to be written [like this one](https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/user_uploaded_profiles/industry_other_heat_curve.gql). I guess using `+` is fine if you're sure that each node only has one of the four curves.

I think I was also wrong about the cause of the issue. The chart is not happy when the query returns an empty curve, rather than one with values. I think that happens because the `FILTER(...)` is returning no nodes. For example:

```ruby
FILTER(
  ALL(),
	"fever && (fever.group == :buildings_appliances)"
)
```

This returns an empty collection: there are no nodes whose fever group is "buildings_appliances". Valid Fever groups are:

- `buildings_space_heating`
- `space_heating` (households)
- `households_hot_water`

I'm guessing this was copy/pasted from queries that check the merit order configuration, where `group` really means "profile"? Producers in Fever don't have a profile that you can filter on: the curve is generated dynamically based on the total demand for heat in households or buildings, and the curve is assigned to the node by Fever after the calculation is complete.

Did you mean:

```ruby
FILTER(
  ALL(),
	"merit_order && (merit_order.group == :buildings_appliances)"
)
```

PR #2761, opened 2022-10-03
Reset curves for NI

marliekeverweij:
During the addition of the EU datasets, we accidentally removed the NI specific curves from the NI dataset. This PR resets the original NI curves.


marliekeverweij:
I have update both the `solar_pv.csv` as the `solar_pv_profile_1.csv`. Could you check if this is necessary @DorinevanderVlies ?

DorinevanderVlies:
**Comment**: the week and weekend days of the transport curves are misaligned with the other demand curves:

- the other demand curves link to the **2019** curves of NL.
- the transport curves concern **2018**

image

**Note**: the temperature curve has been added, but the space heating curves (households, buildings and agriculture) still link to Dutch curves. I think that we now have the information to also make country-specific curves for space heating.

marliekeverweij:
@DorinevanderVlies good point! I have updated the EV profiles to match with the week / weekends of 2019

DorinevanderVlies (approved):
LGTM

PR #2763, opened 2022-10-03
Enable more curves to be adjusted through API

marliekeverweij:
New PR
Old: https://github.com/quintel/etmodel/pull/4016

### DIFF
--- a/config/user_curves.yml
+++ b/config/user_curves.yml
@@ -318,6 +318,18 @@ interconnector_12_export_availability:
   disables:
     - electricity_interconnector_12_export_availability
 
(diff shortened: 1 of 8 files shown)


marliekeverweij:
I have simplified and expected to fix the queries in my last commit. However, I am still getting an error... Do you also get this error @antw ?
Screenshot 2022-10-03 at 12 16 58

antw (approved):
I fixed the queries: they needed `SUM_CURVES` – that's why you received an error in ETModel.

This looks good to me. I'll let you merge it so that you can give it a try and check that everything looks as you expect.

PR #2766, opened 2022-10-04
Adding Curves, ICT industry and large scale gas CHP to Singapore

Charlottevm:
To finish Singapore we have: 
- added weather curves (this had not been added yet), 
- ran the source_analysis for the ICT sector (we forgot to include this)
- and added large scale gas CHP's (this was wrongly added in ETLocal)


marliekeverweij:
I took a short look at this PR and saw that the curves are not in the correct folder. You can have a look at the `nl2019` dataset to see where each curve should be:
- cooling curves should be in the `curves` folder
- solar & wind should be in the `curves/weather/default/` folder

Additionally, you should add a `solar_pv_profile_1.csv` curve with the values as `solar_pv`, this is necessary for the flh sliders where two curves are interpolated (`solar_pv_profile_1.csv` & `solar_pv_profile_2.csv`). 

Screenshot 2022-10-04 at 10 42 49
Screenshot 2022-10-04 at 10 44 23

mabijkerk (commented):
- Curves: I don't have much experience with these curves so I'll defer to the comment from @marliekeverweij 
- ICT demand: the graph shows a central ICT demand of 8.97 PJ but I'm don't know how to trace whether this is the right number (I would expect it to be in the source analyses input folder of the energy_balance_generator on ETDataset)
- Gas CHP: I still don't fully understand why this CHP is producing (this much) heat, especially since all 81.61 PJ of its heat production goes unused

PR #2767, opened 2022-10-05
World balance

Charlottevm:
Additional changes to Singapore:

- inclusion of gas CHP (resolve mistake in etlocal)
- addition of curves
- addition of ICT sector (added source_analysis)


antw (approved):
@Charlottevm This might be better reviewed by someone familiar with the changes or the Singapore dataset, rather than me.

It looks fine to me though. :+1:

Charlottevm:
Mathijs said it is better to have it reviewed by you as he already reviewd the content but you could review the technical part ;) So this should do, thanks!

PR #2771, opened 2022-10-11
Making curves for e ships & planes modifiable

marliekeverweij:


antw (approved):
Looks good to me! :+1:

PR #2802, opened 2022-11-08
Reverting GM Rhenen back to 2018 as requested by the municipality (ML)

Charlottevm:
The municipality of Rhenen was updated to 2019 on ETLocal but not yet in ETSource.
As the municipality is currently using the dataset it was requested to keep the dataset at 2018 for now.


datasets/GM0340_rhenen/GM0340_rhenen.derived.ad
```diff
@@ -32,31 +32,31 @@
 - households_insulation_employment_constant = 0.014
 - co2_emission_1990_aviation_bunkers = 4.6046
 - co2_emission_1990_marine_bunkers = 34.95619
-- co2_percentage_free = 0.4
-- co2_price = 0.02483
-- offshore_ccs_potential_mt_per_year = 65.0
+- co2_percentage_free = 0.85
+- co2_price = 0.0078
+- offshore_ccs_potential_mt_per_year = 41.95
```
mabijkerk, line 37: This seems like a generic update that applies regardless of the analysis_year. Are you sure that it should be reverted?

datasets/GM0340_rhenen/GM0340_rhenen.derived.ad
```diff
@@ -136,43 +136,43 @@
 - solar_pv_profile_2_share = 0.0
 - annual_infrastructure_cost_gas = 1195421.857
 - areable_land = 21.97
-- co2_emission_1990 = 0.151582735
+- co2_emission_1990 = 0.147750174637452
```
mabijkerk, line 139: This seems like a generic update that applies regardless of the analysis_year. Are you sure that it should be reverted?

mabijkerk (approved):

datasets/GM0340_rhenen/GM0340_rhenen.derived.ad
Charlottevm: This has not been changed by us but due to more recent figures of the data-source. I'll check why and how this changes..

PR #2813, opened 2022-11-30
Implements `hourly_supply_electricity_from_renewables` and `hourly_share_electricity_from_renewables`

thesethtruth:
To discuss:
- [x] Is this the correct implementation or should I use `sustainability_share`?
- [x] Have I included the right technologies (that should be considered renewables)?
- [x] How to deal with import?
- [x] How to deal with curtailment/flexibility?

To implement:
- [ ] Correctly determine the share of renewability of the waste (it's a mix of biogenic & non-biogenic)


thesethtruth:
Notifying @noracato and @marliekeverweij.

We want to use such a curve in https://github.com/ZEnMo/Holon-webapp, for now I will just implement this query at our side.

marliekeverweij:
>Is this the correct implementation or should I use sustainability_share?

Yes, this is the correct implementation. The only risk is in the supply query... if we add new carriers, updating that query could easily be forgotten. But cannot think of another way!

> Have I included the right technologies (that should be considered renewables)?

Almost, only waste is not renewable at all times. It's a mix of biogenic & non-biogenic, so we have to do the same trick as for the power plants using network gas: spllitting up the queries into electricity from green gas and electricity from natural gas. 

I could pick this up. What is the urgency of this improvement?

> How to deal with import?

Good question. In the ETM we consider import as not renewable because on a yearly basis there are always emissions related and we do not have information about the renewable fraction per hour.

>How to deal with curtailment/flexibility?

I do not think we have to deal with this, as we are looking at the electricity production

thesethtruth:
Thanks for the swift reply @marliekeverweij!

1. Agreed on the risk of adding new carriers, we'll just have to keep an eye open for new renewable technologies!
2. I've created a task based on your comment on the waste not being completely renewable. Implementation can wait, I'll track it a tech debt over at https://github.com/ZEnMo/Holon-webapp/issues/175. For now, it's implemented as multiple queries that are processed in the module itself.
3.  Nice. Let's not deal with import in that case.
4. My train of thought was that if renewable energy is being produced in each of those hourly curves but subsequently curtailed due to supply overload, it should not be considered part of the renewable fraction of production. But including the hourly curtailment for each non dispatchable technology might be complicated?

marliekeverweij:
1. Yess!
2. Alright!
3. 👍
4. Thanks for sharing your thoughts. Tracking the hourly curtailment of each volatile power plant is indeed not easy and something that's not possible at the moment in the ETM. If this is something that's desired?

thesethtruth:
> Thanks for sharing your thoughts. Tracking the hourly curtailment of each volatile power plant is indeed not easy and something that's not possible at the moment in the ETM. If this is something that's desired?

I think that for our current use case it is fair enough to approximate it without tracking curtailment, considering that it would be more complicated to properly implement that. However, I think it would be right to add a disclaimer (comment in the `gql`?) that states this limitation. It would be possible to generate an ETM scenario that is utterly overloaded on volatile renewable energy sources which might skew the share of renewables as a result. 

Perhaps we should also add a comment that disclaims the dependency on manually adding all renewable energy carriers in the summing query.

thesethtruth:
> It would be possible to generate an ETM scenario that is utterly overloaded on volatile renewable energy sources which might skew the share of renewables as a result.

Actually, thinking about this again, it might not hold due to the copper plate approach in the ETM? The reason I was thinking of simultaneous curtailment and production was due to local net contraints (e.g. a solar plant is outputting power but is being curtailed because of local (LV, MV) network constraints. Correct me if I'm wrong, but that can't happen in the ETM?

Then; curtailment would only occur when the production out sizes the demand. This impacts the consumption of renewable energy (since not all is consumed because a part is curtailed) but does not affect the share of renewable energy, it would remain at 100% renewable.

The only exception I can think of is that of must-run energy sources. What is the 'merit' or curtailment? Could it occur that volatile renewables are curtailed in favor of keeping must-run sources running? In that case, the share of renewables would be overestimated because of curtailment.

marliekeverweij:
>Actually, thinking about this again, it might not hold due to the copper plate approach in the ETM? The reason I was thinking of simultaneous curtailment and production was due to local net contraints (e.g. a solar plant is outputting power but is being curtailed because of local (LV, MV) network constraints. Correct me if I'm wrong, but that can't happen in the ETM?

True! All volatile producers can deliver their production to the net. However, you can do curtailment / overplanting in the ETM: install more capacity than the grid connection (see [here](https://pro.energytransitionmodel.com/scenario/flexibility/flexibility_net_load/curtailment-solar-pv)). This does not influence the renewability fraction as this type of curtailment is already subtracted from production.

>Then; curtailment would only occur when the production out sizes the demand. This impacts the consumption of renewable energy (since not all is consumed because a part is curtailed) but does not affect the share of renewable energy, it would remain at 100% renewable.

Yes.

>The only exception I can think of is that of must-run energy sources. What is the 'merit' or curtailment? Could it occur that volatile renewables are curtailed in favor of keeping must-run sources running? In that case, the share of renewables would be overestimated because of curtailment.

Volatile producers and must-run producers both deliver all their electricity to the net. If the total production exceeds total demand, then curtailment takes place; we sometimes call this type of curtailment last-resort curtailment. I am not sure if this should impact the renewability fraction of produced electricity..

thesethtruth:
@mabijkerk Previously I talked with @marliekeverweij about this issue. Is it still something you would like to implement? I could try to update the PR?

mabijkerk:
Thanks for notifying me about this PR @thesethtruth. I can pick this up, but I do have two questions:

**1. Imports of electricity**
- I saw your discussion about excluding imported electricity in the selection of renewables. This seems fine to me.
- However, if you're interested in the share of renewable production in the **total** electricity supply at every hour, then I think you **should** include import in the total electricity supply.
- In other words: import should be added to the denominator in the [hourly_share_electricity_from_renewables](https://github.com/quintel/etsource/pull/2813/files#diff-eae4b63e61f6d40089ae9cb1cd16ff6ab8bcdbd44bbbc7e40579c42ae327a0c3) query.
- Do you agree?
- In that case, the query `hourly_production_electricity` should be replaced by `hourly_total_electricity_production`

**2. Waste** 
- If I read the discussion correctly, the waste still needs to be split into biogenic and non-biogenic is that right?

thesethtruth:
> In other words: import should be added to the denominator in the [hourly_share_electricity_from_renewables](https://github.com/quintel/etsource/pull/2813/files#diff-eae4b63e61f6d40089ae9cb1cd16ff6ab8bcdbd44bbbc7e40579c42ae327a0c3) query.
Do you agree?

Right, that makes sense! 

> If I read the discussion correctly, the waste still needs to be split into biogenic and non-biogenic is that right?

Correct. If you can point me in the right direction I might be able to do this myself. (an example perhaps?)

mabijkerk:
I have added two queries to `master` which you can now use in your `hourly_supply_electricity_from_renewables` query instead of the generic waste query. You can check the commits that reference this PR for more information.

- [hourly_supply_electricity_from_biogenic_waste](https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/vertical_stacked_bar_electricity_mix_power_to_gas/hourly_supply_electricity_from_biogenic_waste.gql)
- [hourly_supply_electricity_from_non_biogenic_waste](https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/vertical_stacked_bar_electricity_mix_power_to_gas/hourly_supply_electricity_from_non_biogenic_waste.gql)

mabijkerk:
Two final notes on this:

- Since hydrogen is not produced from renewable sources by default, should we implement it in this way?
- Is uranium actually recognized as renewable?

thesethtruth:
I'm sure at some point we'll find the time to implement this!

Issue #2841, opened 2023-01-31
The nl2019 dataset refers to old (2015) electricity and heat demand curves for the other industry sectors


redekok:
One of our clients pointed out that our electricity demand curves for other industry seem to start on a different weekday (a Thursday) compared to our other demand curves (which start on a Tuesday). After looking in to this, I found out that the `nl2019` dataset refers to old curves from the 2015 dataset: [`industry_other_electricity.csv`](https://github.com/quintel/etsource/blob/master/datasets/nl2019/curves/industry_other_electricity.csv) and [`industry_other_heat.csv`](https://github.com/quintel/etsource/blob/master/datasets/nl2019/curves/industry_other_heat.csv). I noticed that there is [only one curve available for 2019](https://github.com/quintel/etdataset/blob/master/curves/demand/industry/other/data/nl/2019/output/industry_other.csv) in our [etdataset repository](https://github.com/quintel/etdataset/tree/master/curves/demand/industry/other). This curve represents the heat demand. No curve is available for the electricity demand. This is probably due to the fact that our data source ([NEDU](https://www.nedu.nl/cgi-sys/suspendedpage.cgi)) isn't available anymore. The 2019 [`industry_other.csv`](https://github.com/quintel/etsource/blob/master/datasets/nl2019/curves/industry_other.csv) curve has been exported to our nl2019 dataset on ETSource but isn't actually used at the moment.

[MFFBAS](https://www.mffbas.nl/documenten/) also offers the NEDU demand curves. However, they only offer curves starting from 2020. Curves from 2019 doesn't seem to be (publicly) available. In addition, I read [something](https://www.acm.nl/sites/default/files/documents/2019-06/ontwerp-codebesluit-uitfaseren-profielsegment-e3.pdf) about phasing out the E3A, E3B, E3C, and E3D curves and aggregating those to an overarching E3 curve. For the last couple of years MFFBAS still seems to offer the more detailed E3A-D curves so this might not be a very urgent issue we have to deal with. 

@mabijkerk and I have shortly discussed a few solutions for this issue:

1. For lack of 2019 curves, use the 2015 curves and shift those a few days to ensure they start on a Tuesday instead of a Thursday. 
2. Check with MFFBAS whether it's possible to get the NEDU 2019 curves. If not, we could also check this with other parties such as the system operators.
3. Use the 2019 heat demand curve (which is derived from the gas balance) for both electricity and heat demand.

[...]

Issue #2842, opened 2023-01-31
Should all curves start on a Monday?


redekok:
Currently, the curves used in the ETM always start on the weekday that corresponds with January 1st of the dataset base year. In 2019, January 1st corresponds with a Tuesday. Hence, the first weekday of all 2019 curves is a Tuesday. 

Recently, one of our clients asked us whether it would be possible to have all ETM curves start on a Monday. This would better match their process and definitions. That made me wonder whether this could benefit our workflows and processes as well. 

Some benefits are:
1. Make the ETM more compatible to other (external) processes 
2. Make the workflow around curves more transparent: it would for instance be easier to compare curves for different base years
3. Make the workflow around curves less error prone: whenever datasets refer to curves of other datasets, we don't have to worry about the base year and starting weekday (also see https://github.com/quintel/etsource/issues/2841)
4. In addition to 3, make it easier to use/exchange curves for other datasets (with other base years)

The main disadvantage would be the time we'd have to invest in 1) modifying all scripts (i.e., adding a function to shift all curves and have them start on a Monday) and 2) updating the curves for all datasets.

Personally, I'm not sure if the benefits outweight the time investment (yet) but very curious to hear @quintel/core's opinion on this! Also feel free to elaborate on the pros and cons please. :)

mabijkerk:
**Reflections**
I think it is only relevant for our demand curves e.g. services electricity demand. This is because these curves will show the week/weekend pattern, whereas supply curves e.g., solar radiation, don't take week/weekend patterns into account.

If only demand curves are in scope I would not expect the effort of creating a shifting script to be too high. It may depend on whether we want to implement it as a function in every specific curve script, or add it as a separate generic module.

**Concerns**
A concern I do have is not creating the script itself, nor updating the curves once we have the script, but rather to verify that we actually have the proper source year for each dataset. For example, if we were to create this script now, and we wouldn't have known about https://github.com/quintel/etsource/issues/2841, then we would have shifted a 2015 curve with x days, even though it should be shifted with y days.

Another concern, which to me seems most relevant, is that the curve start value and end value will probably not align. I would expect that simply shifting the curve would lead to a trendbreak at the joint.

redekok:
Thanks for sharing your thoughts @mabijkerk!

Hereby a quick response to some of your reflections and concerns:

[...]

PR #2845, opened 2023-02-06
Make all nl2019 curves match the 2019 weekly rhythm

redekok:
- Update industry_other_heat.csv (based on 2019's gas balance) and industry_other_electricity.csv (based on 2019's NEDU E3D curve)
- Shift curves for (electric) vehicles other than cars 2 days backwards in order for the curves to start on a Tuesday instead of a Thursday. In this way, the curves match the 2019 weekly rhythm instead of the 2015 one

Closes https://github.com/quintel/etsource/issues/2841
Goes with https://github.com/quintel/etdataset/pull/966


mabijkerk:
I don't see the electric vehicle profiles in ETDataset. @redekok do you think it would it be good to add them to ETDataset for transparency? Preferably we would add a README that describes the source and the fact that we shifted them 2 days.

Another point (or debt) to take into account (as noted https://github.com/quintel/etsource/issues/2842#issuecomment-1416082924 here), is that shifting the days does cause a misalignment for public holidays. We could add that to the README as well.

mabijkerk (approved):
LGTM!

redekok:
> I don't see the electric vehicle profiles in ETDataset. @redekok do you think it would it be good to add them to ETDataset for transparency? Preferably we would add a README that describes the source and the fact that we shifted them 2 days.

I actually forgot about this but we re-use electric car charging curves by Movares for the other electric vehicles. We could add a short phrase to ETDataset but our documentation also this describes this, see https://docs.energytransitionmodel.com/main/curves/. 

> Another point (or debt) to take into account (as noted https://github.com/quintel/etsource/issues/2842#issuecomment-1416082924 here), is that shifting the days does cause a misalignment for public holidays. We could add that to the README as well.

As for public holidays, we don't take those into account at all when creating curves ourselves. For instance, these Movares curves are built/modelled from scratch based on day and week patterns. Same goes for the space heating curves in households. I do think it's worth it to have that explained somewhere (in a somewhat more general description since it goes for all curves that we model ourselves). But perhaps this could be part of a greater effort around updating (the transparency) around our curves?

PR #2846, opened 2023-02-09
Update EU datasets with updated rake task that sorts graph values alphabetically

mabijkerk:
The ETSource export rake task on ETLocal now sorts the graph_values of datasets alphabetically. I have first updated only the EU datasets with this updated rake task. For more background see https://github.com/quintel/etlocal/issues/456. 

It is essential that this PR produces no changes in **values** for the EU datasets. I therefore wanted to verify whether only the **order** of graph_values had been changed. I therefore checked the changes in graph_values for a random selection of datasets (AT, DE, FI, IT and LV). I did this by removing all the spaces, which creates single row elements, and then sorted and compared the file before and after the update in Excel.

For the selected datasets only the order of graph_values has been changed and `agriculture_chp_engine_network_gas_must_run` has been added to the file. Both changes are as expected (see https://github.com/quintel/etsource/commit/1d32738d994b428e36030686c843e5ba311778e0 for the agriculture addition). A comparison between alphabetically sorted elements produced no further differences.

@redekok would you be able to review this PR? Please let me know if you think that a sample of 5/29 datasets is enough to merge this. I'm open to ideas on how to quickly check 29/29 datasets.


mabijkerk:
I have also tested for CY, EE, EU27 and MT. Only expected changes occurred. FYI - the rake task first sorts nodes in alphabetical order, then edges in alphabetical order.

PR #2873, opened 2023-03-30
Minor fixes for the energy flows (overview) data export

mabijkerk:
This PR includes two fixes:

- Direct electrolysis of solar electricity referred to the wrong node
- Curtailment of solar and wind electricity for direct electrolysis was not included in curtailment query and therefore not included in electricity to electricity losses. @redekok what is you assessment, am I right to include it?


redekok (approved):
LGTM!

PR #2876, opened 2023-03-31
Include additional nodes in curtailment

mabijkerk:
Apart from the 'main' curtailment node, `energy_flexibility_curtailment_electricity`, additional curtailment nodes have been added over the past few years: `energy_battery_curtailed_solar_electricity`, `energy_battery_curtailed_wind_electricity`, and `energy_hydrogen_curtailed_electricity`.

In our primary demand queries however, we still regularly only query the main curtailment node. This PR addresses that by adding a `curtailment_group` to these nodes and querying that group instead of only the main node.

Questions to discuss:

- [x] Is it desirable to query all the curtailment nodes, or was querying the single node intentional?
- [x] How does curtailment work for direct electrolysis? In principle all electricity should be converted to hydrogen, so why is curtailment applied?
- [x] How does **last resort** curtailment work for the battery systems? They have their own curtailment which is activated when the battery is full and the capacity to the grid is reached. So what happens, if their total production exceeds total demand in a scenario? Last resort curtailment needs to be applied. How is this handled?
- [ ] If we choose to use the node group and query all curtailment nodes, are there more queries that should be updated than the ones I have tackled so far?

Notifying @noracato and @ChaelKruip

I will discuss this with @redekok and decide on next steps

### DIFF
--- a/gqueries/general/primary_demand/primary_demand_of_biofuels_from_curtailment.gql
+++ b/gqueries/general/primary_demand/primary_demand_of_biofuels_from_curtailment.gql
@@ -3,9 +3,9 @@
 - unit = PJ
 - query =
     SUM(
-       V(energy_flexibility_curtailment_electricity, primary_demand_of_biodiesel),
-       V(energy_flexibility_curtailment_electricity, primary_demand_of_bio_ethanol),
-       V(energy_flexibility_curtailment_electricity, primary_demand_of_bio_kerosene),
-       V(energy_flexibility_curtailment_electricity, primary_demand_of_bio_lng),
-       V(energy_flexibility_curtailment_electricity, primary_demand_of_bio_oil)
+       V(G(curtailment_group), primary_demand_of_biodiesel),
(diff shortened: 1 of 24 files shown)


mabijkerk:
- Querying the last resort curtailment is intentional, if not, you're missing out on primary demand
- Querying the integrated battery curtailment is debatable, it depends where you want to allocate primary demand i.e. which flows you want to make explicit. **Action**: discuss with relevant people
- Direct electrification curtailment is likely implemented because the input capacity of the electrolyser is not dimensioned to exactly fit the output capacity of the solar/wind. **Action**: research whether this is the case
- Querying the direct electrification curtailment is debatable, similar to the integrated battery curtailment, the question is whether you want to allocate primary hydrogen demand or allocate primary electricity. **Action**: discuss with relevant people

kndehaan:
This PR requires more effort in solving and is therefore rewritten into an [issue](https://github.com/quintel/etsource/issues/3100).

PR #2893, opened 2023-05-03
Set specific edges from share to flexible to generate fallback shares

mabijkerk:
**Issue**
In a number of industry sub sectors and for a number of transport applications, all edges that are set by inputs in the frontend were set to `share`. In datasets with 0.0 MJ demand for those sub sectors or transport applications these inputs would all be 0.0%. Resetting any of the inputs would cause the model to break. See screenshots below for an example.

**Solution**
For each of the affected sub sectors and transport applications, a single edge is set to `flexible`, which forces that edge to adopt a 100% share when no demand is present. The edge that has the most demand in `nl2019` is chosen.

Closes https://github.com/quintel/etmodel/issues/4107
Closes https://github.com/quintel/etlocal/issues/378

**Example of issue for Aarhus dataset**

This in the frontend:
Screenshot 2023-05-03 at 17 04 27

Is caused by this in the backend:
Screenshot 2023-05-03 at 17 05 27

### DIFF
--- a/graphs/energy/edges/agriculture/agriculture_network_gas_hybrid_heat-agriculture_useful_demand_useable_heat@useable_heat.ad
+++ b/graphs/energy/edges/agriculture/agriculture_network_gas_hybrid_heat-agriculture_useful_demand_useable_heat@useable_heat.ad
@@ -1,2 +1,2 @@
-- type = share
+- type = flexible
 - reversed = false
--- a/graphs/energy/edges/industry/industry_chemicals_other_burner_crude_oil-industry_useful_demand_for_chemical_other_useable_heat@useable_heat.ad
+++ b/graphs/energy/edges/industry/industry_chemicals_other_burner_crude_oil-industry_useful_demand_for_chemical_other_useable_heat@useable_heat.ad
@@ -1,3 +1,3 @@
-- type = share
+- type = flexible
 - reversed = false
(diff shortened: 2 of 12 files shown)


Charlottevm:
Thanks for adding these @mabijkerk !

I have approved the PR because it fixes all our current problems.
However, should we do this for all shares?
The reason I ask this is because countries like Singapore as well as future countries like Australia & UAE do not have a heating demand in e.g. agriculture or the built environment. This means the model could crash there as well. 
What do you think @mabijkerk ? How much time would something like this take and are there cons for this approach?

Notifying @ChaelKruip, we could include this effort (+/- 1 day?) for the UAE proposal to prevent the model from crashing once we have created a dataset.

Charlottevm (approved):

Charlottevm:
Just checked, I can only find non-flexible shares for agriculture heat demand so it is quite easily fixed. Shall I change those within this PR as well @mabijkerk ? Seems like we are all set after that

mabijkerk:
> However, should we do this for all shares?

It seemed to me that we had tackled most edges. So if you have spotted any for which this is not the case, please let me know. The effort to change it is minimal.

> Just checked, I can only find non-flexible shares for agriculture heat demand so it is quite easily fixed.

Well-spotted. Feel free to add it to this PR. It can be merged after that final change.

PR #2898, opened 2023-05-10
Remove redundant sustainable attribute from network_gas carrier file

### LINKED ISSUE #2897: Review sustainable attribute in network_gas carrier file
The carrier file of `network_gas` has the attribute `sustainable = 0.0`. However, `network_gas` is a mix of `greengas` (sustainable 1.0) and `natural_gas` (sustainable 0.0). We should verify whether we can safely remove the attribute from the [network_gas.ad](https://github.com/quintel/etsource/blob/master/carriers/network_gas.ad) file.

mabijkerk:
The carrier `network_gas` only exists as a blend of `greengas` and `natural_gas`. Nowhere in the graph is supplied as a form of primary demand. Therefore, its sustainability is always determined by the blended components and the `sustainable` attribute is redundant.

I have removed attribute, and tested it for a blank nl2019 scenario and for II3050 National governance. I both cases, the engine computes normally and there is no difference in dashboard parameters between local and beta.

Closes #2897

### DIFF
--- a/carriers/network_gas.ad
+++ b/carriers/network_gas.ad
@@ -1,2 +1,2 @@
-- sustainable = 0.0
 - infinite = false
+- graphviz_color = grey


ChaelKruip (approved):
LGTM!

PR #2906, opened 2023-05-19
Update electricity storage costs

mabijkerk:
In https://github.com/quintel/etsource/pull/2894 the possibility to adjust the relative storage volume was added for all electricity storage technologies. Building on that, a few improvements have been made:

- The maximum relative storage volume is set to 10 hours for li-ion batteries
- For other storage technologies it is kept at the 100 hours value specified for flow batteries
- The costs of all storage technologies are set to scale with volume, which means that the initial investment and costs per MW are set to 0
- For flow batteries the two costs components, one per MWh and one per MW, are maintained
- For pumped hydro and UPHS preferably the costs would also be split into two components, but due to a lack of data availability this has not been done yet
- For a number of electricity storage technologies, this opportunity was used to update the costs

Goes with https://github.com/quintel/etmodel/pull/4115 and https://github.com/quintel/etdataset/pull/968

Debts

- [x] **Blocking**: the relative storage costs of electric cars are ~27 hours, but the slider max has been set at 10 hours. This means that either the maximum value has to be relaxed or the specifications of the electric cars have to be updated
- [x] Open an issue to structurally review all parameters of electricity storage technologies
- [x] Open an issue to split the costs of UPHS and pumped hydro into two components


ChaelKruip (changes requested):
Looks good!

> the relative storage costs of electric cars are ~27 hours, but the slider max has been set at 10 hours. This means that either the maximum value has to be relaxed or the specifications of the electric cars have to be updated

I would suggest relaxing the slider max!

PR #2909, opened 2023-05-23
Geertruidenberg

Charlottevm:
Change in potential for solar panels buildings (as requested) to enable municipality to fill in RES ambition (higher than currently possible).


Charlottevm:
(approved in real life by @noracato )

PR #2910, opened 2023-05-23
Change in potential for solar panels buildings to enable municipality…

Charlottevm:
… to fill in RES ambition (higher than currently possible)


noracato (approved):

mabijkerk:
I would not merge this without at least verifying why the gas infrastructure costs are suddenly lower as well and why the interconnector capacity goes from 60000 MW to 6000 MW.

PR #2913, opened 2023-05-25
P kev2

Charlottevm:
Update of all provinces to 2019


DorinevanderVlies (commented):
I see two noticable things. Please double check them before merging.

1. Groningen: much higher heat demand for buildings than in previous dataset
image

2. Zuid-Holland: high amount of electricity production and own use by gas CHP's
image

image

DorinevanderVlies (approved):

Charlottevm:
Thanks for checking @DorinevanderVlies , the large heat demand in Groningen is probably due to the 'new' waste plant. In reality this heat is probably transferred to other industry but for now we leave it with buildings for comparability with other provinces. The energy own use demand in ZH is indeed due to the CHP's and therefore probably realistic.

PR #2914, opened 2023-05-31
Revert flexible edge in agriculture

mabijkerk:
66417e11ad024f6f0d053c69488ccfdd0c80e460 set a flexible edge in agriculture for https://github.com/quintel/etlocal/issues/378. This passed locally, but produced a caching error on beta. 05f97a741fc5fbd5d896d17290a439dd5644d711 then was implemented as a potential fix, but did not solve the caching error. Given the Deploy to production next week the changes were reverted in ffc1a67cc20a3326ea57c7cc4f9aeabaae7512f6.

This PR resets both changes, to be merged after the Deploy, when there is time for @noracato to investigate the caching error on beta.

PR #2939, opened 2023-08-23
Initial commit for research PR

thomas-qah:
This PR will research why adding a flexible edge in agriculture made the CI fail.

See also: https://github.com/quintel/etsource/pull/2914


thomas-qah:
Seems no longer necessary, see #2914

PR #2947, opened 2023-09-25
Change unit from factor to %

Charlottevm:
Queries had 'factor' as unit whilst the unit should be '%'.

### DIFF
--- a/gqueries/output_elements/output_series/vertical_stacked_bar_49_renewability/share_of_coal_and_derivatives_in_renewability.gql
+++ b/gqueries/output_elements/output_series/vertical_stacked_bar_49_renewability/share_of_coal_and_derivatives_in_renewability.gql
@@ -14,4 +14,4 @@
       Q(geothermal_own_use_in_sectors)
      )
     )
-- unit = factor
+- unit = %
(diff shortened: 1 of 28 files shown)


gqueries/output_elements/output_series/vertical_stacked_bar_49_renewability/share_of_coal_and_derivatives_in_renewability.gql
```diff
@@ -14,4 +14,4 @@
       Q(geothermal_own_use_in_sectors)
      )
     )
-- unit = factor
\ No newline at end of file
+- unit = %
```
KoenvanB, line 17: This should in fact be a factor, not %.

KoenvanB (changes requested):
All approved, except for the one file with a comment!

gqueries/output_elements/output_series/vertical_stacked_bar_49_renewability/share_of_coal_and_derivatives_in_renewability.gql
Charlottevm: Top! Done

KoenvanB (approved):
Approved!

PR #2955, opened 2023-11-01
Update Industry other with non-energetic sliders

KoenvanB:
Added following nodes:

1. industry_useful_demand_for_other_non_specified_hydrogen_non_energetic
2. industry_final_demand_for_other_non_specified_hydrogen_non_energetic
3. industry_final_demand_for_other_hydrogen_non_energetic

Edge from 2 to 3 has parent share 0.0; other (existing) edge to 3 has parent share 1.0.

Added input statements for non-energetic use of:

1. coal
2. biomass
3. network gas
4. crude oil
5. hydrogen
6. cokes

Also added input statement for energetic use of cokes.

Cokes are now omitted from the initialization value of the 'coal' slider (which now only includes coal and coal gas).

@mabijkerk  Can you let @thomas-qah know  when to perform the migration?

Solves https://github.com/quintel/etmodel/issues/4116


mabijkerk (approved):
🙌

PR #2965, opened 2023-11-28
Adding Efate without curves

Charlottevm:
This is the first concept, still to be reviewed by the client.
Curves will be added at a later stage.


redekok (approved):
LGTM!

PR #2969, opened 2023-12-05
Resolving faulty household demand and adding curves

Charlottevm:
This PR resolves the faulty numbers in household demand and adds all available curves to Efate.


redekok:
Opening the hourly hydrogen demand curve results in the following error:

image

Also, the "modify curves" page still returns in an error which makes it harder to review all the curves unfortunately..

image

redekok:
@Charlottevm the specs are still failing for this PR. It seems as if the values do not sum up to 1/3600 for a bunch of curves:

image

Issue #2967, opened 2023-12-01
Number of vehicles does not change with modal split


mabijkerk:
**Background**
For a cars, buses, vans and trucks we calculate the number of units. This is mainly done to determine the available storage volume for the electric variants. For example, for cars it works like this, the `transport_useful_demand_cars` node gets the number_of_units assigned in the start year:
```ruby
~ number_of_units = AREA(number_of_cars)
```

The `transport_car_using_electricity` the determines its number of units using the inheritable number of units group, which simply takes the share of electric cars and multiplies it with the number of units for all cars.
```ruby
- groups = [inheritable_nou, ...]
```

If the **share of electric cars in the mix of cars** is changed, automatically the number of electric cars will change as well, through the interitable_nou group. You can query this:
```ruby
EACH(
  V(transport_car_using_electricity,number_of_units),
  V(transport_useful_demand_cars,number_of_units),
  INPUT_VALUE(transport_car_using_electricity_share)
)


[
  4,265,500.0,
  8,531,000.0,
  50.0,
]
```

**Problem**
The problem is that if the **modal split of cars in the total passenger transport** is changed, the total number of cars does not change. This means that if I change `transport_cars_share` or if I change the total passenger demand `transport_useful_demand_passenger_kms`, the number of cars will stay at 8,531,000.0.

**Goal**
For cars and buses their number of units need to scale with the change in passenger transport demand **and** with their share in the modal split for passenger transport. For vans and trucks their number of units need to scale with the change in freight transport demand **and** with their share in the modal split for freight transport. This means that we need to adjust the following inputs to also update the relevant number of units:

- tranport_useful_demand_passenger_kmgs
- transport_cars_share
- transport_busses_share
- transport_useful_demand_freight_tonne_kms
- transport_trucks_share
- transport_vans_share

kndehaan:
Some effort and discussion has been put into solving this issue (see https://github.com/quintel/etsource/pull/2970), however more effort is required than initially expected. The solutions consist of the following steps:

1. Update inputs as specified by Mathijs so that `number_of_units` of cars, busses, trucks and vans are updated, minding to check / set priority correctly
2. The area attribute `number_of_xxx` should be updated with the newly calculated number_of_units

Part 1 is relatively easy, however implementing part 2 in the useful demand inputs specifically is more complicated due to its `update_type = %y`. This results in strange numbers of updating the area attribute. 

[...]

PR #2970, opened 2023-12-05
Adding change in number of vehicles with change of demand in the model

kaskranenburgQ:
Fixing the issues as explained in issue [#2967 ](https://github.com/quintel/etsource/issues/2967).

### DIFF
--- a/inputs/demand/transport/transport_freight/transport_truck/transport_trucks_share.ad
+++ b/inputs/demand/transport/transport_freight/transport_truck/transport_trucks_share.ad
@@ -2,29 +2,25 @@
 # The 2nd update is done by taking the amount of trucks in the current data, divided by the corresponding share of trucks in the demand for freight transport. This is than multiplied by the user input. 
 
 - query =
-      EACH(
-        UPDATE(
-         EDGE(transport_useful_demand_trucks, transport_useful_demand_freight_tonne_kms),
-          share,
-          DIVIDE(USER_INPUT(), 100))
-      ,
-        UPDATE(
-          V(transport_useful_demand_trucks),
-          number_of_units
-          ,
-            DIVIDE(
-              PRODUCT(
-                DIVIDE(USER_INPUT(),100),
-                AREA(number_of_trucks)
-              ),
-              QUERY_PRESENT(-> {V(transport_useful_demand_trucks,share_of_transport_useful_demand_freight_tonne_kms) } 
-              )
-            )
+    EACH(
+      UPDATE(
+        EDGE(transport_useful_demand_trucks, transport_useful_demand_freight_tonne_kms),
+        share,
+        DIVIDE(USER_INPUT(), 100)
+      ),
+      UPDATE(
+        V(transport_useful_demand_trucks),
+        number_of_units,
+        DIVIDE(
+          PRODUCT(
+            DIVIDE(USER_INPUT(),100),
+            AREA(number_of_trucks)
+          ),
+          QUERY_PRESENT(-> {V(transport_useful_demand_trucks,share_of_transport_useful_demand_freight_tonne_kms) } 
           )
         )
-      
-
-
+      )
+    )
 
 - share_group = freight_transport
 - priority = 2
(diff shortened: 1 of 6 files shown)


mabijkerk:
Seems to work great, nice work @kaskranenburgQ! I have slightly editing the formatting in the inputs, for consistency. Additionally, I have two questions:

- The AREA attribute is not updated with these inputs, so if I make changes to the modal split of e.g., cars, the number_of_units on the node is updated accordingly, but if I query AREA(number_of_cars), I get the same value. Is this something we want to update as well? We do not seem to use in currently in any queries, but we may in the future. What do you think?
- If I increase the efficiency of e.g., electric vehicles, number_of_units does not change either. This means that, if vehicles get more efficient, we don't use fewer vehicles but instead the vehicles that we use need less energy for each tkm or pkm. Implicitly this means that we assume that the average tkm or pkm delivered per vehicle per year is not affected by efficiency changes. I'm okay with this implicit assumption. What do you think?

kaskranenburgQ:
@mabijkerk Thanks for the changes!
Regarding your questions:
1. I will follow your call on this matter, at the moment I don't have enough experience with the AREA attributes to form an opinion about this. I'm not quite sure how and when these attributes are used and I therefore find it difficult to get to a conclusion. Would like to discuss this some time. 
2. I agree with you. I think the efficiency should affect the tkm.

kndehaan:
I've tried to implement updating the AREA attribute `number_of_xxx` when changing the share of certain transport technologies in passenger demand and freight demand, as well as updating this attribute when changing expected yearly change in passenger km or freight kms. The latter is not that straightforward, since  `transport_useful_demand_passenger_kms` and `transport_useful_demand_freight_tonne_kms` have update_type = %y. Therefore, the following gives very strange results in the updated AREA attribute:

```
EACH(
      UPDATE(V(transport_useful_demand_passenger_kms), preset_demand, USER_INPUT()),
      UPDATE(V(transport_useful_demand_cars), number_of_units, USER_INPUT()),
      UPDATE(V(transport_useful_demand_busses), number_of_units, USER_INPUT()),
      UPDATE(
        AREA(),
        number_of_cars, 
        V(transport_useful_demand_cars, number_of_units)
      ),
      UPDATE(
        AREA(),
        number_of_busses, 
        V(transport_useful_demand_busses, number_of_units)
      )
    )
```
AREA(number_of_cars) --> 1.4e242
AREA(number_of_busses) --> 8.7e147

Some questions / solutions:

- Is there a good/quick method to correctly update the AREA attributes in or outside this query, while maintaining the update_type = %y (so no required changes to the first 3 update statements)
- A possibility is to not use the update_type = %y, but to manually apply the equation behind this to the first 3 UPDATE statements. However, it might not be desirable to write this out since we have created the method update_type = %y ETEngine for a reason.
- Other possibility is to change the input type of change in passenger km or freight km to not indicate the expected change per year, but the overall change between start and future year... And then change this query accordingly. 

Already discussed with @kaskranenburgQ. @mabijkerk what do you think is the best way?

kndehaan (changes requested):
Incorrect interpretation from my side of update_type = %y, and therefore unjustifiably removed. This should be corrected, but this awaits a response of Mathijs on the method of writing this query.

mabijkerk:
Some thoughts:

- I think the issue here may be that both the "transport demand change" input and the "modal split" inputs update the `number_of_units`. If these updates are applied sequentially, then you will get a rather large number of units indeed, as in shown in the results of @kndehaan .
- The difficulty is that in any scenario, only one of the two input types may be set. We therefore cannot just apply the number of units change in one of the two input types.
- We will need to find a way to only update the number of units once, if one or both inputs are set.
- The most straightforward way, which I also applied in the household section, is to calculate the value in both inputs and set it 'hardcoded' instead of applying a change to the value. This works, but is quite cumbersome, as we're calculating the same value in two inputs. I feel there may be better ways to go about this, but as of yet I have not come up with one. Any ideas are welcome.
- Finally, I'm not overly fond of the `update_type` functionality. I would rather have dedicated GQL for this, for example UPDATE_PER_YEAR.

kndehaan:
I will rewrite to issue since the implementation of the fix is larger than initially expected.

kndehaan:
More effort is required to solve this issue, therefore this PR will be closed and proposed next steps are described in this issue https://github.com/quintel/etsource/issues/2967

PR #2980, opened 2023-12-13
Add LOHC and LH2 to carriers.csv for the Efate dataset

redekok:
In addition to https://github.com/quintel/etsource/pull/2978, extend the `carriers.csv` file with (imported) liquid hydrogen and LOHC for the Efate dataset


Charlottevm (approved):
LGTM 👍

PR #2981, opened 2023-12-15
Introduce more fine-grained load-shifting

thomas-qah:
## What?
This PR adds support for setting up more fine-grained load-shifting by our users.

## Why?
The currently existing load-shifting functionality caused too much of a peak in the demand for the electricity system. The user needed a way to spread out this demand more evenly over time.

## How?
By adding new inputs to etsource through which the 'dampening' values can be set.


mabijkerk (approved):
- Added failsafe to load shifting inputs: number of units is set to 0 if either capacity is set to 0
- Rewrote input capacity inputs so that the user does not set input_capacity_from_share directly but effectively sets the input capacity
- Added input_capacity_from_share attribute to nodes

mabijkerk:
Note @thomas-qah: Gemfile should be updated when Atlas is mergeed.

PR #2982, opened 2023-12-22
Input and chart curve for residential forecasting

noracato:
Changes the old boolean toggle for storage optimization for households for setting the `merit_order.subtype` to `optimizing_storage`, to a radio button input that can set the subtype to:
- `default`: uses the nodes default subtype (usually `storage`)
- `optimizing_storage`: enables the battery to join in the forecasting algorithm for the full systems residual load curve
- `optimizing_storage_households`: enables the battery to join in the forecasting algorithm for the households sector residual load curve

And adds a query to show the households residual load curve in charts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ gem 'rake'
 
 group :development, :test do
   gem 'roo'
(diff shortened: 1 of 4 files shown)


mabijkerk (approved):
⚡

Issue #2979, opened 2023-12-13
Review costs of carriers import/export queries


mabijkerk:
**Issue**
In the mece_costs queries for 4_energ_carriers, most queries have a structure similar to coal and coal products:

```ruby
- query =
    SUM(
      PRODUCT(V(CARRIER(coal), cost_per_mj), V(energy_extraction_coal, demand)),
      PRODUCT(V(CARRIER(coal), cost_per_mj), V(energy_import_coal, demand)),
      PRODUCT(V(CARRIER(lignite), cost_per_mj), V(energy_extraction_lignite, demand)),
      PRODUCT(V(CARRIER(lignite), cost_per_mj), V(energy_import_lignite, demand)),
      PRODUCT(V(CARRIER(cokes), cost_per_mj), V(energy_import_cokes, demand))
    )
    -
    SUM(
      PRODUCT(V(CARRIER(coal), cost_per_mj), V(energy_export_coal_backup, demand)),
      PRODUCT(V(CARRIER(lignite), cost_per_mj), V(energy_export_lignite_backup, demand)),
      PRODUCT(V(CARRIER(cokes), cost_per_mj), V(energy_export_cokes_backup, demand))
    )
    -
    SUM(
      PRODUCT(V(CARRIER(coal), cost_per_mj), V(energy_export_coal_baseload, demand)),
      PRODUCT(V(CARRIER(lignite), cost_per_mj), V(energy_export_lignite_baseload, demand)),
      PRODUCT(V(CARRIER(cokes), cost_per_mj), V(energy_export_cokes_baseload, demand))
    )
```

However, for biomass products, instead of the demand, the primary demand is queried:
```ruby
- query =
    SUM(
      PRODUCT(V(CARRIER(torrefied_biomass_pellets), cost_per_mj), V(energy_torrefaction_wood, primary_demand_of_torrefied_biomass_pellets)),
      PRODUCT(V(CARRIER(wood_pellets), cost_per_mj), V(energy_production_wood_pellets, primary_demand_of_wood_pellets)),
      PRODUCT(V(CARRIER(bio_kerosene), cost_per_mj), V(energy_production_bio_kerosene, primary_demand_of_bio_kerosene)),
      PRODUCT(V(CARRIER(bio_ethanol), cost_per_mj), V(energy_production_bio_ethanol, primary_demand_of_bio_ethanol)),
[...]

PR #2984, opened 2024-01-04
Close issue #2979. Changing cost calculation queries of import and export.

kaskranenburgQ:
I researched the way the oil and biomass query was set up.

- For biomass the cost calculation method was not coherent with the other cost calculations. 
The cost calculation standard in the other queries is : Import - export (baseload) - export (backup). 
Since there is only import of dry biomass in the model, it would not be coherent to calculate all the production costs of all carriers as it was done before. I changed this.

- For oil see the screenshot from a blank scenario underneath:

Screenshot 2024-01-04 at 09 39 41

Screenshot 2024-01-04 at 09 40 30

When calculating the costs with the final_demand_of_crude_oil, not the whole energy export would be taken into account. For this reason I changed the oil query to the 'demand' variant.

### DIFF
--- a/gqueries/general/costs/mece_costs/4_energy_carriers/costs_carriers_biomass.gql
+++ b/gqueries/general/costs/mece_costs/4_energy_carriers/costs_carriers_biomass.gql
@@ -1,16 +1,7 @@
+# Carrier costs based on net flows (imports - baseload export)
+
 - query =
-    SUM(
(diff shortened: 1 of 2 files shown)


kndehaan:
Is this PR still up-to-date and can this be merged and closed in the near future? Or is the better alternative here to close the PR and rewrite to an issue?

PR #2988, opened 2024-01-05
Fix imported liquid hydrogen query

mabijkerk:
Minor error in primary demand query caused the [primary_demand](https://github.com/quintel/mechanical_turk/blob/master/spec/demand/primary_demand_spec.rb) MT spec to fail. This query fixes that.


redekok (approved):
LGTM!

PR #3006, opened 2024-02-05
Fix interconnector query

### LINKED ISSUE #2991: Future costs of electricity query should be updated
**Background**
The [costs_carriers](https://github.com/quintel/etsource/blob/master/gqueries/general/costs/mece_costs/4_energy_carriers/costs_carriers.gql) contains the costs for electricity, which has present and future costs. The query for future costs, [costs_carriers_electricity_future](https://github.com/quintel/etsource/blob/master/gqueries/general/costs/mece_costs/4_energy_carriers/costs_carriers_electricity_future.gql), has two issues. It looks like this;

```ruby
    future:SUM(
      PRODUCT_CURVES(
        V(energy_interconnector_1_imported_electricity, electricity_output_curve),
        Q(interconnector_1_marginal_cost_curve)
      ),
      PRODUCT_CURVES(
        V(energy_interconnector_2_imported_electricity, electricity_output_curve),
        Q(interconnector_2_marginal_cost_curve)
      ),
      PRODUCT_CURVES(
        V(energy_interconnector_3_imported_electricity, electricity_output_curve),
        Q(interconnector_3_marginal_cost_curve)
      ),
      PRODUCT_CURVES(
        V(energy_interconnector_4_imported_electricity, electricity_output_curve),
        Q(interconnector_4_marginal_cost_curve)
      ),
      PRODUCT_CURVES(
        V(energy_interconnector_5_imported_electricity, electricity_output_curve),
        Q(interconnector_5_marginal_cost_curve)
      ),
      PRODUCT_CURVES(
        V(energy_interconnector_6_imported_electricity, electricity_output_curve),
        Q(interconnector_6_marginal_cost_curve)
      )
    )-
    PRODUCT(
      V(energy_export_electricity, weighted_carrier_cost_per_mj),
      V(energy_export_electricity, demand)
    )
```

**Issue**

1. The query is missing imported costs for the interconnectors 7 to 12
2. The query uses `weighted_carrier_cost_per_mj` and `demand` to calculate the exported costs, which seem to be yearly attributes. The question is why, if we use hourly values to calculate import costs, we do not seem to use hourly values to calculate export gains.

**Solution**

- [x] Add remaining interconnectors to imported costs
- [ ] Research whether yearly or hourly values are taken for the exported costs
- [ ] If yearly values are taken, discuss wether hourly values should be used instead
- [ ] If hourly values should be used, update the query accordingly

@KoenvanB I will pick up the first to do. Could you pick up the remaining to do's? We can discuss them together.

KoenvanB:
This PR adds interconnectors 7-12 to the `costs_carriers_electricity_future` query. 

**Discussion**
Should the ETM use the domestic electricity price for exporting electricity, rather than the weighted carrier costs?

According to the [documentation](https://docs.energytransitionmodel.com/main/cost-methods) the ETM only takes into account the primary carrier costs when exporting electricity:

> For export the ETM charges the costs of the primary carrier that is needed for that export. The export of electricity and transit of oil is cost neutral and independent of the market price. A country does not "earn" money from processing oil. And for electricity the ETM deducts the costs of the primary carriers needed to produce that electricity. 

Consequently, however, the import costs are calculated on an hourly basis whereas the export costs are aggregated on an annual basis. 

The difference between using the hourly market price (marginal costs curve) and the weighted carrier costs is significant; see [this example](https://github.com/quintel/etsource/issues/2991#issuecomment-1911641626).

Closes #2991 .

### DIFF
--- a/gqueries/general/costs/mece_costs/4_energy_carriers/costs_carriers_electricity_future.gql
+++ b/gqueries/general/costs/mece_costs/4_energy_carriers/costs_carriers_electricity_future.gql
@@ -23,11 +23,34 @@
       PRODUCT_CURVES(
         V(energy_interconnector_6_imported_electricity, electricity_output_curve),
         Q(interconnector_6_marginal_cost_curve)
+      ),
+      PRODUCT_CURVES(
+        V(energy_interconnector_7_imported_electricity, electricity_output_curve),
+        Q(interconnector_7_marginal_cost_curve)
+      ),
+      PRODUCT_CURVES(
+        V(energy_interconnector_8_imported_electricity, electricity_output_curve),
+        Q(interconnector_8_marginal_cost_curve)
+      ),
+      PRODUCT_CURVES(
+        V(energy_interconnector_9_imported_electricity, electricity_output_curve),
+        Q(interconnector_9_marginal_cost_curve)
+      ),
+      PRODUCT_CURVES(
+        V(energy_interconnector_10_imported_electricity, electricity_output_curve),
+        Q(interconnector_10_marginal_cost_curve)
+      ),
+      PRODUCT_CURVES(
+        V(energy_interconnector_11_imported_electricity, electricity_output_curve),
+        Q(interconnector_11_marginal_cost_curve)
+      ),
+      PRODUCT_CURVES(
+        V(energy_interconnector_12_imported_electricity, electricity_output_curve),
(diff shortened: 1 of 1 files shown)


mabijkerk:
@KoenvanB, thanks for reminding me of this PR. We should definitely merge this commit and I will do so. However, this does **not** close #2991 since it does not conclude the debate on whether or not to use the hourly or yearly electricity price for export.

PR #3013, opened 2024-02-29
Propane addition 

kaskranenburgQ:
In this PR the following changes have been implemented :
- Propane is added to the model as a carrier, an import node and a distribution node  and a link for the national gas network is added for this carrier.
- For the final demand queries: propane is added within the natural gas and derivatives group.
- For the primary demand queries: propane is added within the natural gas and derivatives group. 
- Primary co2 emission queries: Propane is added within the natural gas and derivatives group.
- For the graphs that do not calculate their queries based on final or primary demand (such as import/export graphs): Propane is added under “natural gas”.
- For the multi year charts, propane is added under natural gas and derivatives.
- For the co2-sheets: No changes were needed since the queries are based on network gas.
- Furthermore at most places “share_of_natural_gas_in_gas_network” has been replaced by “share_of_fossil_gas_in_gas_network” where fit. At all places “share_of_greengas_in_natural_gas_network” has been replaced by “share_of_sustainable_gas_in_gas_network”.
- For the final demand queries: propane is added within the natural gas and derivatives group.
- A flat-profile output network gas output curve is added for the network gas production chart, based on the energy_distribution_propane node


mabijkerk (changes requested):
In the sankey_csv, the following lines set the import of natural gas:

```
  - Group: Primary demand
    Carrier: Natural gas
    Category: Import
    Type: Output
    Value: import_of_natural_gas
```

The query `import_of_natural_gas` does not include propane and is now also missing LNG. It does include LNG on `master`. Moreover, when downloading the sankey_csv or Energy flows (overview) in the data export section, the download fails.

- [x] Fix Energy flows (overview) download

PR #3016, opened 2024-03-14
Fix typo in natural gas chart queries

KoenvanB:
This PR fixes 2 typo's for the final energy demand mekko and the present energy exports chart.


mabijkerk (approved):
🙌

PR #3019, opened 2024-03-19
BECCUS

kaskranenburgQ:
With this update, the option to model BECC(U)S plants with the ETM becomes possible.

For the ETSource repository this means the following changes have been implemented:

- Addition of power-production only biomass plants to the energy graph. 
- Addition of dispatchable biomass power plants to the energy graph.
- Addition of CCS biomass power plants to the energy graph.
More specifically the following nodes have been added:
- energy_power_wood_pellets_must_run
- energy_power_wood_pellets_ccs_must_run
- energy_power_wood_pellets_dispatchable
- energy_power_wood_pellets_ccs_dispatchable
- energy_chp_local_ht_wood_pellets_must_run
- energy_chp_local_ht_wood_pellets_ccs_must_run
- energy_chp_local_ht_wood_pellets_dispatchable
- energy_chp_local_ht_wood_pellets_ccs_dispatchable
- energy_chp_local_mt_wood_pellets_must_run
- energy_chp_local_mt_wood_pellets_ccs_must_run
- energy_chp_local_mt_wood_pellets_dispatchable
- energy_chp_local_mt_wood_pellets_ccs_dispatchable

- Addition of new specifications as specified in NSA's for biomass plants to the respective ETsource files.
- Implementation of these new nodes in all graphs related to biomass production.
- Addition of new area attribute "captured biogenic co2 price" to the Datasets 
- Creation of queries for new graph called : "Net effect of demand and supply of CO2 on emissions".


gqueries/output_elements/output_series/waterfall_captured_emssions/negative_captured_bio_emissions_waterfall_captured_emissions.gql
```diff
@@ -0,0 +1,3 @@
+- query = 
+        NEG(Q(emissions_captured_biogenic_total))
+- unit = tonne
```
mabijkerk, line 3: This query uses hardcoded references to power plants. This means that torrefied biomass pellets (and potentially other biogenic) flows are missing. I would rewrite the query to be more generic, somethin like:
`SUM(V(G(electricity_production),"demand * weighted_carrier_potential_co2_per_mj * free_co2_factor"))`

kaskranenburgQ: I just tested this query, seems to be accurate for what we are searching.

PR #3021, opened 2024-03-22
Hydrogen improvements

kndehaan:
Modelling of hydrogen supply, demand and storage has been improved. Summarised:

- A break down in must-run and dispatchable plants of the hydrogen supply technologies ATR, SMR and ammonia reforming
- Possibility to change the merit order of dispatchable hydrogen production and demand
- Modelling of hydrogen storage is improved by adding salt cavern and depleted gas field technologies and allowing the user to set installed volume and relative capacity.
- Changes in hydrogen transport assumptions and modelling
- New charts: "Installed storage volume per carrier" and "Flexible hydrogen supply and demand per hour"

Changes applied in this repo specifically are (summarised):
- Added hydrogen storage salt cavern and depleted gas field nodes and queries
- Break down of must-run and dispatchable hydrogen production nodes and queries
- Added merit order attributes to nodes
- Changes and additions of queries for charts and tables

Goes with:
- https://github.com/quintel/etsource/pull/3021
- https://github.com/quintel/etmodel/pull/4223
- https://github.com/quintel/documentation/pull/175
- https://github.com/quintel/etdataset/pull/994
- https://github.com/quintel/atlas/pull/166
- https://github.com/quintel/etengine/pull/1421


mabijkerk (changes requested):
The `hydrogen_spec` on `mechanical_turk` fails because it uses the query `turk_hydrogen_mekko_supply`, which uses a number of outdated queries:

```
- query =
    SUM(
        Q(energy_import_hydrogen_for_mekko),
        Q(energy_hydrogen_steam_methane_reformer_for_mekko),
        Q(energy_hydrogen_steam_methane_reformer_ccs_for_mekko),
        Q(energy_hydrogen_flexibility_p2g_electricity_for_mekko),
        Q(energy_hydrogen_electrolysis_solar_electricity_for_mekko),
        Q(energy_hydrogen_electrolysis_wind_electricity_for_mekko),
        Q(energy_hydrogen_biomass_gasification_ccs_for_mekko),
        Q(energy_hydrogen_biomass_gasification_for_mekko)
    )
```

@kaskranenburgQ or @kndehaan could you please make sure the query is updated properly.

mabijkerk (changes requested):
The chart `energy_use_of_ccus` is outdated as well, since it uses the outdated query `energy_use_of_hydrogen_production_co2_capture`:

```
- query =
    SUM(
      Q(energy_use_of_energy_hydrogen_steam_methane_reformer_ccs_co2_capture),
      Q(energy_use_of_energy_hydrogen_biomass_gasification_ccs_co2_capture)
    )
```

This for example does not include ATR. Moreover, it also does not include the updated node names for SMR in the `energy_use_of_energy_hydrogen_steam_methane_reformer_ccs_co2_capture` query. I wonder whether it would be possible to generate this chart in a more robust way and/or write a `mechanical_turk` spec to verify whether all CCS technologies are included? @kaskranenburgQ or @kndehaan could one of you pick this up?

kndehaan:
The query `turk_hydrogen_mekko_supply` is updated with the correct query names: https://github.com/quintel/etsource/pull/3021/commits/8f7fa4131238e23e16aef2d0abaa12e61a359f4a

kndehaan:
query `energy_use_of_hydrogen_production_co2_capture` for Energy use for CCUS chart is updated in two commits:

- https://github.com/quintel/etsource/pull/3021/commits/0786df41503d77b0c3be3cc47b8c2373324239d0: non-robust method
- https://github.com/quintel/etsource/pull/3021/commits/42fdc131e281ef1c24f31a042a4125340692e3bd: more robust method using filters and groups. Review requested of @mabijkerk , if not approved otherwise this commit should be removed.

graphs/energy/nodes/buildings/buildings_space_heater_combined_hydrogen.converter.ad
```diff
@@ -20,8 +20,9 @@
 - fever.group = buildings_space_heating
 - fever.type = producer
 - fever.technology_curve_type = tech_day_night
-- hydrogen.profile = self: hydrogen_input_curve
+- hydrogen.group = flat
 - hydrogen.type = consumer
+- hydrogen.subtype = generic
```
mabijkerk, line 25: This does not work. Need to rewrite this to merit-like. Will discuss this with @noracato and @kaskranenburgQ on Monday.

graphs/energy/nodes/energy/energy_hydrogen_steam_methane_reformer_ccs_must_run.ad
```diff
@@ -17,16 +17,16 @@
...
 - network_gas.profile = flat
 - network_gas.type = consumer
-- forecasting_error = 0.0
 - free_co2_factor = 0.669
 - ccs_capture_rate = 0.669
-- land_use_per_unit = 0.03235581614498978
+- hydrogen_output_capacity = 300.0
 - takes_part_in_ets = 1.0
 
 ~ demand = 0.0
```
mabijkerk, line 32: @kndehaan shouldn't this hydrogen_output_capacity of SMR with CCS be lower due to the lower output efficiency compared to the SMR without CCS? Is this something you can look into?

kndehaan: Regarding your new question about `hydrogen_output_capacity`, I see in ETEngine that there's a function to define  calculate this attribute if it is not already defined in the hydrogen node:
```
# Public: Returns the hydrogen output capacity of the node.
#
# If no value is assigned in the ETSource data, a capacity will attempt to be derived from the
# `typical_input_capacity`. Note that we do not call `input_capacity` and instead rely on the
# typical capacity, as `input_capacity` may itself call `hydrogen_output_capacity`.

      def hydrogen_output_capacity
        fetch(:hydrogen_output_capacity, false) do
          typical_input_capacity ? typical_input_capacity * hydrogen_output_conversion : 0.0
        end
      end
```

I'm not sure why we've defined `hydrogen_output_capacity` in some of the nodes, if this is really required for some reason or if it was first our intention to define this for all nodes and that later on we've decided to use a function to set this attribute. In the end, all hydrogen production nodes will obtain the `hydrogen_output_capacity` attribute. 

Do you think it's best to leave it as it is or to remove the defined `hydrogen_output_capacity` in the specific nodes and only use the defined function for all hydrogen production nodes?

mabijkerk: I think hydrogen_output_capacity can be set explicitly in case where it should not be determined by the output efficiency. For example for hydrogen storage we want to define input and output capacity independently.

I expect that we can remove it from the SMR and that would be my preference. In the future I would prefer to have all of our CCS technologies to receive a lower output capacity than their non-CCS counterpart instead of having a higher input capacity.

kndehaan: I've removed the defined `hydrogen_output_capacity` attribute in the SMR, ATR and ammonia reforming nodes. https://github.com/quintel/etsource/pull/3021/commits/2bd22c7a91090f058ce1cec5efb595c9bf953c27

PR #3025, opened 2024-03-28
Update and add queries for external coupling of emissions

mabijkerk:
This PR adds a number of queries to couple external models that use CO2 flows:

- For electricity production, for electricity production from waste and for hydrogen production the following queries are available
  - Biogenic emitted
  - Biogenic captured
  - Biogenic total
  - Fossil emitted
  - Fossil captured
  - Fossil total
  - Total captured


gqueries/modules/external_coupling/captured_fossil_co2_emissions_of_produced_hydrogen.gql
kaskranenburgQ, line 1: This query does not add up with what I see in the engine. In a blank scenario with both 5000 ATR and SMR capacity and 100% CCS. I see the following in the molecule graph: 
Screenshot 2024-03-29 at 13 56 51
The query returns the following: 
Screenshot 2024-03-29 at 13 57 35
I think this difference lies in the method of the total fossil emissions "primary_co2_emissions_without_capture"

The demand of these nodes together = 74.9 + 72.27 (SMR + ATR) = 146,56
The primary_demand of these nodes together = 75.96 + 73.29 (SMR + ATR) = 149,25

This difference is procentually 0,982

mabijkerk: Very well spotted! This is probably the method `primary_co2_emission_of_fossil` uses primary demand and therefore increase the total with the loss that follows 'after' the SMR. This is the loss in the `energy_treatment_natural_gas` node, which is exactly 0.986016667985657.

I will update the queries.

PR #3030, opened 2024-04-08
Hybrid oil heat pump in ETM

KoenvanB:
This PR adds hybrid oil heat pumps to the ETM for households and buildings in the form of three new nodes:
- `households_space_heater_hybrid_crude_oil_heatpump_air_water_electricity`
- `households_water_heater_hybrid_crude_oil_heatpump_air_water_electricity`
- `buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity`
Corresponding inputs are included under the 'space heating (& hot water) section'.

Further changes include:
- Restore UKNI air temperature curve to unnormalized form (corrects a mistake)
- Queries to include oil hhps to the `number_of_houses_per_space_heating_technology` chart
- Queries to set the heatpump in the heat merit order.
- Queries to set the cop of the hybrid heatpump
- Queries to extract the number_of_units, cop & useable_heat


mabijkerk (changes requested):
The query `electricity_used_in_heatpumps` has a number of (hardcoded) heat pumps. The hybrid crude oil and hybrid hydrogen heat pumps are missing from this query. Can we update this query to be more robust or if it's not used anywhere, can we delete the query (my preferred solution)?

graphs/energy/edges/households/households_space_heater_hybrid_crude_oil_heatpump_air_water_electricity_aggregator-households_useful_demand_for_space_heating_apartments_1945_1964@useable_heat.ad
```diff
@@ -0,0 +1,6 @@
+- type = share
+- reversed = false
+
+
+
+
```
mabijkerk, line 6: @KoenvanB / @kaskranenburgQ why do we have all these trailing returns in the edge files?

kaskranenburgQ: I think this is because the files have been copy-pasted from the other technologies. These have trailing whitespaces in them.

kaskranenburgQ:
I have removed the _electricity_used_in_heatpumps_ query since it was not used in any repositories.

PR #3035, opened 2024-04-22
Updating all municipalities to 2019

Charlottevm:
This PR updates all municipal regions to 2019.

For more information, check out the ETLocal PR [here](https://github.com/quintel/etlocal/pull/516). 

One last export needs to be done with regard to the municipalities (@KoenvanB ).


KoenvanB (changes requested):
The new data is largely in order, except for one share group in the Eemsdelta dataset.

**Eemsdelta**
The `present_share_of_<housing_type>_in_useful_demand_for_space_heating` share group is in percentages rather than a factor. The rspec does not return an error but these values deviate from all other datasets where they are a factor.

**Other remarks**
I have noticed the following remarkable changes across all datasets:
- `residences/buildings_roof_surface_area_for_pv` has increased by a factor of 2-10, or even more for buildings
- `total_land_area` has increased (e.g. Almere factor >2)
- `coast_line` has increased by a large factor  (e.g. Steenwijkerland from near 0 to about 30)
- Some municipalities now have a coastline (e.g. Zaanstad)
- `offshore_suitable_for_wind` has increased by >factor of 2 (e.g. Deventer)
- `total_costs_present` for electricity infrastructure have changed by a few percent (e.g. Rotterdam)
- `co2_emissions_1990` have changed by a few percent (e.g. Rotterdam)

I cannot judge whether these changes make sense based on the (new) data sources, but I mention them for completeness sake.

Charlottevm:
Thanks for your thorough look into the datasets @KoenvanB .
For now the issue regarding the `present_share_of_<housing_type>_in_useful_demand_for_space_heating` is resolved. 
Furthermore for the `coast-line` attribute all 'binnenwater' data has been removed which means only coastal municipalities  can now use coastal wind turbines. The `total_land_area` attribute first included both land area and water, water has now been removed. 

With regards to all other changes, these are to be expected during an update. If anything is amiss for these figures this should be looked into during the 2022 update.

KoenvanB (approved):

PR #3045, opened 2024-05-06
Update energy mix visual

kndehaan:
The energy mix visual was outdated and is hereby updated. Updating the visual encompasses the following changes:

- Updating underlying queries of 'Opgave per thema 2050' components
- Updating underlying queries of 'Energiebronnen 2050'
- Locales improvements
- Incorporate ETM logo

Goes with:
- https://github.com/quintel/etmodel/pull/4261
- https://github.com/quintel/etsource/pull/3045


mabijkerk (approved):
Good improvements. All the queries in the `factsheet_total_supply` query should get a critical review and be rewritten to more robust queries. For example, ammonia, LOHC and LH2 are missing, some queries contain hardcoded node names, the total supply does not sum to primary demand caused by final etc.

However, this is not worth the effort at this point.

Issue #2999, opened 2024-01-19
Sankey should show biomass and not 'aardgas' when all network gas is green gas


MartLubben:
<img width="1103" alt="image" src="https://github.com/quintel/etsource/assets/33415156/f6d225a3-5b30-4b71-bb3b-56a5fb7861d4">

In a scenario with only green gas The Sankey shows 'aardgas'.

kaskranenburgQ:
I think it will be difficult to change the label to a conditional format. 
Would it be better to change the label to "netwerkgas"? 
@mabijkerk What do you think?

kaskranenburgQ:
I have delved at this issue a bit further:
We could add biogas to biomass products.
We would then need to remove network_gas from the network_gas flow.
 I will investigate if this is a possible and wishful solution.

kaskranenburgQ:
Thanks for the clarification @mabijkerk ! 
As far as I can see the query biomass_products_to_agriculture_in_sankey is not referred to in the query for biomass to agriculture, final_demand_of_biomass_products_in_agriculture_energetic is used instead. 


In this query greengas is not used: 
```
# Final demand of the 'biomass_products' carrier group

- unit = PJ
- query =
    SUM(
      V(
        FILTER(
          FILTER(
            FILTER(
              EG(final_demand),"sector == :agriculture"
            ),
            energetic?
          ),
          biodiesel?
        ),
        value
      ),
      V(
        FILTER(
          FILTER(
            FILTER(
              EG(final_demand),"sector == :agriculture"
            ),
            energetic?
          ),
          bio_ethanol?
        ),
        value
      ),
      V(
        FILTER(
          FILTER(
            FILTER(
              EG(final_demand),"sector == :agriculture"
            ),
            energetic?
          ),
          wood_pellets?
        ),
        value
      ),
      V(
        FILTER(
          FILTER(
            FILTER(
              EG(final_demand),"sector == :agriculture"
            ),
            energetic?
          ),
          bio_lng?
        ),
        value
      )
    ) / BILLIONS
```
This would mean that greengas is not counted double in the current state of the sankey but fully attributed to natural gas instead.  
I will use the biomass_products_to_x_in_sankey queries instead and use your approach to rewrite the natural gas queries.

kaskranenburgQ:
This issue has been fixed with the addition of network gas as an intermediate carrier in the sankey chart:

![Image](https://github.com/user-attachments/assets/ae3f2a27-8d85-4332-8466-28557e4852af)

PR #3051, opened 2024-05-14
Fixing gas flows in Sankey energy overview

kaskranenburgQ:
This PR aims to fix the gas flows in the energy overview as described in [Issue #2999](https://github.com/quintel/etsource/issues/2999)
It does this by altering the final_demand_of_natural_gas_in__sector_ queries and the final_demand_of_biomass_products_in__sector_ queries. 

The altered queries are also used in the graph final_demand_mekko_energetic. 
In this graph the old queries did make some sense since they were placed under the label 'network_gas' and not 'natural_gas'.
This means that we either need to create different queries for the sankey overview or that we need to change the label 'network_gas' to 'natural_gas' in the final_demand_mekko_energetic. 
What do you think @mabijkerk ?

### DIFF
--- a/gqueries/general/final_demand/mece_energetic/final_demand_of_biomass_products_in_agriculture_energetic.gql
+++ b/gqueries/general/final_demand/mece_energetic/final_demand_of_biomass_products_in_agriculture_energetic.gql
@@ -50,5 +50,21 @@
           bio_lng?
         ),
         value
+      ),
+      PRODUCT(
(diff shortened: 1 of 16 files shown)


kaskranenburgQ:
In discussion with @mabijkerk we have decided to close this PR.
Since in the sankey download these flows are shown correctly, we can use those queries to visualize the biogas flows.

PR #3057, opened 2024-05-16
Sankey download improvements for natural gas carrier

kndehaan:
With this PR some improvements are made in accounting for natural gas flows in the sankey download, as well as the sankey chart. Improvements include:
- Accounting for share of network gas in compressed network gas (share does not equal 1.0 which was first assumed)
- Accounting for share of network gas in torrefied wood pellets
- Renaming, restructuring and fixing minor errors in queries


mabijkerk:
@kndehaan note that by changing queries like `final_demand_of_natural_gas_and_derivatives_in_industry_energetic` to include the `Q(share_of_fossil_gas_in_gas_network)`, you're effectively changing al the final demand charts. This query is used for example in the chart `final_demand_mekko_energetic_mece`.

In the final demand charts we want to keep the 'mixed carriers' / 'end user' approach as discussed. Therefore at this point we only want to fix the final demand queries that are used by the Sankey (chart and download).

kndehaan:
I've reverted the changes to the final demand queries for natural gas, so the sankey chart and final demand charts will not change on these aspects. Where necessary, I've created new final demand and conversion queries for the purpose of the sankey download only. 

@mabijkerk could you review this again? With these changes, the difference in the supply / demand balance is in the order 1e-10 (doesn't pass the 1e-12 test yet, but getting very close).

kndehaan:
Closing this PR due to long inactivity and opened this issue instead https://github.com/quintel/etsource/issues/3194

PR #3059, opened 2024-05-21
Removed duplicate entry from hydrogen levelised costs export

kndehaan:
The following keys were listed twice in this export, and therefore these duplicates are removed:
- Production curve wind electrolysis (MWh)
- Production cost curve wind electrolysis (Eur/MWh)


kaskranenburgQ (commented):
Great spot!

PR #3060, opened 2024-05-24
Co2 sankey fix

kaskranenburgQ:
This PR aims to solve issue [#4203](https://github.com/quintel/etmodel/issues/4203) from ETModel.

The main issue lay in that the hydrogen co2 streams were not queried correctly due to name changes from nodes.

### DIFF
--- a/gqueries/output_elements/output_series/sankey_190_co2/agriculture_network_gas_total_in_co2_sankey.gql
+++ b/gqueries/output_elements/output_series/sankey_190_co2/agriculture_network_gas_total_in_co2_sankey.gql
@@ -3,6 +3,17 @@
 - unit = tonne
 - query =
     PRODUCT(
-        V(agriculture_final_demand_network_gas, demand),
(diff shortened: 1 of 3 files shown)


mabijkerk:
@kaskranenburgQ I think the agriculture CHPs may not be handled properly in this chart either. Could you look into this?

- agriculture_chp_engine_network_gas_dispatchable
- agriculture_chp_engine_network_gas_must_run
- agriculture_chp_wood_pellets
- agriculture_chp_engine_biogas
- agriculture_heat_backup_burner_network_gas

kaskranenburgQ:
I updated the branch and took in the points from @mabijkerk. 

> A smaller change however would be to add a similar structure for energy_power_mv_distribution_network_electricity as we have for energy_power_hv_network_electricity in the query.

This was a great way to get rid of some of the hardcoded nodes.

I implemented the agriculture CHP's by using the nodes that have  agriculture_heat hourly calculations.

PR #3064, opened 2024-06-06
Update pv roof surface area households & buildings for RES11

KoenvanB:
This PR updates the the rooftop surface area available for solar pv for RES11 Holland Rijnland. See [this ETLocal PR](https://github.com/quintel/etlocal/pull/521).


datasets/RES11_nl2019/RES11_nl2019.derived.ad
```diff
@@ -0,0 +1,223 @@
+- area = RES11_nl2019
```
Charlottevm, line 1: This should be RES11_holland_rijnland, seems like something went wrong?

Charlottevm (changes requested):

datasets/RES11_nl2019/RES11_nl2019.derived.ad
KoenvanB: I have fixed the typo. The correct data is now shown in the ETM, see the 'land use of solar and wind' figure.

PR #3073, opened 2024-06-25
Adding non-energetic oil mix to external coupling

kaskranenburgQ:
This PR adds nodes for the non energetic oil mix to the base ctm-etm branch.

It consists of 2 parts:

- Adding nodes and edges for the non energetic oil mix.
- Adding external coupling inputs to set the non energetic oil mix.

At the moment semaphore returns an error message, this is because the node industry_final_demand_other_oil_non_energetic is specified as a 'final_demand' node. 
However, no demand is specified on this node at the moment, the semaphore check does not expect this. How could we tackle this the best @mabijkerk ?


mabijkerk:
@kaskranenburgQ, I made a few changes to the inputs:

1. Updated the folder to match the folder name for the 'regular' inputs, for clarity
2. Updated the names to match the 'regular' inputs for energetic oil mix inputs, for clarity
3. Added a share_group to the external coupling inputs for non energetic oil mix, this important because it ensures that they are checked to sum to 100%

Could you also add external coupling inputs for the energetic oil mix in industry? Don't forget to add the disabled_by attribute to the corresponding 'regular' input.

Issue #2986, opened 2024-01-05
Greengas backup export node missing


kaskranenburgQ:
In the model, energy_distribution_greengas has 2 outgoing edges to the nodes : energy_national_gas_network_natural_gas, and energy_export_greengas_baseload. It has 4 incoming edges , with 3 of those coming from nodes that can be set as shares by the user in the model. Only taking these first three edges into account, there won't be any need for a backup greengas export since the demand will set the supply.
The 4th incoming edge is from the node industry_residual_greengas. This node can only supply greengas to the distribution when the CTM is coupled with the ETM. As the name suggests, residual greengas from the industry is supplied to the node. 
If you open the node in the[ II3050 v2 - Decentral incentives scenario ](https://energytransitionmodel.com/saved_scenarios/14552), you will see the following : 
<img width="1294" alt="Screenshot 2024-01-05 at 10 14 20" src="https://github.com/quintel/etsource/assets/150331345/2e3288b6-6fd6-48a9-bdee-688c8f479316">

If we than set the industry_residual_greengas to 50 PJ, the following occurs: 
<img width="1357" alt="Screenshot 2024-01-05 at 10 05 26" src="https://github.com/quintel/etsource/assets/150331345/b130c1ef-b098-4361-894f-e14b531dbff8">

This leads to two issues, one described in #2987.

The second issue: 
As can be seen the excess energy from industry_residual_greengas has nowhere to go. Thus to make the energy balance correct again we need to add a greengas backup export node.

Notifying @mabijkerk, will pick this up myself.

kaskranenburgQ:
While reviewing this we discovered that this issue is more complicated then we initially thought. Adding a backup greengas export node as the other export nodes are modeled leads to an error in the graph. 

There could be 2 possible reasons:

1.  The data initialisation of greengas production in the graph could lead to some issues. This is not as straightforward as we thought.

2. In the prefered modeling route greengas is exported as the carrier network gas. This is due to the mxture of greengas and natural gas in the network. In the current state of the model natural gas is also exported as network gas. It could be that the graph does not allow to 2 flexible export nodes of network gas. 

Will need to investigate these 2 further.

mabijkerk:
There are a number of additional points about the modelling of green gas and natural gas:

- Preferably users would import and export network gas, currently they import natural gas and export natural gas and green gas. Import of green gas is not available
[...]

PR #3076, opened 2024-06-27
Greengas modeling fixes

kaskranenburgQ:
This PR tries to solve issue #2986 

It change the following in the energy graph:
It adds a node between the conventional greengas production nodes and the distribution node, energy_production_greengas.
It also adds a backup export node for greengas.


mabijkerk:
There are a few open to do's before we can merge this:

- [ ] Check if all relevant inputs produce expected annual results in all edge cases @kaskranenburgQ 
- [ ] Check if annual results match hourly results in all edges cases, especially for `network_gas.type = export` in the green gas backup export node @kaskranenburgQ 
- [ ] Update all ETLocal datasets on ETLocal and on ETSource to link the green gas production routes to the new `energy_green_gas_production` node @kndehaan 
- [ ] Decide whether to merge these changes into `master` and if so what other queries might have to be updated @mabijkerk @kaskranenburgQ

kndehaan:
ETLocal datasets are updated and exported to ETSource. Semaphore test on ETLocal should be re-run when branche ETSource is merged.

graphs/energy/nodes/energy/energy_export_greengas_backup.ad
kndehaan, line 1: Shouldn't his be 

`network_gas.type = export`

kaskranenburgQ: Good catch, @mabijkerk and I choose to model it with
`network_gas.type = consumer`. This is because a node with `network_gas.type = export` will act as a last resort balancer for the carrier network_gas. If you add this to both nodes, both nodes will act as a last resort balancer at the same time. 
This means that the export of gas will be doubled in the hourly calculations. 
Therefore we choose to go with `network_gas.type = consumer`, so that the export of gas would not be extremely high in the hourly calculations.

kndehaan (approved):
LGTM! I checked hourly and annual calculations and flows, I think it works as expected.

graphs/energy/nodes/energy/energy_export_greengas_backup.ad
mabijkerk: Preferably we would fix this at some point. See https://github.com/quintel/etsource/issues/3080.

PR #3079, opened 2024-06-28
Het Hogeland update train electricity demand

kndehaan:
Electricity demand for trains was set to 0 in ETLocal and now exported to the dataset in ETSource. Files that are changed as a result are:
- transport_final_demand_electricity
- transport_final_demand_electricity-transport_final_demand_for_road_electricity@electricity


KoenvanB (approved):
LGTM!

PR #3083, opened 2024-07-01
 Adding nodes for final demand of methanol

kaskranenburgQ:
This PR adds the final demand nodes for industry other non specified for non energetic methanol.
With that it also adds an methanol distribution node.


mabijkerk:
@kaskranenburgQ did you branch from the latest version `ctm-etm-coupling`? This branch has got some commits that do not relate to ammonia and has a conflict with the energy_node_positions.

mabijkerk:
@kaskranenburgQ why is there a commit related to ammonia on this branch?

mabijkerk (approved):
All looking good. I think this is a good one to properly test in deliverable 5, especially concerning primary emissions.

Before merging it, I think we should check if we should normalize the input shares of `energy_production_synthetic_methanol` to sum to 1.0. At the moment the output is set to 1.0 and I'm not sure if this was intentional. Relates to this PR on ETDataset: https://github.com/quintel/etdataset/pull/990. @kaskranenburgQ @kndehaan what do you think?

kndehaan:
I agree that inputs of `energy_production_synthetic_methanol` should add up to 1.0 and that there should be a loss output. I will solve this on ETDataset and ETSource and push it to this branche.

kndehaan:
The in- and output values of the NSA are now updated where the inputs add up to 1.0 and output has methanol < 1.0, therefore containing losses (related to [this PR](https://github.com/quintel/etdataset/pull/990))

graphs/energy/edges/energy/energy_production_synthetic_methanol-energy_distribution_methanol@methanol.ad
kndehaan, line 1: Shouldn't this be `type = share`, in line with e.g. edge from production kerosene to distribution kerosene?

kaskranenburgQ: Good spot, I think both are possible. While looking at the hydrogen modelling, flexible reversed edges are used intead of shares.

kaskranenburgQ:
> Additionally, we should check if this has fallout for any existing scenarios. Does updating it lead to a different value for a given scenario with the input `output_of_energy_production_synthetic_methanol` set?

I tested this for a blank 2019 scenario with the synthetic methanol production set to max. 
The dashboard KPI's do not change.
The input flows slightly change.
Screenshot 2024-07-03 at 16 34 06

Screenshot 2024-07-03 at 16 34 22

The biggest difference however, is in the demand calculation of the energy_production_synthetic_methanol node. This jumps from 110.3 to 140.15 PJ due to the different loss calculation.

I don't know if for these effects a migration is necessary, what do you think @mabijkerk ?

PR #3084, opened 2024-07-01
 Adding nodes for final demand of ammonia

kaskranenburgQ:
This PR adds the final demand nodes for industry other non specified for non energetic ammonia.


mabijkerk (approved):
Good to go!

PR #3089, opened 2024-07-03
Updated H2 pipeline costs

kndehaan:
This PR updates the investment and O&M costs of onshore H2 pipelines, based on the NSWPH dataset. This source is also used for the offshore H2 pipeline costs. 

Reason for updating the costs is that the offshore H2 pipeline costs were significantly lower than onshore pipeline costs. Also, the onshore pipeline costs were based on an old source originating from 2009. 

Goes with https://github.com/quintel/etdataset/pull/1004


kaskranenburgQ (commented):
LGTM!

PR #3090, opened 2024-07-08
Ctm etm coupling 2b hydrogen

kaskranenburgQ:
This PR adds hydrogen carrier to the metal production in the industry section for the ctm-etm coupling.
It does so by creating an external coupling 'process' to the aluminium sector, and letting the ctm override the other metals production node. All of the shares, efficienies, and demand for both these nodes can be overwritten by using the external coupling inputs.

Some remarks:
- The efficiency inputs do not reflect the inputs of the front-end, these are set by improving the efficiency procentually. With these inputs the efficiency is directly set.
- It's still debatable whether this is the most correct way to implement the external coupling of the other metals sector by complete taking over the node.


graphs/energy/nodes/industry/industry_aluminium_external_coupling_node.ad
kndehaan, line 1: Some questions / remarks:
- How are the set shares input and output shares of the presented carriers determined? It looks like it is calculated somehow? 
- energy_balance_group is outdated and can be omitted
- can graph_methods be omitted here since it doesn't contain any items?

kndehaan:
I started with reviewing some of the changes but think we need to have a discussion on certain decisions. Some thing to discuss or that I noticed:

**1. Set-up of external coupling industry_other_metals**

With the current energy graph it is not quite clear which energy carriers are 'active' when there's no external coupling. When there is external coupling, all energy carriers can be set that flow to industry_other_metals_production. However, multiple carriers (network gas, coal, crude oil) can now flow to the production node directly or via the industry_other_metals_burner_network_gas, which can result in double counting of carriers...Also, if this set-up is chosen, steam hot water should also be added as carrier that can flow to production. 

It seems to me however that a more clear design should be possible where a better distinction is made in the external coupling route and the non-external coupling route. However, implications, fallout, etc. should be discussed.

2. Why do the outputs of industry_other_metals_process_heat_useable_heat, industry_other_metals_process_electricity, etc. have a not defined output? Whereas the output of these nodes contain one carrier.
3. Profile of hydrogen demand: industry_final_demand_for_metal_aluminium_hydrogen has `hydrogen.group = flat`, whereas industry_final_demand_for_metal_other_metals_hydrogen has `hydrogen.group = industry_metals_heat`. Which is the correct consumption profile here for these final demand nodes? 
4. I posted some minor questions regarding nodes or edges which are specified in separate comments.

graphs/energy/nodes/industry/industry_aluminium_external_coupling_node.ad
mabijkerk: I have updated these values to reflect the electrolysis current process, correcting for losses in the network gas burner and normalizing the inputs and outputs. Graph methos and energy balance group are now omitted.

PR #3094, opened 2024-07-22
changed subtype to pseudo for heat pumps

louispt1:
This should close [this heatpump issue.](https://github.com/quintel/etengine/issues/1337) 
With the help of Nora's investigation powers, I changed the merit_order subtype to _pseudo_. Based on my initial checks, the input curve is now calculating appropriately.

### DIFF
--- a/graphs/energy/nodes/energy/energy_heat_boiler_ht_electricity.ad
+++ b/graphs/energy/nodes/energy/energy_heat_boiler_ht_electricity.ad
@@ -20,7 +20,7 @@
 - heat_network_ht.type = producer
 - merit_order.group = self: steam_hot_water_output_curve
 - merit_order.level = mv
-- merit_order.subtype = generic
+- merit_order.subtype = pseudo
 - merit_order.type = consumer
 - electricity_output_capacity = 0.0
 - free_co2_factor = 0.0
--- a/graphs/energy/nodes/energy/energy_heat_boiler_lt_electricity.ad
+++ b/graphs/energy/nodes/energy/energy_heat_boiler_lt_electricity.ad
@@ -20,7 +20,7 @@
 - heat_network_lt.type = producer
 - merit_order.group = self: steam_hot_water_output_curve
 - merit_order.level = mv
-- merit_order.subtype = generic
+- merit_order.subtype = pseudo
 - merit_order.type = consumer
 - electricity_output_capacity = 0.0
 - free_co2_factor = 0.0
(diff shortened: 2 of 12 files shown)


mabijkerk:
The same issue seems to occur for the heat boilers, for example `energy_heat_boiler_ht_electricity`. We should verify for all heat producers for `heat_networks` that have `electricity_input_conversion` larger than 0, whether an electricity input curve is assigned to them or not. If so, it seems that pseudo solves the issue here.

It would help to understand why pseudo solves this issue.

louispt1:
My understanding of Pseudo Consumers is that they consume resources in a way that depends on other calculations/conditions (they have dynamic demands).

For energy_heat_boiler_ht_electricity (& mt and lt), changing the - merit_order.subtype to pseudo does not solve this issue. The electricity_input_curve always sums to zero.

Maybe this is just because I don't know the model as well but I despite my playing around with quite a few settings, can't seem to get any electricity input into the boiler (with or without the pseudo subtype).

Screenshot 2024-07-30 at 13 18 31

I might need some help identifying why some of these heat producers (with electricity_input_conversion larger than 0) don't have any input at all.

louispt1:
Checking which heat producers for heat_networks with electricity_input_conversion larger than one led to a few more changes. Thanks to @mabijkerk  again for helping me learn a bit more about the modelling. It should be ready for review now.

### Added `pseudo` to these:
##### Nodes I checked:
- energy_heat_boiler_ht_electricity
- energy_heat_boiler_lt_electricity
- energy_heat_boiler_mt_electricity
- energy_heat_heatpump_drink_water_water_ts_mt_electricity
- energy_heat_heatpump_surface_water_water_ts_mt_electricity
- energy_heat_heatpump_waste_water_water_ts_mt_electricity

##### Nodes I inferred needed the subtype:
- energy_heat_heatpump_drink_water_water_ts_lt_electricity
- energy_heat_heatpump_surface_water_water_ts_lt_electricity
- energy_heat_heatpump_waste_water_water_ts_lt_electricity

### Also checked these but they seem fine to me so I did not alter them
- energy_heat_flexibility_p2h_heatpump_mt_electricity
- energy_heat_flexibility_p2h_heatpump_ht_electricity
- energy_heat_flexibility_p2h_boiler_mt_electricity
- energy_heat_flexibility_p2h_boiler_ht_electricity
- energy_heat_well_deep_mt_geothermal
- energy_heat_well_deep_ht_geothermal
- energy_heat_well_shallow_heatpump_mt_geothermal
- energy_heat_well_shallow_lt_geothermal

mabijkerk:
Seems clear to me! The `p2h` heating technologies are definitely a different type, so I'm not surprised that they don't require the `pseudo` attribute. I'm a bit surprised though that the `geothermal` technologies do work without the `pseudo` attribute, since they look similar to the heat pumps for me?

mabijkerk (approved):

louispt1:
Mathijs and I have seen that the geothermal ones must-run which is why they don't need to be pseudo consumers.

PR #3107, opened 2024-08-23
CTM-ETM coupling: generic transformation nodes

mabijkerk:
This PR adds generic transformation nodes for external coupling by the CTM:

- `energy_chemical_fertilizers_transformation_external_coupling_node`
- `energy_chemical_other_transformation_external_coupling_node`
- `energy_chemical_refineries_transformation_external_coupling_node`

These nodes have a set of input carriers and a set of output carries that allows a robust range of carrier transformations to be loaded from the CTM through external coupling inputs. Not defined input and output can be defined on the nodes as well, to indicate any carriers that are not in scope of the coupling.

**Note:** the robust range of carrier transformation does lead to potential circularities in the graph. This work package therefore includes changes that allow both hourly and annual methods to compute circularities. For the recursive factors we have opted for a workaround that nettifies disallowed inputs to deal with circularity. This is clearly not a full solution, so 'solving' circularities in recursive factors is still a desirable future improvement.

Goes with:
- Refinery: https://github.com/quintel/refinery/pull/62
- Atlas: https://github.com/quintel/atlas/pull/170
- ETEngine: https://github.com/quintel/etengine/pull/1455
- ETModel: https://github.com/quintel/etmodel/pull/4322


kaskranenburgQ:
The Scenario II3050V2 - Decentral Initiatives shows the following at the node:
`energy_chemical_other_transformation_external_coupling_node `

Screenshot 2024-08-26 at 10 03 55
The demand on the node is correct when comparing them to the the inputs.
However, the inputs and outputs of this node sum up to 1. 
Screenshot 2024-08-26 at 10 18 39


In this scenario, the node `energy_chemical_refineries_transformation_external_coupling_node ` is functioning properly:
Screenshot 2024-08-26 at 10 19 11
with the following input and output settings:

Screenshot 2024-08-26 at 10 19 47

Could it be that the migration for the `energy_chemical_other_transformation_external_coupling_node` is missing a *100 for the inputs and outputs?

kndehaan:
> Could it be that the migration for the energy_chemical_other_transformation_external_coupling_node is missing a *100 for the inputs and outputs?

I actually don't have this problem, so I think nothing went wrong with the migration. 

image

kndehaan:
There are quite some significant differences in the default dashboard items of various featured scenarios, for instance in II3050v2 2050 European Integration.

Local:
image


Beta:
image

It seems that especially biomass import differs significantly. Will dive further into this.

kndehaan (approved):
Approved for merge

PR #3113, opened 2024-09-10
Plexos coupling

louispt1:
This PR:
- Adds the coupling inputs for interconnectors.
- Backup profiles are included in case no capacity curve is uploaded when a coupling is activated.
- Inputs such as capacity and availability are disabled by the coupling.
- Full load hours are determined by the capacity profile for export.

Related to [ETModel 4332](https://github.com/quintel/etmodel/pull/4332) and [ETEngine 1459](https://github.com/quintel/etengine/pull/1459)


inputs/modules/external_coupling/interconnectors/external_coupling_export_interconnector_1.ad
kaskranenburgQ, line 1: We discussed that squashing the import and export input for must run would make more sense, since you would typically model the interconnector as a whole (import + export) in the external model. @mabijkerk @louispt1 
Is this a must-have or NTH?

kaskranenburgQ (changes requested):

inputs/modules/external_coupling/interconnectors/external_coupling_export_interconnector_1.ad
mabijkerk: Ah in that case I agree that this is more efficient. For the backend the impact seems limited. Does it involve front-end changes?

PR #3115, opened 2024-09-12
Robust queries for MYC

mabijkerk:
This PR solves https://github.com/quintel/multi-year-charts/issues/70. The edge group method and node group method of querying final demand per carrier leads to different results with the sustainability share. Therefore, the difference between fossil and renewable electricity is dropped from the MYC.

To do:

- [x] Update labels and queries in the MYC
- [x] See if the same changes should be implemented for Heat. @kaskranenburgQ do we want this?


kaskranenburgQ (approved):
LGTM!

PR #3118, opened 2024-09-19
Data update pnh

kaskranenburgQ:
Goes together with ETModel [PR 4339
](https://github.com/quintel/etmodel/pull/4339)

First update of the 2022 dataset for Noord-Holland, this is not the final form of the dataset.
We want to hide this dataset in the beta front-end, so that some initial testing can be performed by external parties. 
When the end product is done we will update the front-end again, so that the full work dataset is updated.

@mabijkerk Can you check if I took the right steps to hide this dataset in the front-end, and check if this could be put on beta?


mabijkerk (approved):
I only checked whether this dataset is hidden on etmodel and available on etengine. This is the case.

PR #3121, opened 2024-09-24
Adding a unique external coupling group for each interconnector.

kaskranenburgQ:
This PR adds a unique external coupling group for each interconnector.
This distinction creates the possibilty to set each interconnector to must-run seperately. 
In the current setup this is not yet possible.


mabijkerk (approved):
⚡

PR #3123, opened 2024-09-27
Renaming external coupling groups in regular front end inputs

kaskranenburgQ:
This PR adds the correct external coupling group in the disabled_by_couplings argument for each interconnector.
@kndehaan Could you check if this is now working correctly?


kndehaan (approved):
Works well now!

PR #3130, opened 2024-10-14
Great Britain country dataset

kndehaan:
Create Great Britain country dataset with as base the UK and Northern Island datasets, using the reversed amalgamator. Since there are some shortcomings of the reversed amalgamator and since the UK and N-I datasets are generated differently, some output values are manually corrected, specifically:

- Full load hours, efficiencies and share splits are directly taken from UK dataset
- Keys where subtractions make sense (e.g. final demand), UK minus N-I is done, except for when the value for N-I was higher than the UK value (causing a negative value as result). In such cases, the UK value is assumed. 

Goes with https://github.com/quintel/etlocal/pull/541


kaskranenburgQ:
LGTM! There still is an ETModel update necessary to get the locales right, I assume that this will be picked with the other new countries?

kaskranenburgQ (approved):
LGTM

PR #3134, opened 2024-10-24
Eu space heating data update

kaskranenburgQ:
This PR adds the Eurostat space heating data update to Etsource.
This builts upon the work done on branch european-datasets. 
It is advised that the work on that branch is merged before this branch.

Goes together with:
[PR Etlocal](https://github.com/quintel/etlocal/pull/554)
[PR ETdataset](https://github.com/quintel/etdataset/pull/1000)


kaskranenburgQ:
The semaphore tests fail on the Luxembourg(households final demand network gas) , Serbia(industry_final_demand_for_metal_other_metals_network_gas) & Croatia (households_final_demand_network_gas) datasets .

kndehaan:
> The semaphore tests fail on the Luxembourg(households final demand network gas) , Serbia(industry_final_demand_for_metal_other_metals_network_gas) & Croatia (households_final_demand_network_gas) datasets .

Did manual corrections for calculated shares (Excel rounding errors / floating point errors), now Semaphore is happy. Not a robust fix though.

kndehaan:
So I found out that the floating point error is actually caused in the source analysis excel after all @kaskranenburgQ .

For both Croatia and Luxembourg, the rounding error occurs in edge 
`households_final_demand_network_gas-households_final_demand_for_cooking_network_gas@network_gas:`

which is ETLocal key
`households_final_demand_network_gas_households_final_demand_for_cooking_network_gas_parent_share`

Can we already apply the finite rounding in the excel file, where in the excel file a check is done if the shares add up to 1.0, and if not, that a correction will be done?

mabijkerk:
If there error is indeed in the source Excel, this is good news since we then do not have to update the transformer gem (at least not for this project).

PR #3138, opened 2024-10-30
New country datasets and application split update

kndehaan:
With this PR, the new country datasets NO, CH, RS and GB are added to the ETM. Furthermore, the household application split values are updated for EU countries. 

Goes with https://github.com/quintel/etmodel/pull/4361


kaskranenburgQ (approved):
LGTM!

PR #3143, opened 2024-11-21
Adding spec for validation of node-calling in V function

kaskranenburgQ:
This PR adds the spec `validate_v_function_spec.rb`

This spec aims to find missing or misspelled nodes in instances of the V() function. 
It does this by following this logic:
It assumes that the V() function is structured as one of the following instances:
1. `V(node1,attribute_of_interest)`
2. `V(node1,node2,node3,attribute_of_interest)`
3. `V(function,attribute_of_interest)`

In all of the cases it will ignore the attribute_of_interest, since this is not a node.
It identifies the attribute by splitting the elements given the V() function by comma seperation.
It ignores all elements given to the V() function that are in uppercase letters, since these are functions and not nodes. It takes the nodes by looking at the folders graphs/energy/nodes and graphs/molecules/nodes. It seems to check instances of MV() as well.

Some examples that illustrate this use-case:
**Case 1: Valid Single Node + Expression**
`V(energy_heat_burner_mt_coal, "variable_costs_per_typical_input * SECS_PER_HOUR / heat_output_conversion")`

	•	energy_heat_burner_mt_coal is validated as a valid node.
	•	"variable_costs_per_typical_input * SECS_PER_HOUR / heat_output_conversion" is skipped as an expression.
	•	Passes.

**Case 2: Valid Single Node + Attribute**
`V(energy_heat_burner_mt_coal, demand)`

	•	energy_heat_burner_mt_coal is validated as a valid node.
	•	demand is skipped as an attribute.
	•	Passes.

**Case 3: Invalid Node**
`V(invalid_node, demand)`

	•	invalid_node is not a file in the two node folders.
	•	Fails.

**Case 4: Nested Function Call**
`V(CARRIER(biogas), cost_per_mj)`

	•	CARRIER(biogas) is skipped as a function.
	•	cost_per_mj is ignored as an attribute.
	•	Passes.

For now the mechanical_turk folder is skipped, since there are a lot of outdated nodes there.

A future improvement would be to check the input files as well.

The spec fails since there are 190 outdated or misspelled nodes in the gqueries folder. 
@noracato Could you check out if the spec is structured correctly?
In the meantime we can fix the failing of the spec


noracato:
Hi @kaskranenburgQ! I'm very sorry, but I can not approve this spec. If you want to add extra validations to GQueries, you should write a dedicated validator in Atlas. That validator will get automatically picked up in ETSource.

This folder contains the validators: 
https://github.com/quintel/atlas/tree/master/lib/atlas/active_document

Then in the [GQuery model](https://github.com/quintel/atlas/blob/master/lib/atlas/gquery.rb) you should add a `validates_with`.

The validator should also have it's own spec in the Atlas specs.

I think it's great that you tried to write your own spec, but next time you can give me a quick call and I can explain where to start 🙂

mabijkerk:
Thanks for your explanation @noracato. It would be good if at some point you could give @kaskranenburgQ a short introduction to writing specs. He can then rewrite the spec in Atlas afterwards.

PR #3153, opened 2024-11-27
Creating two queries used in the pmiek-tool to abstract the maximum a…

Charlottevm:
…nd minimum value of the residual load curve


mabijkerk (approved):
✅

PR #3155, opened 2024-12-01
Changes for external coupling of must-run interconnectors

mabijkerk:
While reviewing documentation of must-run interconnectors in https://github.com/quintel/documentation/pull/203, I started testing the functionality and ran into some issues. I changed the items listed below.

- Renamed all relevant inputs for clarity
- Moved inputs for FLH to the relevant external coupling input folder
- Fixed coupling groups for FLH inputs
- Removed disabled by from regular capacity inputs
- Normalized base curves sum to 1/3600 instead of 8760
- Added base curves to other relevant datasets

Notifying both @kaskranenburgQ and @louispt1. Would be good to have a discussion about this and see whether we can align on how the functionality should work, perhaps I've missed something. There are a couple of to-do's still left before we can merge:

- [ ] Must: fix aliases to the nl2019 curves that don't work yet (see Semaphore error)
- [ ] Must: discuss whether using a capacity profile is correct
- [ ] Must: discuss (with client) wether to migrate inputs
- [ ] Nice: discuss whether to incorporate the must-run input in other inputs

Goes with


mabijkerk:
> Due to these settings, the actual output capacity of the interconnector is around 52 GW in the first week of the year.

Well-spotted, I think this is probably why we prefer using capacity profiles. I'll discuss this with the client.

mabijkerk:
After discussing with the client we drew the following conclusions:

- There's no need for migration
- We should remove the FLH inputs: the goal should be to retrieve the FLH from the profiles. This would also prevent the issue noted by @kaskranenburgQ that states that the resulting capacity can exceed the specified capacity. This will however probably require us to update the `export_must_run` and `import_must_run` subtypes in January.
- We can then merge this PR without the ETModel PR, since that only contains locales for the FLH inputs

Issue #3080, opened 2024-06-28
Network gas hourly and yearly calculations do not add up


kaskranenburgQ:
In a blank nl2019 scenario with the following input settings:
```
natural_gas_total_share: 0.0
green_gas_total_share: 100.0
propane_gas_total_share: 0.0
```
All extracted natural gas is now exported.

If I than compare the natural gas backup export hourly vs yearly with the following query:
```
EACH(
  V(energy_export_natural_gas_backup,demand),
  SUM(V(energy_export_natural_gas_backup,network_gas_input_curve))*3600,
  SUM(NEG(V(energy_export_natural_gas_backup,demand)),SUM(V(energy_export_natural_gas_backup,network_gas_input_curve))*3600)
  )
```
I get the following results:
![Screenshot 2024-06-28 at 17 05 12](https://github.com/quintel/etsource/assets/150331345/3bfbb137-bbc1-4c03-9bbe-da1222006b7f)
Indicating a delta of 4.8 PJ. 
This could occur due to a network gas node missing hourly calculation attributes

For the import of natural gas, we can see the following for a blank 2019 scenario:
 
<img width="892" alt="Screenshot 2024-07-10 at 09 02 28" src="https://github.com/quintel/etsource/assets/150331345/79a56726-3cd2-4b2d-b6b6-b9ff05c02764">

With this query :

`EACH(
  V(energy_import_natural_gas,demand),
 SUM(V(energy_import_natural_gas,network_gas_output_curve))*MJ_PER_MWH,
)
`
 A delta of 6.5 PJ

In addition, the energy_heat_backup_burner_ht_network_gas node shows a slight difference in energy flows:
Balancing issue in https://beta.energytransitionmodel.com/saved_scenarios/5126. 
Output of steam hot water:
```
EACH(
  V(energy_heat_flexibility_p2h_boiler_ht_electricity,output_of_steam_hot_water),
  SUM(V(energy_heat_flexibility_p2h_boiler_ht_electricity,steam_hot_water_output_curve))*MJ_PER_MWH,
  V(energy_heat_flexibility_p2h_heatpump_ht_electricity,output_of_steam_hot_water),
  SUM(V(energy_heat_flexibility_p2h_heatpump_ht_electricity,steam_hot_water_output_curve))*MJ_PER_MWH,
  V(energy_heat_backup_burner_ht_network_gas,output_of_steam_hot_water),
[...]

PR #3156, opened 2024-12-03
Align hourly and annual heat network for DAC

mabijkerk:
This solves part of #3080. The DAC was connected to the HT heat network in the energy graph, but had `heat_network_mt` attributes. This caused a misalignment between annual and hourly balancing in the model. To re-align this, the DAC has been added to `heat_network_ht`, because the input temperature is assumed to be 100 degrees Celcius.

In addition, the `ht_heat` final demand group is added to the edge to the DAC, in order for the energy sector heat demand to show up in the Sankey.


kaskranenburgQ:
Very nice!
on beta/pro:

District heating supply and demand per temperature level 2631680
now:
District heating supply and demand per temperature level 2625762

PR #3157, opened 2024-12-03
Dataset retirement

kaskranenburgQ:
This PR retires the following datasets from ETsource:

- DKGM751_aarhus.derived
- example.full

Goes together with:
ETSource: https://github.com/quintel/etsource/pull/3157
ETDataset: https://github.com/quintel/etdataset/pull/1026
ETLocal: https://github.com/quintel/etlocal/pull/562
internal: https://github.com/quintel/internal/pull/15
ETModel: https://github.com/quintel/etmodel/pull/4374


mabijkerk (commented):
I see the example dataset still listed in `curves.yml` and `dataset.yml`. I am not sure what these files are for. Do you know if we can delete them @noracato?

noracato:
> I see the example dataset still listed in curves.yml and dataset.yml. I am not sure what these files are for. 

These are helper files for rake tasks. For instance if you wanted to bulk update all datasets or curves with imports from ETDataset. I don't believe that is still widely used, as we have almost none of those datasets left. So you can remove anything from their contents which is not up to date anymore!

PR #3162, opened 2024-12-11
Removing greengas flows from biomass to heat conversion query

### LINKED ISSUE #3159: Greengas flows to heat are counted double in sankey energy flow overview
In [this scenario](https://energytransitionmodel.com/saved_scenarios/18884) we see that the heat in- and output does not match. 
Energy flow overview 1250051

If we zoom in on the heat flows we see the following:

Screenshot 2024-12-06 at 16 13 47

The biomass and network gas flows are equal (both 0.93 PJ)
This is because in the query behind the biomass flows to heat, we count the greengas/network gas flow that is used in heat production. 
This flow is also accounted for in the network gas to heat query. 

I therefore propose we remove this flow from the biomass to heat query. 
@mabijkerk @kndehaan what do you think?

kaskranenburgQ:
This PR removes greengas from the biomass to heat query. 
With the addition of the network gas conversion step in the energy flow overview sankey, we had a double count in the flows of greengas to heat production. This flow is also counted via the biomass products --> network gas --> Heat flows. 

Closes #3159

### DIFF
--- a/gqueries/general/conversion/heat/conversion_from_carrier_to_heat/conversion_from_biomass_products_to_heat.gql
+++ b/gqueries/general/conversion/heat/conversion_from_carrier_to_heat/conversion_from_biomass_products_to_heat.gql
@@ -9,13 +9,6 @@
       Q(conversion_from_wood_pellets_to_heat),
       Q(conversion_from_bio_lng_to_heat),
       PRODUCT(
-        PRODUCT(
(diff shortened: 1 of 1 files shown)


kndehaan (approved):
I also removed the query of the sustainable share of network gas in gas power fuel mix to heat since this is also accounted for in the network gas to heat query. Other than that it looks good!

PR #3169, opened 2025-01-03
Mekko main carriers electricity

kaskranenburgQ:
This PR adds the mekko for supply and demand of the electricity carrier. 

Goes together with: 
https://github.com/quintel/etmodel/pull/4382

Please note that the mekko does not balance properly for the scenario II3050v2- NAT at the moment. 
I could partially find the underlying cause. 

We also added a great deal of new queries that output the electricity on a node level on this branch. 
These are not used for the queries in the final graph since these are based on groups. 
Do we want to remove these queries or keep them?


kndehaan (commented):
Looking good! Nice use of converter groups. 

I moved the electricity production queries aggregated per technology type (made by @kaskranenburgQ) to `gqueries/general/electricity/electricity_production_per_technology_aggregated`. The electricity production queries per technology made by @claudiavalkenier are still in `gqueries/general/electricity/electricity_production_per_technology`. I think there's no harm in keeping these queries, since they might be useful in the future and quite some work and effort has been put in creating these queries. @mabijkerk what is your opinion on this?

gqueries/general/converter_groups/chps_per_plant_type/chp_biogas.gql
mabijkerk, line 1: Can we rename these queries to be a bit more verbose? Something like converter_group_chp_biogas or chp_converters_biogas.

gqueries/general/converter_groups/chps_per_plant_type/chp_coal.gql
```diff
@@ -0,0 +1,6 @@
+- query =
+    FILTER(
+      Q(electricity_producers_coal),
+      "steam_hot_water_output_conversion > 0"
```
mabijkerk, line 4: We typically specify both digits, so `0.0` instead of `0`. This ensures that the number is read as a float, instead of as an integer.

gqueries/general/electricity/electricity_demand_split_by_application/households_final_demand_for_electricity_in_space_heating_and_hot_water.gql
mabijkerk, line 1: If we now have these queries as 'base' queries in the `general` folder, shouldn't the Sankey download queries refer to these directly as well. See for example the query `sankey_final_demand_of_electricity_in_households_space_heating_and_hot_water_energetic`.

gqueries/general/electricity/electricity_production_per_technology/energy_battery_solar_electricity_output_of_electricity.gql
mabijkerk, line 1: I think it worth considering deleting these queries: if we have no way to check whether they are complete or up to date, they are likely to become outdated.

gqueries/general/electricity/electricity_curtailment.gql
```diff
@@ -0,0 +1,15 @@
+- query =
+  DIVIDE(
+    SUM(
+      V(
+        energy_flexibility_curtailment_electricity,
+        energy_hydrogen_curtailed_electricity,
+        energy_power_hybrid_curtailed_electricity,
+        energy_battery_curtailed_solar_electricity,
+        energy_battery_curtailed_wind_electricity,
```
mabijkerk, line 9: I don't think these should be included. If we only show the **output** of the integrated battery systems **to the grid** in the mekko, then we do not have to include the curtailment right?

gqueries/general/electricity/electricity_conversion_to_hydrogen.gql
mabijkerk, line 1: Can we make this query more robust?

PR #3170, opened 2025-01-03
Mekko main carriers network gas

kaskranenburgQ:
This PR adds the network gas mekko. 
The branch used in this PR is based upon PR #3169, therefore a lot of files have been changed while comparing this branch to master.

Goes together with:
https://github.com/quintel/etmodel/pull/4383


gqueries/general/final_demand/mece_energetic/final_demand_of_network_gas_in_agriculture_energetic.gql
mabijkerk, line 1: Why are these queries necessary? We already have the queries final_demand_of_natural_gas_and_derivatives and they seem more robust.

gqueries/general/network_gas/network_gas_conversion_to_central_heat.gql
mabijkerk, line 1: Aren't these conversion queries already there for the Sankey CSV? Shouldn't we reuse those?

gqueries/general/final_demand/mece_energetic/final_demand_of_network_gas_in_agriculture_energetic.gql
kndehaan: The reason why I created these new queries is because `final_demand_of_natural_gas_and_derivatives` include final demand of LNG and propane as well. The network gas mekko currently consists of supply categories relevant for the central network gas network. Regarding LNG, this means that only supply of network gas from regasification of LNG is queried, instead of the total LNG supply from distribution (or import). Therefore, the final demand queries in the mekko should not query final demand of LNG. For propane, this is less relevant since all import of propane is currently mixed in the network gas mix and there is no final demand for propane in the ETM. To be consistent with the naming of `network_gas`, I now left out final demand of `natural_gas` in this query as well, which currently doesn't matter since there's no final demand for `natural_gas`. 

I understand that these queries are somewhat less robust than the natural_gas_and_derivatives queries, but using those queries will make balancing of the mekko incorrect. Unless we add all flows related to LNG as well on the supply side. But I'm not sure if that's entirely correct for a mekko that should represent the flows of the central network gas network... 

Please let me know your thoughts on this @mabijkerk

mabijkerk: Right, thanks for the explanation!

mabijkerk (commented):
Can be merged if issue with commit history is addressed.

PR #3173, opened 2025-01-08
Bug fix in edge type ICT load shifting

kndehaan:
Change to flexible


mabijkerk (approved):
The current edge type `share` does not work, because the supplied electricity from load shifting causes the input of electricity on the "after shifting" node to exceed 100%. As indicated in the the [documentation](https://docs.energytransitionmodel.com/contrib/load-shifting#graph-structure), the edge type needs to be `flexible`. This solves the issue.

PR #3175, opened 2025-01-09
Residences space heating hybrid oil

kaskranenburgQ:
This PR closes https://github.com/quintel/etmodel/issues/4375

It uses queries similar to the other queries used in the "Number of residences per space heating technology" graph.

Goes together with:
https://github.com/quintel/etmodel/pull/4389


kndehaan (approved):
LGTM!

PR #3185, opened 2025-01-13
Removing external coupling inputs for P2H WTP

kaskranenburgQ:
This PR removes the external coupling inputs for power-to-heat willingness-to-pay in the industry. 

It consists of 5 input files that are removed. In addition, the regular counterpart inputs are altered since they no longer need to be hidden in the front-end. 

Goes together with:
https://github.com/quintel/etengine/pull/1488
https://github.com/quintel/etmodel/pull/4391


mabijkerk (approved):
Nice!

PR #3186, opened 2025-01-15
Adding dedicated offshore wind electrolyser node to flh input statement

kaskranenburgQ:
This PR updates the FLH input statement of the dedicated offshore wind electrolyzer to include both the electrolyzer and the offshore wind turbine nodes. 

This solves the issue concerning the offshore wind electrolyzer in https://github.com/quintel/etengine/issues/1487.


mabijkerk:
Looks good. For clarity, @kaskranenburgQ, should we add one line of comment to indicate why the electrolyser scales with the wind turbine?

inputs/full_load_hours/flh_of_energy_hydrogen_wind_turbine_offshore.ad
```diff
@@ -1,8 +1,10 @@
 - query =
     IF(
       EQUALS(AREA(weather_curve_set), "default"),
```
mabijkerk, line 3: It seems like this input is only applied when the weather curve set is at "default". What happens when this is not the case? Are the full load hours of the hydrogen wind turbine and/or the electrolyser still overwritten?

kaskranenburgQ: When the curve set is set to a one of the weather years we support (1987,1997,2004), this input is disabled. 
In the input file settings_weather_curve_set, the flh's for the hydrogen dedicated producers are updated, the electrolyser counterpart is not. This should also be added in this input, otherwise the FLH's of the electrolyser and powerplant will not match.

PR #3174, opened 2025-01-09
Splitting non-energetic and delayed emissions in co2 chart

kaskranenburgQ:
This PR splits the non-energetic and delayed emissions in the co2 chart. 
In addition, some documentation about the modeling choices cocnerning non energetic final demand of methanol is added.

Goes together with: 
https://github.com/quintel/documentation/pull/213

Issue #2923, opened 2023-06-16
There is an inconsistency between the CO2 emissions graph and CO2 part of the Total greenhouse gas emissions graph


DorinevanderVlies:
There is an inconsistency between the CO2 emissions graph and the Total greenhouse gas emissions graph when it comes to the CO2 part (not the other GHG emissions part). Do you know where this small difference comes from? 

![image](https://github.com/quintel/etsource/assets/19907658/5908bdba-d312-4367-a062-97997eb31161)

The scenario that corresponds to this issue is confidential. Please contact @mabijkerk or @DorinevanderVlies for more info.

mabijkerk:
As it turns out @Charlottevm has worked on the GHG charts and the CO2 sheet that goes with it. I will therefore list the issues here as I encounter them and will discuss them with Charlotte later.

- [ ] The Other sector should be added to the queries
- [ ] The Other sector should be added as a slide to the frontend
- [ ] The Other sector should be added to the CO2 sheet
- [ ] The CO2 sheet queries are not based on the existing CO2 emission queries and are therefore in risk of divergence, preferably they would be rewritten and all refer to the standard CO2 emission queries

kaskranenburgQ:
> Sector: the Other sector should be added as a slide to the frontend

It's not clear to me how this should be implemented.

The sliders for e.g. the Households sector are percentual sliders based upon the area attribute `energetic_emissions_other_ghg_households`.
A similar area attribute for the other ghg emissions of the other sector does not exist, meaning that a percentual slider like the other sectors have can't be implemented. 

How can we implement this?

kndehaan:
This [PR](https://github.com/quintel/etsource/pull/3216) solves the discrepancy between the CO2 emissions in the CO2 emissions chart and GHG emissions chart. It also tackles the first todo mentioned in the comment of @mabijkerk 

> Sector: the Other sector emissions are allocated to the built environment in co2_sheet_buildings_households_other_co2_emissions and should be in a separate category

However, to fully address the other mentioned todos regarding the CO2 sheet (add Other sector, apply robust queries, etc.), it would be good to first have a discussion on what we want with the content of the CO2 sheet. It seems quite outdated, so a larger effort would be required to update this properly and align this with charts in the regular frontend.

kndehaan:
[...]

PR #3187, opened 2025-01-15
Apply robust query for non-energetic primary CO2 emissions

kndehaan:
This PR updates the query `emissions_non_energetic_all_co2`, used in chart 'Total greenhouse gas emissions (CO2 eq)'. Changes:
- `primary_co2_of_non_energetic` is used to query the modelled non-energetic CO2 emissions in a robust way
- Query notes are updated

This PR addresses the observation mentioned in https://github.com/quintel/etsource/pull/3174#issuecomment-2586506265


kndehaan:
Closing this PR as we should take up this improvement in a bigger effort to improve all queries in the GHG emissions chart as pointed out in https://github.com/quintel/etsource/issues/2923

PR #3190, opened 2025-01-21
Update external coupling inputs technical lifetime min and start value

kndehaan:
The min and start value of the external coupling inputs that set technical lifetime are set to non-zero value. This avoids the inputs to be set to zero, which breaks the `dashboard_total_costs` query in loading the ETM front-end.


kaskranenburgQ (approved):
LGTM!

Issue #3002, opened 2024-01-22
Geothermal energy deep should not be coupled to the slider of geothermal shallow


MartLubben:
When the slide geothermal (shallow) is set to 80 MW and geothermal (deep) to 0 MW, the chart shows both 80 MW. 

<img width="1062" alt="image" src="https://github.com/quintel/etsource/assets/33415156/c56385ee-fd1b-46dd-9475-380a1f9f0f81">

mabijkerk:
Thanks for spotting this @MartLubben. This issue is closed by https://github.com/quintel/etsource/pull/3004

PR #3200, opened 2025-01-27
Pv curtailment

### LINKED ISSUE #3199: Investigate differences in curtailment behaviour for solar PV plant and other solar PV types
**Background**
In the [IP2024 - National Drivers](https://energytransitionmodel.com/saved_scenarios/14389) scenario (2023), the following settings are made for the curtailment of solar PV. Note this relates to "pre" curtailment, also known as overplanting, where the grid connection is underdimensioned compared to the peak electricity output capacity. "Post" curtailment is the curtailment applied when, after balancing supply and demand, inflexible supply exceeds total demand.

Image

The expectation is that these sliders determine the reduction of peak capacity. Looking at the results in the table below, it is clear that this is the case for solar PV plants, where a 50% slider reduces the peak demand with 50%. For example for households however, this is clearly not the case:

Image

**Issue**
It is clear that the settings that do work for solar PV plants, do not work properly for at least households and buildings PV. This has likely happened when the households and buildings PV sliders were changed from a percentage of total potential to the absolute capacity sliders that there are now.

On a new `nl2019` [scenario](https://tyndp2024.energytransitionmodel.com/saved_scenarios/194) on the TYNDP 2024 server, where the sliders settings are still based on percentage of total potential, curtailment of 50% shows to correclty translate to the table:

Image

**Solution**
When changing the sliders to absolute capacity sliders, we probably forgot to fully model them as 'regular' volatile power producers, similar to the solar PV plant, for a couple of inputs. When searching for the inputs where `energy_power_solar_pv_solar_radiation` is mentioned, there is a difference with households and buildings in at least the following inputs:

- flh_of_solar_pv_solar_radiation_user_curve
- flh_of_solar_pv_solar_radiation
- settings_weather_curve_set
- efficiency_energy_power_solar_pv_solar_radiation

I propose to thoroughly examine the inputs for the following nodes, to see if they can and should be aligned with `energy_power_solar_pv_solar_radiation`:

- [ ] Households PV
- [ ] Buidlings PV
- [ ] Offshore PV
- [ ] Hydrogen PV
- [ ] Batttery PV

kaskranenburgQ:
This PR closes #3199 by updating the capacity_of_buildings_solar_pv_solar_radiation and the capacity_of_households_solar_pv_solar_radiation.

In addition this PR partly solves #3002 by revising the flh inputs for the following nodes:

- households_solar_pv_solar_radiation
- buildings_solar_pv_solar_radiation
- energy_power_solar_pv_offshore
- energy_battery_solar_pv_solar_radiation

For energy_hydrogen_solar_pv_solar_radiation this issue still remains, since this producer is not in the preset_demand group. At the moment, this is not a great issue since dedicated hydrogen pv plants are not in any of the datasets.

### DIFF
--- a/inputs/flexibility/weather/settings_weather_curve_set.ad
+++ b/inputs/flexibility/weather/settings_weather_curve_set.ad
@@ -32,11 +32,11 @@
       cool_factor = 1.0 + 0.5669449 * x + 0.1598105 * x**2 + 0.0253755 * x**3 + 0.00163210 * x**4;
       original_number_of_solar_parks = V(energy_power_solar_pv_solar_radiation, number_of_units);
 
-      solar_original_nou = V(energy_power_solar_pv_solar_radiation, number_of_units);
-      solar_households_nou = V(households_solar_pv_solar_radiation, number_of_units);
-      solar_buildings_nou = V(buildings_solar_pv_solar_radiation, number_of_units);
-      solar_offshore_nou = V(energy_power_solar_pv_offshore, number_of_units);
-      solar_battery_nou = V(energy_battery_solar_pv_solar_radiation,number_of_units);
+      orginal_number_of_power_solar_pv = V(energy_power_solar_pv_solar_radiation, number_of_units);
+      orginal_number_of_households_solar_pv = V(households_solar_pv_solar_radiation, number_of_units);
+      orginal_number_of_buildings_solar_pv = V(buildings_solar_pv_solar_radiation, number_of_units);
+      orginal_number_of_power_solar_pv_offshore = V(energy_power_solar_pv_offshore, number_of_units);
+      orginal_number_of_battery_solar_pv = V(energy_battery_solar_pv_solar_radiation,number_of_units);
       original_number_of_wind_turbine_coastal = V(energy_power_wind_turbine_coastal, number_of_units);
       original_number_of_wind_turbine_inland = V(energy_power_wind_turbine_inland, number_of_units);
       original_number_of_wind_turbine_offshore = V(energy_power_wind_turbine_offshore, number_of_units);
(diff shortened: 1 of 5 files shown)


mabijkerk:
@kaskranenburgQ when I try to run the National Drivers (IP2024, 2030) scenario locally with this branch, I get the following error:
```
(eval):1: syntax error, unexpected constant, expecting ')'
...ased_on_number_of_units))UPDATE(AREA(),solar_pv_profile_1_sh...
...                         ^~~~~~
(eval):1: syntax error, unexpected ',', expecting '}'
...,solar_pv_profile_1_share,1.0),UPDATE(AREA(),solar_pv_profil...
...                              ^
(eval):1: syntax error, unexpected ')', expecting end-of-input
...,solar_pv_profile_2_share,0.0)) }
...  
```
It seems that there's a syntax error somewhere in the changed inputs?

mabijkerk:
> In addition this PR partly solves https://github.com/quintel/etsource/issues/3002 by revising the flh inputs for the following nodes:

The issue linked here does not seem to be related, or am I mistaken?

mabijkerk:
The FLH inputs contain the following comment:
```
# The query starts by fetching the original number of units (nou): we want the new FLH
# to increase the preset demand, so we have to set this value after updating FLH
# to prevent ETEngine from calculating a new (lower) NoU.
#
# See https://github.com/quintel/etmodel/issues/2906
```

The link to the issue is only to add explanation for the first part of the comment right? To me it seems confusing to link the issue: it suggests that there's still something wrong with this input. Instead, the comment should be self-explanatory (which it already is?). What do you think @kaskranenburgQ?

kndehaan:
The work done in https://github.com/quintel/etsource/commit/e46151a0e877f2863a5c40f9574a435c6d2a2929 regarding FLH input of dedicated hydrogen solar is moved to this branch. A separate branch will be opened to tackle the influence of the area attribute.

kndehaan:
Fixed with https://github.com/quintel/etsource/pull/3221

PR #3202, opened 2025-01-30
Update dedicated hydrogen solar attributes with consistent decimals

kndehaan:
Inconsistency in rounding of dataset values related to dedicated hydrogen production from solar PV caused imbalance in flows between `energy_hydrogen_solar_pv_solar_radiation`, `energy_hydrogen_electrolysis_solar_electricity` and  `energy_hydrogen_curtailed_electricity`, as described in https://github.com/quintel/etengine/issues/1487#issuecomment-2592264861. It concerns:
- Area attribute `hydrogen_electrolysis_solar_pv_capacity_ratio`
- Full load hours of `energy_hydrogen_electrolysis_solar_electricity` in central_producers.csv
- Shares defined in energy_hydrogen_solar_pv_solar_radiation_parent_share.csv

In this PR a rounding to 8 decimals is applied. This nr is (among others) determined by testing the flows in the `EU27_european_union_27_countries` dataset when dedicated hydrogen from solar PV was set to the max possible capacity of its input. 

Goes with https://github.com/quintel/etdataset/pull/1030


datasets/GM0345HA_het_ambacht_veenendaal/GM0345HA_het_ambacht_veenendaal.derived.ad
mabijkerk, line 1: Shouldn't this dataset be retired?

mabijkerk:
I haven't checked everything yet, but so far, it look good. However, with the changes, still there remains a minor differences between the edge demand and the input of electricity. For example with 1 GW of dedicated solar PV in a blank `nl2019` scenario:

```ruby
EACH(
  V(energy_hydrogen_electrolysis_solar_electricity,input_of_electricity),
  V(EDGE(energy_hydrogen_solar_pv_solar_radiation,energy_hydrogen_electrolysis_solar_electricity),demand)
)

[
  10,186,120,396.363634,
  10,186,120,183.848555,
]
```

Another country, like `AT_austria`, also still returns a very minor difference. Is this difference fixable at all, or have we reduced it as much as makes sense and is the actual solution to do the calculations dynamic within the model, instead of static as preprocessing?

kndehaan:
> Is this difference fixable at all, or have we reduced it as much as makes sense and is the actual solution to do the calculations dynamic within the model, instead of static as preprocessing?

Allowing more decimals seems to reduce the difference somewhat, but a slight difference remains as a result of the static preprocessing.  For example, when I use the 'complete' numbers for Austria (full decimals as calculated in the CSV file):

```
hydrogen_electrolysis_solar_pv_capacity_ratio: 0.356681340488
FLH energy_hydrogen_electrolysis_solar_electricity: 2.6796765426E+03
parent share energy_hydrogen_electrolysis_solar_electricity: 9.54835785504E-01
parent share energy_hydrogen_curtailed_electricity: 4.51642144960001E-02
```

I get:

```
EACH(
  V(energy_hydrogen_electrolysis_solar_electricity,input_of_electricity),
  V(EDGE(energy_hydrogen_solar_pv_solar_radiation,energy_hydrogen_electrolysis_solar_electricity),demand)
)

[
  14,616,417,505.090908,
  14,616,417,505.101406
]
```

However, the same approach for nl2019 doesn't further decrease the differences in results, compared to what you reported @mabijkerk. Therefore, I think that the difference is not fixable by just increasing the allowed decimals of inputs. The best way to solve the difference is to do dynamic calculation of curtailment.

datasets/GM0345HA_het_ambacht_veenendaal/GM0345HA_het_ambacht_veenendaal.derived.ad
kndehaan: Dataset will be retired through these PRs:
- https://github.com/quintel/etsource/pull/3203
- https://github.com/quintel/etlocal/pull/566

PR #3204, opened 2025-02-05
Treatment losses natural gas

### LINKED ISSUE #2144: Network gas hourly balance should include treatment losses
The ``energy_treatment_natural_gas`` node has a 1.4% output loss. Including these losses in the hourly network gas calculation results in an incalculable graph. This is because:
- Including the losses in the time-resolved calculation requires a new node that contains these losses (so we can add a type and profile to this node)
- However, adding this new node means that Refinery can no longer balance imports and exports, as losses depend on total demand of the ``energy_treatment_natural_gas`` node and so do imports and exports

Discussed this with @antw and I think there are four possible solutions:

- Remove the losses from the ``energy_treatment_natural_gas`` node (they are not on the EB, but maybe we include them to account for the conversion of H-gas to L-gas?)
- Move the losses to ``energy_national_gas_network_natural_gas`` (this node already has a 0.1% loss) or maybe to ``energy_import_natural_gas`` (if possible).
- Do not include the losses in the hourly calculation (like we do now). This has no effect on the total year volumes in the graph, but it does mean that the hourly demand curves are slightly off. The missing loss is either assigned to the import curve (in a scenario with import of natural gas) or to the export curve (in a scenario where domestic production is higher than demand).
- Alter Refinery such that the graph _can_ be calculated when losses are included. This is possible but not trivial.

Notfiying @AlexanderWirtz, @ChaelKruip, @marliekeverweij: Which option do you prefer?

kaskranenburgQ:
This PR partially closes #2144 

It implements the first 3 steps of the approach by @mabijkerk in said issue:

> 1. The current treatment node is renamed to balancing node and the loss is removed.
> 2. A new treatment node is added that only applies to extraction.
> 3. This node has either energy use (of natural gas, oil and electricity) or losses.

For the 4th step:

> This energy use or loss is based on values from the energy balance.

More research and coordination is required, as well as implementation of data pipelines.

For now I have implemented the same natural gas output ratio on the energy_treatment_natural_gas node as the current node has  (0.986016667985657). 
@mabijkerk we had talks that this should be changed to 1.0 at one point, curious to hear what your thoughts with this implemtation.

I could not test these changes with the mechanical turk specs at the moment, therefore I put this PR as a draft.

Goes together with:
https://github.com/quintel/etmodel/pull/4400


kaskranenburgQ:
I found that using the query structure:
>         V(
>           Q(time_resolved_network_gas_demand_converters),
>           input_of_loss
>         )

To account for all losses does not work since the hybrid heatpumps for space heating and hot water have input of loss in their nodes to balance the future year calculations. 
This can be seen while running the query `TXT_TABLE(Q(time_resolved_network_gas_demand_converters),key,input_of_loss)`:

Screenshot 2025-02-06 at 11 57 54

We should therefore query the output of the treatment and network nodes. 

I wil open an issue for the balancing of the hybrid heatpumps.

kndehaan:
Looking good! Just a minor comment, I wonder if the edge type from treatment to the loss node, currently set to reversed share, is correct. I think in the current setup it works fine. However, if more nodes would ever be added in this set up in the future, would it make sense to set it to flexible or dependent to make sure it balances? (dependent is used for other loss nodes in the energy graph, but the documentation states that this attribute value is legacy)

mabijkerk (changes requested):
@kaskranenburgQ did you rename the share of gas input? We can only do that if we also have a migration to go with it. As it stands, I might be in favour of not renaming the input until we have released the stable version.

mabijkerk (approved):

PR #3208, opened 2025-02-11
Lower minimum value for buildings insulation inputs

mabijkerk:
**Issue**
In Cyprus, the start value for buildings insulation is 24.34 kWh/m2. This is also the minimum value for Cyprus, because it is set to be the lower value of either the start value or 25.0 kWh/m2. The issue here is that this means that users are not able to reduce heat demand for buildings in Cyprus.

**Solution**
To fix the issue addressed above, the hardcoded minimum value is set to 10 kWh/m2. This should gives users enough room to reduce heat demand for buildings.

@kaskranenburgQ could you please verify that buildings insulation sliders can be set lower than 25 kWh/m2, to 10 kWh/m2 at the least? I was not able to test it on my local model.


kaskranenburgQ (approved):
LGTM! Setting the sliders to 10.0 is possible now.

PR #3209, opened 2025-02-11
update of transport emissions for aviation

Max-Kerpel:
Fixed error in aviation emissions


Charlottevm (approved):
LGTM! nice work

datasets/RES28_west_brabant/RES28_west_brabant.derived.ad
kaskranenburgQ, line 1: It seems a different value has slipped in here.
I think it's correct that this value is set to 0.0078, since this is also the value dat is in the parent dataset (nl2016).
But it's good to check this.

kaskranenburgQ: I see now that we missed this dataset in the PR for the captured biogenic prices, therefore LGTM!

kaskranenburgQ (approved):

PR #3212, opened 2025-02-12
Updating Antwerpen household heat demand to belgium values

kaskranenburgQ:
This PR updates the Antwerpen data of household heat demand to the values of Belgium. 
This is a temporary update, since better for specifically Antwerpen is necessary.
With this update, the number of units in the household heating demand can be attributed accordingly by Fever. 

Goes together with:
https://github.com/quintel/etlocal/pull/572


Charlottevm (approved):
Nice work!

PR #3214, opened 2025-02-13
Lower minimum value for households insulation inputs

mabijkerk:
Similar to https://github.com/quintel/etsource/pull/3208, the minimum value for insulation sliders in Cyprus for households is not much lower than the default value. Therefore, users do not have a lot of leeway to reduce heat demand in Cyprus. As requested by a client this PR therefore lowers the minimum values to align with those in the buildings sector.


louispt1 (approved):
LGTM!

PR #3216, opened 2025-02-18
Align CO2 emissions in GHG chart with CO2 emissions chart

kndehaan:
This PR aligns the energetic and non-energetic CO2 emissions depicted in the GHG emissions chart with the emissions depicted in the CO2 emissions chart. This is achieved by applying the same underlying, robust queries for both charts.

With this PR:
- Queries used for the CO2 emissions chart are moved to the general folder
- Both the queries used in the CO2 emissions chart and the GHG emissions chart now use the same queries from the general folder. Note that the energetic and non-energetic CO2 emissions in the GHG emissions chart are now queried **without using the CO2 sheet queries**
- Some first effort has been put into the issues mentioned [here](https://github.com/quintel/etsource/issues/2923#issuecomment-1598610385) regarding the CO2 sheet. Specifically, the PR excludes CO2 emissions from the Other sector from the emissions in the built environment and creates separate queries for emissions in the Other sector. These queries still need to be used in the CO2 sheet however. To fully solve the issues with the CO2 sheet, further discussion is required on what we want to show with the CO2 sheet. 

This PR solves the main issue mentioned in https://github.com/quintel/etsource/issues/2923 (solving discrepancy in CO2 emissions between the charts for CO2 emissions and GHG emissions).  

Goes with https://github.com/quintel/etmodel/pull/4411


kaskranenburgQ (approved):
LGTM!

PR #3218, opened 2025-02-20
Remove green gas series in MYC to fix double-counting

kndehaan:
This PR removes the green gas series in the final demand per carrier and final demand per sector per carrier overviews in Collections. This PR includes:
- Removal of green gas series, locales and myc queries
- Updating final demand natural gas and derivatives query to also include green gas

Goes with https://github.com/quintel/multi-year-charts/pull/84

Solves https://github.com/quintel/multi-year-charts/issues/82

### DIFF
--- a/gqueries/modules/multi_year_charts/final_demand_chart/per_carrier/myc_final_demand_of_green_gas.gql
+++ b/gqueries/modules/multi_year_charts/final_demand_chart/per_carrier/myc_final_demand_of_green_gas.gql
@@ -1,8 +1,0 @@
-# Final demand of greengas carrier group
-
-- unit = MJ
-- query =
-    SUM(
-      V(G(final_demand_group), input_of_network_gas).sum * V(energy_national_gas_network_natural_gas,greengas_input_conversion),
-      V(G(final_demand_group), input_of_compressed_network_gas).sum * V(energy_national_gas_network_natural_gas,greengas_input_conversion)
-    )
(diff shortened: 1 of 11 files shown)


gqueries/modules/multi_year_charts/final_demand_chart/per_carrier/myc_final_demand_of_green_gas.gql
kaskranenburgQ, line 1: This query is still used in the Renewables chart via the query 'myc_final_demand_of_renewable_gaseous_biofuels'. 
I see two options:
1. Restore this greengas query. 
2. Remove the myc_final_demand_of_renewable_gaseous_biofuels query from the Renewables chart since we would like to review this chart anyway. See: https://github.com/quintel/multi-year-charts/issues/83. 
I have no strong opinion, @kndehaan or @mabijkerk do you?

If we go for option 2, we should also remove this query from the query myc_total_final_demand_of_renewables.

mabijkerk: Good catch @kaskranenburgQ.

I think it's a bit too straightforward to simply delete the green gas series from the Renewables chart, simply because we want to revise the chart at some point. I would go for option 1.

kndehaan: Good catch indeed! Agree with option 1, I will process this.

kaskranenburgQ (approved):
LGTM!

Issue #2706, opened 2022-06-29
Is our calculation of SCOP correct for hybrid heat pumps?


marliekeverweij:
[Here](https://github.com/quintel/etengine/blob/master/app/models/qernel/node_api/capacity_production.rb#L102-L119) you can see our calculation for the yearly average COP (SCOP):

![Screenshot 2022-06-29 at 14 54 17](https://user-images.githubusercontent.com/32833996/176441335-95828342-5fe1-4f6d-a360-7855477842d2.png)

This calculation works for all-electric heat pumps, because then you do:
heat_output_share (= `1`) / electricity_input_share (= `non_ambient_share`)

For hybrid heat pumps, the `non_ambient_share` consists of both electricity and gas/hydrogen, right @antw ? If that's the case, then the calculation should be adjusted I think.

I started looking into this calculation, because I figured that the SCOP of hybrid and gas hybrid heat pumps are different and expected them to be the same. For buildings we are adding a hydrogen HHP and then the SCOP is 4.0. The gas HHP has a SCOP of 5.5 in buildings.

I am not sure who to assign, but added @DorinevanderVlies because for now! Feel free to reassign

Notifying @quintel/core

DorinevanderVlies:
> because I figured that the SCOP of hybrid and gas hybrid heat pumps are different and expected them to be the same

I agree, if the HP part of gas and hydrogen HHP have the same efficiency and cut-of COP I would indeed expect the SCOP of both to be the same.

@marliekeverweij do you know the definition of  SCOP of HHP's? I see two options:
1. SCOP of HHP is about the whole HHP. So all non ambient energy in / heat out. 
2. SCOP of HHP is about the HP part. So electricity in / heat out for HP part.

SCOP of option 1 will be lower since the efficiency of the gas part of the HHP is lower than efficiency of the HP/electric part.

Equations:
1. `heat_output_share  / (electricity_input_share + gas_input_share + hydrogen_input_share)`
2. `heat_output_share (when electricity is used) / electricity_input_share`

I am not sure if it is possible/easy to determine `heat_output_share (when electricity is used)`

mabijkerk:
@Charlottevm and I also ran in some problems trying to trace back the COP of hybrid heat pumps.

To be clear, the COP of hybrid heat pumps is calculated using fever. In the snippet @marliekeverweij shared you can see the following condition:

```ruby
if fever&.efficiency_based_on && fever&.efficiency_balanced_with
[...]

PR #3219, opened 2025-02-21
Adding queries for manual calculation of COP of HHP

kaskranenburgQ:
This PR adds queries to calculate the COP of Hybrid heat pumps "manually" via the yearly input flows of ambient heat and electricity. The underlying calculation formula is:

(S)COP = (ambient_heat_input_conversion + electricity_input_conversion)/ electricity_input_conversion. 

With the idea that we calculate the units of produced heat compared to the units of electricity necessary. 

The engine method for the calculation of COP's "coefficient_of_performance" which is used in the queries in the folder general/heat/cop  does not work correctly for hybrid heat pumps since the efficiency of the boiler component is not taken into account in it's calculation. 
For more information on this issue see https://github.com/quintel/etsource/issues/2706#issuecomment-2668508000.


kndehaan (approved):
Looking good! I restructured the queries a bit and added an IF statement to avoid COP becoming 0 is electricity input conversion is 0 (is also done within the coefficient_of_performance method).

PR #3222, opened 2025-03-03
Adding queries for manual calculation of COP of HHP 

kaskranenburgQ:
This PR adds queries to calculate the COP of Hybrid heat pumps "manually" via the yearly input flows of ambient heat and electricity. The underlying calculation formula is:

(S)COP = (ambient_heat_input_conversion + electricity_input_conversion)/ electricity_input_conversion. 

With the idea that we calculate the units of produced heat compared to the units of electricity necessary. 

The engine method for the calculation of COP's "coefficient_of_performance" which is used in the queries in the folder general/heat/cop  does not work correctly for hybrid heat pumps since the efficiency of the boiler component is not taken into account in it's calculation. 
For more information on this issue see https://github.com/quintel/etsource/issues/2706#issuecomment-2668508000.

### DIFF
--- a/gqueries/general/heat/cop_manual_calculation/buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity_cop_manual_calculation.gql
+++ b/gqueries/general/heat/cop_manual_calculation/buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity_cop_manual_calculation.gql
@@ -1,0 +1,13 @@
+- query =
+    IF(
+      V(buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity, electricity_input_conversion) == 0.0,
+      1.0,
+      DIVIDE(
+        V(
+          buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity,
+          "ambient_heat_input_conversion + electricity_input_conversion"
+        ),
+        V(buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity, electricity_input_conversion)
+      )
+    )
+- unit = COP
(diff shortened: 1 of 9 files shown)


gqueries/general/heat/cop_manual_calculation/buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity_cop_manual_calculation.gql
```diff
@@ -0,0 +1,13 @@
+- query =
+    IF(
+      V(buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity, electricity_input_conversion) == 0.0,
+      1.0,
```
mabijkerk, line 4: Why is this line necessary?

mabijkerk:
It would be good to add a description to these queries, explaining why this manual calculation is necessary.

gqueries/general/heat/cop_manual_calculation/buildings_space_heater_hybrid_crude_oil_heatpump_air_water_electricity_cop_manual_calculation.gql
kaskranenburgQ: If this line is not added the query would return 0.0, which would seem low. 
@kndehaan Would you like to expand on this?

mabijkerk: Ah I see that the `coefficient_of_perfomance` method also returns 0 when both ambient and electricity input conversion is 0%. Fine to leave it as is.

mabijkerk (approved):

PR #3223, opened 2025-03-05
Custom curve flhs hydro river, solar thermal & geothermal

### LINKED ISSUE #3119: Solar thermal and geothermal custom curve full load hours don't work
When uploading a custom curve for solar thermal and geothermal full load hours, full load hours of the relevant nodes are not updated. 

For solar thermal, part of the issue is that `energy_heat_solar_thermal` in the `flh_of_solar_thermal` input is outdated. These nodes are now subdivided in `energy_heat_solar_ht_solar_thermal`, `energy_heat_solar_mt_solar_thermal`, `energy_heat_solar_lt_solar_thermal`.

```
- query = UPDATE(V(energy_heat_solar_thermal, buildings_space_heater_solar_thermal, households_water_heater_solar_thermal), full_load_hours, USER_INPUT())
- priority = 3
- unit = hours
- update_period = future
```

Similarly, for `flh_of_geothermal_heat` the node `energy_heat_well_geothermal` is outdated. We currently have the nodes: `energy_heat_well_deep_ht_geothermal`, `energy_heat_well_deep_mt_geothermal`, `energy_heat_well_shallow_lt_geothermal` (and `energy_heat_well_shallow_heatpump_mt_geothermal`?) to be updated here. 

However, when I implement the correct node names, the full load hours still don't update. So there is another problem here which should be divided in further.

kaskranenburgQ:
This PR aims to close #3119. 

It does so by updating the flh inputs for geothermal, solar thermal and hydro technologies. 
Please note that the produced energy of each of the technologies in the input statements will change with the flh's expect for **buildings_space_heater_solar_thermal** & **households_water_heater_solar_thermal**, since the energy production of these technologies is set via share inputs and not via a combintion of capaicty and flh's. 

Goes with:
- https://github.com/quintel/etengine/pull/1537
- https://github.com/quintel/etmodel/pull/4496
- https://github.com/quintel/etsource/pull/3223


mabijkerk:
@kaskranenburgQ will this change model results? If so, we cannot merge it without discussing it with our clients.

kndehaan:
Just did a quick scan of the changes. In addition to Mathijs' question and remark, I think we also need to write a migration, since the inputs that will be deleted with this PR can be set in scenarios... Let's discuss in the sprint.

kaskranenburgQ:
In the modeler sprint we discussed whether a scenario migration was necessary for the changes that were made in these inputs. We concluded that an investigation about the current working of the input statements was necessary: Do these inputs work when you set them in the backend?  And do they work when we upload a curve?

I found that the input statement does work. If  set:
`flh_of_geothermal_heat: 1000` in a blank nl2019 scenario on master, I see that the full load hours of the node **agriculture_geothermal** will be changed to 1000.0:
Screenshot 2025-03-11 at 11 40 57


However, when I then upload a capacity profile via the front-end:

Screenshot 2025-03-11 at 11 45 25

The flh setting is removed from the inputs:
Screenshot 2025-03-11 at 11 46 02

And the full load hours of agriculture_geothermal is reverted back to it's orginal value:
Screenshot 2025-03-11 at 11 46 22


To conclude: 
- The flh inputs work as a back-end input
- Uploading a curve is not functional and resets the input when it is set. 
- Therefore there is a great chance that users who 'accidentally' have set the input have overwritten the input with a curve. 
- We should check if this is the case.

kndehaan:
The user curve for hydro river works as it should! 

There's an issue in the geothermal and solar thermal user curve in scenarios where there's already installed capacity of related technologies in the start year. For example, take the a blank EU27 scenario and upload custom curves for solar thermal and geothermal. This correctly changes the full load hours, but incorrectly updates the number of units and incorrectly doesn't affect the production output, see example below for two technologies. 

image

I think this is because the flh_..._user_curve input should include a correction where the original NoU is set and preset demand is updated based `production_based_on_number_of_units`. In the electricity production technologies, this is done with `preset_demand_by_electricity_production`. For steam hot water production, this method doesn't exist. 

Is there another way to tackle this with the methods we already have? Or do we need to create this method for steam hot water as well? @kaskranenburgQ @mabijkerk

kaskranenburgQ:
Based on the PR of @louispt1 (https://github.com/quintel/etengine/pull/1537), I have added the preset_demand_by_steam_hot_water_production update in the user curve inputs. While testing I found the following points:

1. I have implemented the input statement based on 'production_based_on_number_of_units'. This does not work since this function in turn is based on the typical_electricity_production_per_unit. For hydrogen producers (i.e. for energy_hydrogen_electrolysis_solar_electricity in the input flh_of_solar_pv_solar_radiation) we use hydrogen_production_based_on_number_of_units. In order to be able to update the preset_demand_by_steam_hot_water_production correctly, I propose we also add a function 'steam_hot_water_production_based_on_number_of_untis. 
2. While uploading a curve for solar thermal a huge flow from energy to heat occurs, see picture below:
Energy flow overview 1289457 (1), it could be that this occurs due to the issue described in point 1. We should check whether this still occurs when the fix for point is implemented.

@louispt1 Could you add this in your PR? Then I will change the inputs to accomodate this change.

kaskranenburgQ:
After a rebase of the engine branch the issue has been solved. @kndehaan Could you do one final review?

inputs/full_load_hours/flh_of_geothermal_heat_user_curve.ad
```diff
@@ -0,0 +1,42 @@
...
+
+      UPDATE(V(energy_heat_well_shallow_heatpump_mt_geothermal), full_load_hours, USER_INPUT()),
+      UPDATE(V(energy_heat_well_shallow_heatpump_mt_geothermal), number_of_units, original_nou_energy_heat_well_shallow_heatpump_mt_geothermal),
+      UPDATE(V(energy_heat_well_shallow_heatpump_mt_geothermal), preset_demand_by_steam_hot_water_production, V(energy_heat_well_shallow_heatpump_mt_geothermal, steam_hot_water_production_based_on_number_of_units)),
+
+      UPDATE(V(industry_heat_well_geothermal), full_load_hours, USER_INPUT()),
+      UPDATE(V(industry_heat_well_geothermal), number_of_units, original_nou_industry_heat_well_geothermal),
+      UPDATE(V(industry_heat_well_geothermal), preset_demand_by_steam_hot_water_production, V(industry_heat_well_geothermal, steam_hot_water_production_based_on_number_of_units)),
+
+      UPDATE(V(agriculture_geothermal), full_load_hours, USER_INPUT())
```
kndehaan, line 33: Was deliberately chosen to not apply the same method of setting the FLH for `agriculture_geothermal`?

inputs/full_load_hours/flh_of_solar_thermal_user_curve.ad
```diff
@@ -0,0 +1,33 @@
...
+      UPDATE(V(energy_heat_solar_lt_solar_thermal), full_load_hours, USER_INPUT()),
+      UPDATE(V(energy_heat_solar_lt_solar_thermal), number_of_units, original_nou_energy_heat_solar_lt_solar_thermal),
+      UPDATE(V(energy_heat_solar_lt_solar_thermal), preset_demand_by_steam_hot_water_production, V(energy_heat_solar_lt_solar_thermal, steam_hot_water_production_based_on_number_of_units)),
+
+      UPDATE(V(energy_heat_solar_mt_solar_thermal), full_load_hours, USER_INPUT()),
+      UPDATE(V(energy_heat_solar_mt_solar_thermal), number_of_units, original_nou_energy_heat_solar_mt_solar_thermal),
+      UPDATE(V(energy_heat_solar_mt_solar_thermal), preset_demand_by_steam_hot_water_production, V(energy_heat_solar_mt_solar_thermal, steam_hot_water_production_based_on_number_of_units)),
+
+      UPDATE(V(buildings_space_heater_solar_thermal), full_load_hours, USER_INPUT()),
+      UPDATE(V(households_water_heater_solar_thermal), full_load_hours, USER_INPUT()),
```
kndehaan, line 24: Same question for `buildings_space_heater_solar_thermal` and `households_water_heater_solar_thermal`, not necessary here to apply the same method?

kndehaan:
An uploaded curve for `flh_of_geothermal_heat_user_curve` will set the full load hours of all geothermal nodes to the same full load hours. However, the default full load hours vary per technology (not dataset-dependent):

energy_heat_well_shallow_lt_geothermal: 4750
energy_heat_well_deep_mt_geothermal: 6000
energy_heat_well_shallow_heatpump_mt_geothermal: 4750
energy_heat_well_deep_ht_geothermal: 6000
industry_heat_well_geothermal: 7000
agriculture_geothermal: 6000

So, wouldn't it be undesirable that with one curve, all technologies' full load hours are updated to the same value?

kndehaan:
The custom curve upload works well now for all technologies that have steam_hot_water output, thus the new methods `preset_demand_by_steam_hot_water_production` and `steam_hot_water_production_based_on_number_of_units` work as expected. 

However, the same problem (not updating production but incorrectly updating nr of units as mentioned [here](https://github.com/quintel/etsource/pull/3223#issuecomment-2724373131)) still exists for all technologies with useable_heat output. More specifically, for:

```
agriculture_geothermal
buildings_space_heater_solar_thermal
households_water_heater_solar_thermal
```

I tried to apply the same method here as is done for the other technologies (including copying the method in ETEngine as done in https://github.com/quintel/etengine/pull/1537 for useable_heat). For it to work, it requires to add `preset_demand` to the node groups. However, I'm not so certain whether these nodes represent preset demand. 

So question is, are these nodes preset demand? If so, we could add `preset_demand` to the node groups and create and apply the same method for useable_heat as well.

kaskranenburgQ:
The following to-do's are still open before we can merge this:

- [ ] Split flh_of_geothermal_user_curve into inputs for shallow & deep geothermal.
- [ ] Write migration to remove outdated inputs from scenario's --> With the comment of @louispt1 in mind: https://github.com/quintel/etsource/pull/3223#issuecomment-2718367704, this might not be explicitly necessary.
- [ ] Change documentation in files flh_of_solar_thermal_user_curve & flh_of_geothermal_user_curve, so that it's clear why agriculture_geothermal, buildings_space_heater_solar_thermal, and households_water_heater_solar_thermal have a different update structure. 
- [ ] Check of number of units of agriculture_geothermal, buildings_space_heater_solar_thermal, and households_water_heater_solar_thermal are updated correctly in the current structure

louispt1:
I think it would be best practice to still include a migration to remove the outdated inputs - I only checked on a local pro db

kndehaan (approved):
Nice work!

PR #3235, opened 2025-03-18
Apply node group approach to final demand per sector per carrier queries for biofuels and oils

kndehaan:
This PR applies the `final_demand_group` node group to the the per sector per carrier queries used in Collection for biofuels and oils and derivatives. Note that this is a quick fix to align the total final demand of these carriers in the per sector per carrier queries and per carrier queries, as mentioned here https://github.com/quintel/multi-year-charts/issues/83#issuecomment-2709869486


kaskranenburgQ (approved):
LGTM!

PR #3245, opened 2025-04-03
Move industry final demand edge group to include DSR demand

mabijkerk:
Before, only one of the edges from buildings local production to the ICT sector was marked with an edge group. The edge to DSR was empty. To solve this, the edge group is moved 'deeper' into the graph, where the DSR edge recombines with the regular demand edge. This way, all ICT demand is include in the final demand edge group.

In scenarios with DSR active in ICT, this query `final_demand_of_electricity_in_ict_industry` should now match the annual demand for ICT in the graph.


kaskranenburgQ (approved):
LGTM!

PR #3255, opened 2025-04-30
Merge household appliances efficiency sliders into one

mabijkerk:
This PR adresses https://github.com/quintel/etmodel/issues/4479. There are currently 8 household appliances efficiency sliders. This seems excessive. They are merged into one.

Goes with: 
- https://github.com/quintel/etmodel/pull/4481


kndehaan:
Yup, the inputs both still work when they're both set.

PR #3266, opened 2025-06-02
Rake task to remove a specified attribute from nodes

### LINKED ISSUE #3097: Write rake task to remove outdated attributes from node files
There are a number of attributes that are outdated. See https://github.com/quintel/documentation/issues/158, but to also give examples here:

- hours_maint_nl
- hours_place_nl
- hours_prep_nl
- ...

Currently, we simply delete them whenever we encounter them. It would be more efficient however to build a rake task and remove them in one go from all node files, whenever we retire an attribute. Note: we would then preferably retire the attributes completely. Not only from node files, but also the relevant queries and the attribute itself on [Atlas](https://github.com/quintel/atlas/blob/master/lib/atlas/energy_node.rb#L52-L58).

louispt1:
As discussed in #3097, this Rake task walks through the files in graphs/energy/nodes and removes any line with the specified attribute. It should be clear in the git diff which changes have been made.

For example, from testing:
image

Closes #3097


mabijkerk (approved):
Works like a charm, very very nice! Could you maybe check if we use `energy_balance_group` anywhere? If not, we can also remove it.

louispt1:
> Works like a charm, very very nice! Could you maybe check if we use `energy_balance_group` anywhere? If not, we can also remove it.

We only use it in ETSource and Atlas, so I've removed it.

PR #3270, opened 2025-06-11
Ruby upgrade to 3.3.5

louispt1:
Simpler than Local and Model, only had to update the load spec to use BigDecimal rather than floats to get the precision checks to pass.


noracato:
Cherrypicked the BigDecimal commit to master.

Let's keep the version on 3.2, to be the exact same as engine!

PR #3278, opened 2025-07-08
Add queries for electricity input of cooling technologies

mabijkerk:
This PR adds missing input of electricity queries for cooling technologies in the built environment. Should be cherry-picked to tyndp-2026 after merging.


kaskranenburgQ (approved):
LGTM!

PR #3284, opened 2025-07-18
Align analysis manager between ETSource and ETDataset

mabijkerk:
When importing `nl2019` in ETSource, there were some diffs in the dataset. This should not be the case, as ETDataset and ETSource should be fully aligned for the analysis manager pipeline and the `nl2019` dataset. This PR tackles the diffs for various sectors and re-aligns the two repositories.

Goes with https://github.com/quintel/etdataset/pull/1042


kndehaan (approved):
LGTM

PR #3294, opened 2025-08-25
Municipal NL dataset update to 2023

kaskranenburgQ:
This PR updates the muncipal datasets to 2023. 
It also adds new datasets for the RES-regions and provinces which are created using the municipal datasets and the amalgamator. It also removes the 2019 datasets for the RES regions (starting with 'RES'), merged municipalities and PV21_friesland (because it is now called fryslan). In other words, there should be no more municipal/provincial/RES datasets for 2019.

Goes together with:
https://github.com/quintel/etlocal/pull/618
https://github.com/quintel/etdataset/pull/1049
https://github.com/quintel/etmodel/pull/4549

Points of attention for the review:

1. Do you see changes for the datasets that you expect?
2. Do you see any noticeable / unexpected changes within the datasets?
3. Can you verify that all relevant 2019 datasets have been removed?
4. Can you open a local blank scenario for one of the new/updated datasets?
5. Can you check where the merge conflicts for non-municipal datasets come from?


mabijkerk:
@KoenvanB the branch is out of date with master, leading to some merge conflicts. It needs to be rebased.

datasets/ES01_groningen/ES01_groningen.derived.ad
mabijkerk, line 1: If we're renaming the GEO IDs of these datasets, then we should also have a migration that changes this for existing scenarios right? I don't see a linked ETEngine PR. Is a migration there?

mabijkerk: Additionally, what do we do with the old region ID datasets on ETLocal?

KoenvanB: Rename on ETEngine; retire on ETLocal.

kndehaan (commented):
Generally, thing are looking well! Regarding the questions posed:

1. Do you see changes for the datasets that you expect?
The intended changes to the datasets seem fine to me

2. Do you see any noticeable / unexpected changes within the datasets?
See my some of my comments regarding unexpected changes. The most important is that the `area_per_building_residence_equivalent` attribute should be preserved for the ETLocal datasets, otherwise slider that sets building numbers will break the model. 

3. Can you verify that all relevant 2019 datasets have been removed?
Yes!

4. Can you open a local blank scenario for one of the new/updated datasets?
Yes, I tried it for municipal, RES and provincial datasets and I just opened the energy flow overview chart. This works fine and seems well. I didn't thoroughly tested sliders etc. 

5. Can you check where the merge conflicts for non-municipal datasets come from?
In the meantime, I think the branch is rebased and there are no merge conflicts anymore? However, as mentioned one of my comments, it is needed to rebase the ETLocal branch and re-export the datasets.

kndehaan:
I was just thinking again about the `area_per_building_residence_equivalent` attribute: it's not connected to ETLocal. The parent datasets should have this attribute, the child datasets should inherit this from the parent datasets. Since nl2023 does have this attribute, I don't understand why the municipal datasets haven't inherited this. Similar to the international datasets, they should inherit this from the eu dataset, which does have this attribute.

So I don't know what goes wrong here... @KoenvanB do you have any ideas?

KoenvanB:
> @KoenvanB the branch is out of date with master, leading to some merge conflicts. It needs to be rebased. Perhaps that will also remove all the changes for the European datasets that should not be there.

Rebase is done. No more merge conflicts for European datasets; it involved only the id anyway.

kaskranenburgQ:
@KoenvanB It's looking good. I only had to fix the RS_serbia graph values. 
One question: The value for the area_per_building_residence_equivalent for European datasets is 293.627703981492, while we use 130.0 for all regional datasets at the moment. 
The Dutch national dataset also assumes that the value is 293.627703981492. 
Should we take over this value?

PR #3300, opened 2025-09-09
Alter EB method for coke oven cokes in steel sector

kaskranenburgQ:
This PR extracts the datasets from the state of etlocal after the fix of https://github.com/quintel/etlocal/pull/614. 
Goes together with:
https://github.com/quintel/etlocal/pull/614

Assigned to @mabijkerk for now, reassign at will


mabijkerk:
For Romania, the blast furnace is slightly off. This is because not all the flows for coalg as on the EB are available in the graph. Coal gas can go to energy own use of electricity and heat generation on the EB, and to distribution losses. Neither are incorporated in the graph yet.

There is no simple fix for this, but we should note it down somewhere.

image

mabijkerk (approved):
Conclusion: when coal gas only goes to electricity and heat generation in the transformation flows, or to final consumption of iron and steel or to energy own use of blast furnaces and cokes ovens, then the mapping will work. This is the case for countries like Italy and the Netherlands. When there are other flows, like distribution losses, or final consumption non-metallic minerals etc. then it does not work.

This new mapping has been verified with IT_italy and works perfectly!

PR #3325, opened 2025-10-14
Fix bugs signaled by mechanical turk

kaskranenburgQ:
This PR fixes bugs signaled by the mechanical turk from the carbon modelling project. Goes together with:
https://github.com/quintel/mechanical_turk/pull/200


kndehaan (approved):
Looking good.

PR #3334, opened 2025-10-20
Carbon modelling to-jet technologies 

kndehaan:
This PR adds the following processes to the model as part of the carbon modelling project:
- methanol-to-jet
- biomethanol-to-jet
- bio-ethanol-to-jet

The PR includes the following changes:
- NSAs for the new processes
- Energy graph updates
- Inputs
- Locales
- Query updates

Goes with:
- https://github.com/quintel/etsource/pull/3334
- https://github.com/quintel/etdataset/pull/1059
- https://github.com/quintel/etmodel/pull/4573


config/sankey_csv.yml
mabijkerk: @kndehaan what I don't really get in the Sankey download is that the labels now don't seem to fully align with the new processes. For example, "Synthetic kerosene production" is now a bit outdated right?

mabijkerk:
This is probably related to the municipal datasets, so maybe @kaskranenburgQ can answer this. Why is there a refinery in Schiermonnikoog? Does that make sense? It was not there in the 2019 version.

image

kndehaan:
The Schiermonnikoog issue has been solved. The mismatch was related to double counting of losses from green gas shares in network gas. Since network gas is a separate carrier in the Sankey chart, a different query has to be used here compared to the Sankey download (where network gas is not a separate carrier.

mabijkerk (approved):

PR #3346, opened 2025-11-07
updated the biomass potentials

Max-Kerpel:
## Description

Update the biomass potentials for all municipalities and amalgamated to the regions and provinces.

## Type of change

- [ ] Bug fix

## Checklist


kndehaan (approved):
Functional review: I found duplicate attributes in the area files of PV20 and PV27_2022 (I think they were already there before the work for this PR), and removed these. And I think the biomass potentials of PV20 were not updated yet, so did this as well. Please double check if these values are correct.

Issue #3347, opened 2025-11-07
Weater_properties.csv files GM0060_ameland are missing flh_of_energy_power_solar_pv_offshore input


kndehaan:
## Bug Description

While testing the scenario updater changes from https://github.com/quintel/etengine/pull/1657, I discovered that setting a weather year for GM0060_ameland leads an error message saying `flh_of_energy_power_solar_pv_offshore` is not defined in the weater_properties.csv files of the weather years folders. 

<img width="1336" height="610" alt="Image" src="https://github.com/user-attachments/assets/bb55202e-85c3-4ee6-bfa3-04978bff0fcd" />

This missing attribute is required for setting weather years. 

---

### Steps to Reproduce

The error message was shown on ETEngine branch `scenario-updater`. Trying to set a weather year on beta gives a different error, but is probably about the same:

<img width="943" height="620" alt="Image" src="https://github.com/user-attachments/assets/6152d834-334c-44d1-bd55-7b3c84b5e1db" />

On pro, you can set the weather year without getting an error. The results just don't change. 

---

### Proposed solutions

- The flh_of_energy_power_solar_pv_offshore input should be added to all weather_properties.csv files of Ameland
- We should check if all other weather_properties.csv files from other datasets include all required inputs

mabijkerk:
It apparently also applies to all municipalities in the province of Utrecht, which were syminked to 2018 in https://github.com/quintel/etsource/commit/1f64538322c9210a7cff6ee40cc08f957b9719e3.

mabijkerk:
Two questions to @kndehaan, 1) how is it that this only comes up as an error now, we did not merge anything yet from Solar PV and 2) why did this not pop up in Semaphore?

PR #3348, opened 2025-11-10
Remove outdated curve from datasets

mabijkerk:
This PR closes https://github.com/quintel/etsource/issues/3347. It makes the following changes:

- Remove curves folder for Ameland, these were originally adapted to show seasonal variation but they are now outdated. It is better if they just revert to `nl2023`.
- Remove symlinked curves for Utrecht municipalities and the province. This was wrongly stil symlinked to `nl2018`, but will now revert to `nl2023`.


mabijkerk:
Nice catch @kndehaan! I removed the curves for those datasets.

kndehaan (approved):

Issue #3382, opened 2025-12-12
Unused heat should be added as loss in Sankey chart queries


kndehaan:
## Bug Description

In #2931 we did changes so that unused heat is accounted for as a loss instead of final demand. The flow from heat to loss currently misses in the Sankey chart (also see set sliders for reproducing):

<img width="1194" height="593" alt="Image" src="https://github.com/user-attachments/assets/71798aac-70d8-4d49-914d-26389b6568e2" />

--

## Proposed solutions

- [ ] the loss flow from unused heat to loss should be added in the Sankey chart
- [ ] the Sankey download should be checked if this issue occurs there as well

mabijkerk:
Before, the unused heat was part of final demand, so e.g., in the query `sankey_1_to_2_heat_to_energy` it would register in the undelrying query `final_demand_of_heat_in_energy_energetic`. Now we need to add it to the `sankey_1_to_2_heat_to_loss` query.

**Heat networks**
This "loss" occurs on the balancing nodes and is therefore a steam hot water to steam hot water balancing loss. Within the Sankey query `sankey_1_to_2_heat_to_loss` it should therefore be added to the query `conversion_loss_steam_hot_water_to_steam_hot_water`:

```
    DIVIDE(
      SUM(
        V(
          FILTER(G(sankey_conversion_group),"steam_hot_water_output_conversion > 0.0 && steam_hot_water_input_conversion > 0.0"),
          "steam_hot_water_output_conversion * steam_hot_water_input_conversion / (1 - loss_output_conversion) * output_of_loss"
        )
      ), 
      BILLIONS
    )
```
There are two options. The _first_ option is to incorporate an additional condition into the existing query structure, making sure edges marked as loss are also included and corrected for. This will make the query quite intricate. The _second_ option is to simply SUM the total of all steam hot water edges that are marked as loss to the existing query structure.

**Conclusion**
@kndehaan @kaskranenburgQ which of the two options do you prefer?

Whichever we choose, we should also incorporate something similar for residual heat in the electricity to hydrogen conversion for the P2G node.

kndehaan:
I think the second option would be best, but we should find a way how we can query this robustly. It would be nice to use the `FILTER()` function, but that works with groups, e.g. `FITER(EG(...), "treat_as_loss == true")` if this works(?) So should we add edge groups? Or is there a better method without adding new attributes. 

We could also just do hard-coded nodes or edges, but that is maybe prone to get outdated quickly...

[...]

PR #3384, opened 2025-12-18
Add unused heat to loss in Sankey

mabijkerk:
This closes https://github.com/quintel/etsource/issues/3382.

It adds unused heat to the Sankey (chart and download, since they use the same underlying query). Before, the loss was part of the final demand of specific sectors. Now that it is no longer part of final demand, we have to filter on `treated_as_loss` and add it to the steam hot water to steam hot water conversion losses. We also have to filter on steam hot water as a carrier to prevent residual heat from P2G to be taken into account.


kndehaan (approved):
LGTM!

PR #3385, opened 2025-12-18
Queries for final demand passenger transport road and rail per carrier

### LINKED ISSUE #3380: Add queries for final demand passenger transport so split out non-biogenic and biogenic carriers per modality.
In the transport sector there are several queries for modalities that split out the demand per carrier. 

These queries do not yet exist for personal transport modalities (e.g. cars).

We should add these.

kndehaan:
## Description

This query adds new queries for final demand of passenger transport road and rail per carrier. Note:
- Queries are only added for road and rail. Per carrier queries for domestic aviation already exist. 
- The existing queries for domestic navigation have been moved to a new folder. It would be good to inform the client on this since we provided the Github link to the existing queries earlier on, which now have been changed

Discussion point: the new queries now have MJ as unit, as this is the case for most of the existing transport queries, but not for all however. Should we leave the unit as MJ or change it to PJ?

A Nice-to-have to pick up would be to create queries for the other transport modi in the general folder as well, and link the queries that are in specific output element folders to the queries in the general folder. We could also apply consistent units as well then.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Documentation

## Checklist

## Related Issues

Closes #3380


kaskranenburgQ:
I have reviewed the queries and it's looking nice! 
SInce we still have enough time on this subproject, I feel that we can pick-up the NTH point as well.

kndehaan:
I noted the nice-to-have for adding more transport queries in the general folder in this separate [issue](https://github.com/quintel/etsource/issues/3388).

PR #3383, opened 2025-12-18
Oil share queries

kaskranenburgQ:
### Share Queries (10 new files)

- Added biogenic share calculations for agriculture, buildings, households, and industry sectors (energetic use)
- Added biogenic share calculation for industry non-energetic use
- Added complementary non-biogenic share queries (calculated as 1 - biogenic share) for all sectors

The biogenic share queries calculate the proportion of biodiesel, bio-kerosene, and bio-oil input conversions in crude oil final demand nodes. 

### Final Demand Queries (11 new files)

- Added biogenic oil final demand queries for agriculture, buildings, households, and industry (energetic and non-energetic)
- Added non-biogenic oil final demand queries for the same sectors
- Added industry non-energetic oil and derivatives final demand query

Industry-Specific Implementation: The industry sector queries use a different approach than other sectors:
Other sectors multiply total oil demand by the biogenic/non-biogenic share

- Industry queries separate naphtha (which is modeled independently as naphtha vs bionaphtha) from other oil products:
  - For biogenic: crude_oil demand × biogenic_share + full bionaphtha demand
  - For non-biogenic: crude_oil demand × non-biogenic_share + full naphtha demand

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Documentation

## Checklist

- [ ] I have tested these changes
- [ ] I have updated documentation as needed
- [ ] I have tagged the relevant people for review

## Related Issues

Closes #3377

PR #3393, opened 2025-12-29
Update crude oil final demand queries to reflect stand alone queries

kaskranenburgQ:
## Description

This PR updates the queries added  in #3383 by creating queries that can be directly used to determine the actual final demand of biogenic oil carriers and non-biogenic oil carriers in the sectors. 

## Type of change

- [ ] Enhancement

## Checklist

- [ ] I have tested these changes

## Related Issues

Closes #


kndehaan (approved):
I applied a fix in the biogenic queries. I think it works as it should now but it would be good if you could do a final check @kaskranenburgQ

PR #3414, opened 2026-01-22
Add prefix turk_ queries for pimary demand spec

mabijkerk:
By adding the prefix, the queries can be cached and the spec duration in Mechanical Turk can be significantly reduced.

This is possible since the efficiency restructure of Mechanical Turk in https://github.com/quintel/mechanical_turk/pull/213. Goes with https://github.com/quintel/mechanical_turk/pull/215.


kndehaan (commented):
I think all the queries that are now referenced in the newly added queries are only used in the Mechanical Turk spec. It would therefore make more sense to just rename the existing queries to include prefix 'turk_' instead of adding new queries. 

While at it, it would be more neat to move the existing `turk_queries_primary_demand_per_sector.gql` inside the primary demand folder.

PR #3441, opened 2026-05-04
Add dataset value molecule nodes with emissions data

kndehaan:
#### Context
Adds molecule nodes that contains dataset value-based emissions data. Sector specific emissions node groups have been added to the molecule nodes.

## Checklist

### DIFF
--- a/config/molecules_node_positions.yml
+++ b/config/molecules_node_positions.yml
@@ -1,10 +1,19 @@
 ---
+agriculture_non-specified_non_energetic_other_ghg:
+  x: 1920
+  "y": 5140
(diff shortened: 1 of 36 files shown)


louispt1 (approved):
LGTM - I was expecting `use` as well but I suppose it isn't necessary?

louispt1 (changes requested):
Just realised the tests failed on semaphore - will check that out before approving

EDIT: The failing specs are datasets not calculating which is not an issue at this stage (the emissions.csv is not in every dataset yet)

louispt1 (approved):

PR #3444, opened 2026-05-29
Add direct emissions method with data export

kndehaan:
#### Context

This PR adds a new direct emissions method to the model, together with a detailed data export containing all greenhouse gas emissions of a scenario. 

#### Implemented changes
- New engine methods and node attributes that calculate relevant direct emissions, based on presence of (molecule) node group `emissions`
- Data processing pipeline that generates start year emissions data for country datasets, based on reported emission data from UNFCCC. Start year data is added as `emissions.csv` file to ETSource using a rake task. 
- New molecule nodes that read static emission data from `emissions.csv`, specified for co2 and other_ghg
- Detailed data export containing relevant direct reporting emission results for all relevant nodes, and categorises these into ETM sector and subsector. 

#### Related

Goes with pull requests:
- https://github.com/quintel/etengine/pull/1754
- https://github.com/quintel/atlas/pull/189
- https://github.com/quintel/etsource/pull/3444
- https://github.com/quintel/etmodel/pull/4719
- https://github.com/quintel/etdataset/pull/1087
- https://github.com/quintel/etlocal/pull/695
- https://github.com/quintel/documentation/pull/321

## Checklist


carriers/ambient_cold.ad
```diff
@@ -2,4 +2,5 @@
 - infinite = true
 - graphviz_color = deepskyblue2
 
+~ co2_conversion_per_mj = 0.0
```
mabijkerk, line 5: These attributes are now specified twice: once in their carrier files and once in the dataset files. Perhaps we can create an issue for this to remove this redundancy?

config/molecules_node_positions.yml
mabijkerk, line 1: In the mocule graph, the energy sector nodes are now a bit scattered, this also applies to the industry sector nodes. In the energy graph we nicely group all nodes. Should we do this for the molecule graph nodes as well?

config/direct_emissions_csv.yml
```diff
@@ -0,0 +1,207 @@
...
+    value: direct_reporting_emissions_other_ghg_emissions
+    transform: "(value * 1e-6).round(6)"
+  - name: Total_GHG_emissions
+    label: "Total GHG emissions[kton CO2-eq]"
+    type: node_attribute
+    value: direct_reporting_emissions_total_ghg_emissions
+    transform: "(value * 1e-6).round(6)"
+
+
+rows:
```
mabijkerk, line 66: This data-exports seems to be alphabetically sorted on Group, does that makes sense? Isn't there given order that we can follow that matches more with UNFCC, EU, PBL...? Something like Energy, Industry, Transport, Households, Services, Agriculture...

graphs/molecules/nodes/agriculture/agriculture_non_specified_non_energetic_co2.ad
mabijkerk, line 1: Still not 100% convinced about these being called `non_energetic`. The client also brought this up: non-energetic/energetic is a distinction from the energy balance that does not always match with combustion/non-combustion emissions. Good to discuss together or check in with an expert?

config/direct_emissions_csv.yml
```diff
@@ -0,0 +1,207 @@
...
+    Subsector: Refineries
+  - Group: emissions_industry_steel
+    Sector: Industry
+    Subsector: Steel
+  - Group: emissions_international_transport_international_aviation
+    Sector: International transport
+    Subsector: International aviation
+  - Group: emissions_international_transport_international_navigation
+    Sector: International transport
+    Subsector: International navigation
```
mabijkerk: These other GHG values need to be dependent on the sliders settings `bunkers_allocated_percentage_aviation` and `bunkers_allocated_percentage_shipping`.

graphs/molecules/nodes/molecules/molecules_other_utilisation_delayed_emitted_co2.ad
```diff
@@ -1 +1,3 @@
+- groups = [emissions, emissions_energy, emissions_energy_ccus]
```
mabijkerk, line 1: This is definitely one to address up in the next project. With the new direct emissions method it is better allocated than it is in the primary emissions method (where it is a separate category). Still, it is even better to reallocate this to the relevant sectors: Food (carbonated drinks, decaf) and Agriculture (horticulture direct CO2 use and agriculture CO2 release from fertilizer use).

tasks/etdataset/import_emissions.rb
```diff
@@ -0,0 +1,123 @@
...
+
+    Optional ENV parameters:
+      DATASET   - Import only this dataset (country code)
+      YEAR      - Use this year (otherwise looks up in datasets.yml)
+      YEAR_NAME - If set, appends year to dataset name (e.g., 'nl2019' not 'nl')
+
+    Examples:
+      rake import:emissions                              # Auto-discover and import all
+      rake import:emissions DATASET=nl YEAR=2019         # Import nl/2019 → datasets/nl/
+      rake import:emissions DATASET=nl YEAR=2019 YEAR_NAME=1  # Import nl/2019 → datasets/nl2019/
```
mabijkerk: In the import_dataset.rb task we use the following:
```
Use YEAR_NAME=true to import dataset with name 'de2011' instead of 'de'

rake import:dataset DATASET=nl YEAR=2023 YEAR_NAME=true
```

Might be good to match that here as well, instead of `YEAR_NAME=1`.

graphs/molecules/nodes/bunkers/bunkers_international_aviation_energetic_other_ghg.ad
```diff
@@ -0,0 +1,4 @@
+- groups = [preset_demand, emissions, emissions_international_transport, emissions_international_transport_international_aviation]
```
mabijkerk: Minor question: to stick closer to the energy graph, should we change this group to `emissions_bunkers`? We also have `emissions_households` and `emissions_buildings` instead of `emissions_residential` and `emissions_commercial_institutional`?

config/direct_emissions_csv.yml
mabijkerk: Note that this should apply both to present and future year!

kndehaan: @noracato the molecule nodes in these groups contain the other_ghg value that is read from the emissions.csv. But actually these values that end up in the user output should depend on the inputs mentioned by Mathijs. Is there a way that we can process this in the data export?

Note that this already goes correctly for the energy nodes within this group; the energy nodes will have a demand and calculated emissions if the inputs are set to > 0.

carriers/ambient_cold.ad
kndehaan: Done! https://github.com/quintel/etsource/issues/3445

config/molecules_node_positions.yml
kndehaan: It's a consideration between distinction per sector or distinction per modelling type. Currently, we placed the static modelling (molecule nodes that read from emissions.csv) at the bottom and dynamic modelling is at the top. If we group per sector, the static modelling of energy and industry sector will be right next to the dynamic modelling. 

Since the molecule nodes are somewhat 'strange' exceptions in the modelling, I don't find it too strange that they are at the bottom. But if we want to apply the graph design of grouping sectors together, then it makes sense to move some of the molecule nodes. 

I don't have a strong preference. @mabijkerk what about you?

config/direct_emissions_csv.yml
kndehaan: @noracato you can ignore my request above, it's a modeller's thing that I will look into.

config/direct_emissions_csv.yml
kndehaan: Let's follow follow IPCC:
Energy
Industry
National transport
Buildings
Households
Agriculture
Other
LULUCF
Waste
International transport

@robindesmit-quintel could you pick this up?

graphs/molecules/nodes/molecules/molecules_other_utilisation_delayed_emitted_co2.ad
kndehaan: I agree, it's listed under future improvements in a document in the project in Linear.

tasks/etdataset/import_emissions.rb
kndehaan: @louispt1 could you look into this?

graphs/molecules/nodes/bunkers/bunkers_international_aviation_energetic_other_ghg.ad
kndehaan: Good point. I'm fine with changing this. We also use `emissions_national_transport...`, so I will also change this to `emissions_transport_...`.

tasks/etdataset/import_emissions.rb
louispt1: Yes this is how it already works - I've updated the example to clarify with "true" (but this also works with `1`)

graphs/molecules/nodes/bunkers/bunkers_international_aviation_energetic_other_ghg.ad
kndehaan: Done

config/molecules_node_positions.yml
mabijkerk: I would prefer sector clustering. It does not have the highest priority but perhaps @robindesmit-quintel can pick this up?

graphs/molecules/nodes/bunkers/bunkers_international_aviation_energetic_other_ghg.ad
mabijkerk: Nice!

config/direct_emissions_csv.yml
kndehaan: Done!

graphs/molecules/nodes/agriculture/agriculture_non_specified_non_energetic_co2.ad
kndehaan: This is a broader discussion about using the energetic / non-energetic terminology for categorisation of emissions. The desire from our part is to get a better understanding of certain industries where people refer to combustion and process, vs. how this is translated to the energy balance as energetic and non-energetic. 

We discussed that if during the validation we encounter significant mismatches in modelled emissions vs. reported emissions, and can't explain these differences due to unclear separation of combustion/process vs.  energetic / non-energetic, we need to talk to experts to get a better understanding of this.

config/direct_emissions_csv.yml
kndehaan: I fixed this in https://github.com/quintel/etsource/pull/3444/changes/b5e2bbc8613e884a176af79d996dc0256e462b4c. @mabijkerk could you give this a final review?

config/molecules_node_positions.yml
kndehaan: Done!

graphs/energy/edges/agriculture/agriculture_final_demand_crude_oil-agriculture_burner_crude_oil@crude_oil.ad
```diff
@@ -1,3 +1,3 @@
 - type = share
 - reversed = false
-- groups = [final_demand]
\ No newline at end of file
+- groups = [final_demand, emissions_skip_crude_oil_mix]
```
mabijkerk, line 3: Ideally, would add a spec that validates the inclusion of this edge so that we don't forget it if we change the graph in the future: all downstream crude oil edges for final demand sectors that have a mix of bio and fossil coming in should be marked with the exception edge group

mabijkerk: I added this in a PR https://github.com/quintel/etsource/pull/3446.

config/direct_emissions_csv.yml
mabijkerk: This works! 🙌

mabijkerk:
@kndehaan what do we do with the emissions of `energy_methanol_production_coal_gas`?

graphs/energy/nodes/energy/energy_production_methanol_synthesis_dry_biomass_ccs.ad
```diff
@@ -18,7 +22,7 @@
 - variable_operation_and_maintenance_costs_per_full_load_hour = 2019.06
 - wacc = 0.07
 - free_co2_factor = 0.522288
```
mabijkerk, line 24: I think this `free_co2_factor` calculation still is not correct. I think we forget to discount for the carbon content in the output products. A simple example with 50 PJ of methanol production from non-biogenic waste shows the difference:

```
EACH(
  V(energy_production_methanol_synthesis_non_biogenic_waste_ccs,primary_co2_emission),
  V(energy_production_methanol_synthesis_non_biogenic_waste_ccs,direct_co2_output_production_emissions_fossil)
)

  3,971,791,508.1967206,
  516,795,081.96721315
```

We now calculate the free_co2_factor = carbon_captured / carbon_content_in, but it should be free_co2_factor = (carbon_captured + carbon_content_out) / carbon_content_in or free_co2_factor = 1 - carbon_emitted / carbon_content_in.

mabijkerk: This applies to the gasification for FT and methanol synthesis. This is only relevant for primary emissions and even then only for the emission costs, since the actual emissions do not get calculated for methanol, kerosene etc. through the primary route.

graphs/energy/edges/agriculture/agriculture_final_demand_crude_oil-agriculture_burner_crude_oil@crude_oil.ad
kndehaan: The proposed PR had been closed with [this comment](https://github.com/quintel/etsource/pull/3446#issuecomment-4667396488).

graphs/energy/nodes/energy/energy_production_methanol_synthesis_dry_biomass_ccs.ad
kndehaan: I put it on my to do list to create an issue for this.

kndehaan:
> what do we do with the emissions of energy_methanol_production_coal_gas?

We didn't put the `emissions` group on this node as this is a dead end in the graph. Only coal gas goes in, no carrier goes out. If `emissions` group is specified here, it gives incorrect emission results based on ingoing coal gas and non-CO2 going out, therefore all CO2 in coal gas being emitted.

datasets/DE_germany/emissions.csv
```diff
@@ -0,0 +1,42 @@
...
+Agriculture,Non-specified,non_energetic,co2,kton CO2-eq,2730.80358
+Agriculture,Non-specified,non_energetic,other_ghg,kton CO2-eq,55794.20352
+Buildings,Non-specified,energetic,other_ghg,kton CO2-eq,178.35372
+Bunkers,International aviation,energetic,other_ghg,kton CO2-eq,253.74532
+Bunkers,International navigation,energetic,other_ghg,kton CO2-eq,43.88243
+Energy,CCUS,non_energetic,co2,kton CO2-eq,0.0
+Energy,Electricity and heat production,energetic,other_ghg,kton CO2-eq,4075.69769
+Energy,Fuels production,energetic,other_ghg,kton CO2-eq,289.4041
+Energy,Fugitive emissions,non_energetic,co2,kton CO2-eq,1608.33747
+Energy,Fugitive emissions,non_energetic,other_ghg,kton CO2-eq,2575.6488
```
mabijkerk, line 12: I'm reviewing the emissions for Germany and I don't know how to reproduce this. I would assume this row equals 1.B Fugitive emissions, but the emissions on ETDataset (input_emissions_UNFCCC_API_DE_germany) do not match with the emissions here:

Image

mabijkerk: In the data_export_direct_emissions_v4 it seems to include all of 1.B, though way too much detail?

Image

datasets/DE_germany/emissions.csv
```diff
@@ -0,0 +1,42 @@
...
+Energy,Hydrogen production,non_energetic,other_ghg,kton CO2-eq,0.0
```
mabijkerk, line 13: I don't see this data row in the UNFCCC input (2B10a). Where do we get the information from? There is no more detail than 2B10.

datasets/DE_germany/emissions.csv
kndehaan: Fugitive emissions is all children of `1.B` excluding `1.B.2.a.iv`, which is allocated to refineries. Do you then get the correct result?

Regarding the detail, we haven't yet discussed how to write down the categories it includes when one of the subsub categories is excluded. We could also write there 1.B excl 1.B.2.a.iv, but currently I took the approach of writing it out more explicit.

datasets/DE_germany/emissions.csv
```diff
@@ -0,0 +1,42 @@
...
+Energy,Methanol production,non_energetic,other_ghg,kton CO2-eq,0.0
+Households,Non-specified,energetic,other_ghg,kton CO2-eq,1184.53801
+Industry,Aluminium,non_energetic,co2,kton CO2-eq,694.34578
+Industry,Aluminium,non_energetic,other_ghg,kton CO2-eq,81.4782
```
mabijkerk, line 17: These aluminium rows come from 2C3 Aluminium, but shouldn't part of the other GHG from 1A2b  Non-ferrous also be allocated to Industry,Aluminium,energetic,other_ghg?

For Germany the other GHG for Non-ferrous is only 1.4 kton, so this is very very minor and not worth to pick up - just good to be aware of that we do not do this now.

mabijkerk: Or wait, that's al included in Industry,Non-specified,energetic,other_ghg,kton CO2-eq right?

datasets/DE_germany/emissions.csv
kndehaan: 2.B.10 is present for Germany`etdataset/source_analyses/DE_germany/2019/15_emissions/input_emissions_UNFCCC_API_DE_germany.csv`. 

However, Hydrogen production is actually category 2.B.10.a according to IPCC, which is not present in the UNFCCC input data via API. The assumption here is made that if only 2.B.10 is reported, and not a distinction in 2.B.10.a Hydrogen production and 2.B.10.b Other, the emissions are allocated to the chemical sector and not hydrogen.

datasets/DE_germany/emissions.csv
```diff
@@ -0,0 +1,42 @@
...
+Industry,Chemicals,non_energetic,co2,kton CO2-eq,1326.32823
+Industry,Chemicals,non_energetic,other_ghg,kton CO2-eq,815.20834
+Industry,Fertilizers,non_energetic,other_ghg,kton CO2-eq,322.30302
+Industry,Food,non_energetic,co2,kton CO2-eq,0.0
+Industry,Food,non_energetic,other_ghg,kton CO2-eq,0.0
```
mabijkerk, line 22: In the data these appear to be empty, are they then assumed to be 0?

Image

datasets/DE_germany/emissions.csv
kndehaan: For combustion emissions, we indeed put the other GHG emissions for 1.A.2 entirely under Industry,Non-specified,energetic,other_ghg,kton CO2-eq.

datasets/DE_germany/emissions.csv
kndehaan: Yes

datasets/DE_germany/emissions.csv
mabijkerk: No, thanks it is good the way it is. Then I indeed arrive at the correct result, thanks!

datasets/DE_germany/emissions.csv
```diff
@@ -0,0 +1,42 @@
...
+Industry,Other metals,non_energetic,other_ghg,kton CO2-eq,75.3237
+Industry,Paper,non_energetic,co2,kton CO2-eq,0.0
+Industry,Paper,non_energetic,other_ghg,kton CO2-eq,0.0
+Industry,Refineries,energetic,other_ghg,kton CO2-eq,68.40507
+Industry,Refineries,non_energetic,other_ghg,kton CO2-eq,1.27471
+Industry,Steel,non_energetic,co2,kton CO2-eq,16948.07554
+Industry,Steel,non_energetic,other_ghg,kton CO2-eq,5.56444
+LULUCF,Emissions,non_energetic,co2,kton CO2-eq,0.0
+LULUCF,Emissions,non_energetic,other_ghg,kton CO2-eq,7706.51255
+LULUCF,Removals,non_energetic,co2,kton CO2-eq,14528.77232
```
mabijkerk, line 36: Are there no other_ghg removals?

datasets/DE_germany/emissions.csv
```diff
@@ -0,0 +1,42 @@
...
+Other,Indirect emissions,non_energetic,co2,kton CO2-eq,0.0
```
mabijkerk, line 37: How do we calculate indirect emissions? I cannot find it in the source file.

datasets/DE_germany/emissions.csv
kndehaan: By definition of IPCC guidelines the removals of other GHG is not possible. I also checked this in the UNFCCC data; none of the countries have negative other GHG.

datasets/DE_germany/emissions.csv
kndehaan: Indirect emissions are determined based on input IPCC file, filtering columns:
- category: Total GHG emissions with LULUCF
- measure: Indirect emissions

Point of discussion: we only take the value for indirect CO2 emissions and currently ignore if there are indirect N2O emissions. We should reconsider this: the indirect N2O emissions are only expressed in kg N2O in the two data source routes (UNFCCC API and UNFCCC Excel). 

Does this mean that expressing this into kg CO2eq would not make sense if we were to implement the indirect N2O emissions in the model? So should we then also express this in kg N2O? This would require some extra thought and work as we currently don't model N2O in the molecule graph...

mabijkerk (approved):
Methods are good to merge, any remaining comments are not must have and can be picked up later. Datasets need more review but are good enough for now, since it is the first beta release.

graphs/energy/nodes/energy/energy_production_methanol_synthesis_dry_biomass_ccs.ad
kndehaan: Issue added https://github.com/quintel/etdataset/issues/1090

PR #3446, opened 2026-06-09
Add a spec that checks for crude oil exceptions

mabijkerk:
#### Context
In the new emissions accounting method added in https://github.com/quintel/etsource/pull/3444/ we add an edge exception case for crude oil mix.

#### Implemented changes
I had AI write a spec that validates whether all relevant crude oil edges downstream of a final demand mixing node carry the exception group (to prevent missing edges) and that checks it the other way around (to prevent adding irrelevant edges).

I explitictly instructed it to use the Atlas graph builder.

## Checklist

### DIFF
--- a/spec/crude_oil_exception_spec.rb
+++ b/spec/crude_oil_exception_spec.rb
@@ -1,0 +1,87 @@
+require 'spec_helper'
+require 'set'
+
+# The :emissions_skip_crude_oil_mix edge group is an exception used by ETEngine's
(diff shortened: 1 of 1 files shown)


kndehaan:
Internally discussed that the work of this spec would still require dev effort in reviewing and modifying. As the crude oil exception is a temporary solution I don't think it's the effort to spend too much additional time on this spec. Therefore closing this issue.

Issue #1452, opened 2017-11-24
Local datasets and international transport


michieldenhaan:
In local datasets, the international transport allocation sliders currently add a scaled amount of energy to the scenario. For example in Groningen:

![inttrans](https://user-images.githubusercontent.com/32056448/33210937-758e65ba-d11c-11e7-81c6-579cd106745c.png)

CO2 emissions in 1990, on the other hand, are not scaled. The full NL 1990 value is added to the local dataset:

![co21990](https://user-images.githubusercontent.com/32056448/33210943-7b9162f0-d11c-11e7-8466-1d10ccfeba61.png)

michieldenhaan:
The second one (total NL CO2 emissions) is intended behaviour, but the first one (scaled energy) is not. Users should be able to allocate the full NL international transport energy consumption to a local dataset.

antw:
The problem is that demand is assigned to the aviation and shipping nodes using the `total_demand` nodes. *Those* demands are scaled with respect to the size of the region. I expect that they'll have to be given a custom value using initializer methods/inputs.

ChaelKruip:
> I expect that they'll have to be given a custom value using initializer methods/inputs.

Could they be exempt from scaling based on their node group (bunkers)?

antw:
> Could they be exempt

If we were going to do that, I'd prefer it be with a `scaling_except = true` attribute in the affected nodes (or something along those lines) rather than having a "magic" hard-coded group.

ChaelKruip:
> If we were going to do that, I'd prefer it be with a scaling_except = true attribute in the affected nodes (or something along those lines) rather than having a "magic" hard-coded group.

I agree that it makes sense to explicitly exclude them from scaling. 👍

marliekeverweij:
This issue still needs to be fixed :)

Just discussed with @michieldenhaan and @jorisberkhout about it and we are in favour of this solution: 
>If we were going to do that, I'd prefer it be with a scaling_exempt = true attribute in the affected nodes (or something along those lines) rather than having a "magic" hard-coded group.

This `scaling_exempt = true` attribute is not yet used/working right? How long would it take to implement this, @antw? And does this mean that in all other nodes need to have a `scaling_exempt = false` attribute?

antw:
> How long would it take to implement this, @antw?

I would guess a day or two.

> And does this mean that all other nodes need to have a scaling_exempt = false attribute?

Other nodes would not need to be explicitly set to `false`; we would use that as the default.

[...]

PR #3448, opened 2026-06-16
Emissions no zeroing of preset value for derived datasets

kndehaan:
#### Context
Preset demand of dataset value molecule nodes is set based on emissions.csv value. Preset values for derived datasets were set to zero before. This overwrite is now removed. 

#### Implemented changes
- Dummy emissions.csv files added to regional Dutch datasets
- Dummy fertilizers_ccus_demands.csv files added to regional Dutch datasets. This dummy file is not required for country datasets as these inherit from `eu` dataset which already contains this file (and is set to zero). 
- attribute `scaling_exempt` added to all dataset molecule nodes to avoid scaling of emissions.csv value

#### Related

Goes with atlas https://github.com/quintel/atlas/pull/193

## Checklist


mabijkerk:
This does not seem to work as expected:

- The data stream for those nodes in the molecule graph is very strange
- For nl2019, the data for the fertiliser processes doesn't add any information, but we cannot remove the DEMAND functions because then the model breaks
- The only actual 'new' information that is in the ccus_fertilizers_demand file is the molecule utilisation demand
- For the derived datasets, we can actually remove the ccus_fertilizers_demand completely and it doens't matter, they seem to be taking their demand from somewhere else, the energy graph?
- For combustion taking the values from the energy graph without DEMAND function does work.
- It is important to note that we do want the processes fertilizers part of the graph to dynamically calculate, because we don't want it to go to 0 if there is a fertilizer sector

Next steps:

- [ ] Figure out how the datastreams are handled for full and derived datasets: is there some first init that is being overwritten?
- [ ] Can we fix this technical bug?
- [ ] What data can we remove when the technical bug is fixed? Perhaps only the other utilisation is relevant to keep.

mabijkerk:
Checklist:

- Derived dataset, Zuid-Holland: still produces the same results as pro with the fertilizers_ccus_demands removed.
- Derived dataset, Austria: does not have the fertilizers_ccus_demands and still produces the same results as on pro.
- Full dataset, Netherlands: no changes, produces the same results as on pro.

kndehaan:
Problem is not yet solved... 

Taking `GM0599_rotterdam` as an example: when I query the demand for molecule node `molecules_other_utilisation_delayed_emitted_co2` when the fertilizers_ccus_demands.csv with zeros is still present in the dataset folder, I get the refinery error:

image

When I remove the fertilizers_ccus_demands.csv for this dataset and query the following local and on pro:

```
MV(molecules_other_utilisation_delayed_emitted_co2, demand)
```
it gives

pro: `0.0`
local: `393,920,859.20038027`

So removing the fertilizers_ccus_demands.csv but with the removal of setting demands of derived datasets to zero on atlas, the results are affected which is not desired. 

I still don't know how this value is determined, but I think we still need to dive deeper into this since we don't have a solution yet to this issue.. 

I expect that the other option of adding ETLocal keys & SGQs will not be an easier alternative as we might also stumble on double defining of values. 

@mabijkerk @louispt1 based on these findings, do you also think we need to dive deeper into steps mentioned in Mathijs' [comment](https://github.com/quintel/etsource/pull/3448#issuecomment-4719599914)?

louispt1:
#### Here's what I think is happening:
1. Molecule demands are set twice:
	- At **build**: The `~ demand = …` queries plus edge shares, compute a demand for every molecule node. This is needed so refinery can balance/validate the graph.
	- At **runtime**: `Qernel::Molecules::FinalCalculation` recomputes molecule demands, but only for `from_energy` nodes.
2. `molecules_other_utilisation_delayed_emitted_co2` is not a `from_energy` node, so its value comes entirely from generation at build time.
	- Before it was being zeroed by our friend `ZeroMoleculeNodes`, but after removing this fix the build time value is persisting
3. The `from_energy` nodes (combustion etc.) are unaffected by the zeroing at runtime - that is why for certain nodes removing the csv doesn't matter, those nodes are set at runtime anyway, so the build time values were not necessary.
4. When the zeroed csv is present, there is a refinery error on `molecules_other_utilisation_delayed_emitted_co2` because demand is double-defined, once by `fertilizers_ccus_demands` and once by the `reversed`, `parent_share = 1.0` share edge tying it to the parent's flow. While zeroing was on, everything in this branch was `0.0`, so the two definitions agreed (0 = 0). With zeroing removed, the upstream flow is non-zero but the **zeros-CSV still forces this node to 0.0** → refinery is over-determined (preset `0.0` vs inflow ≠ 0) → error. 

#### My proposed solution (for now):
Keep `ZeroMoleculeNodes` but add a clause excepting nodes with attribute `scaling_exempt`. As for the purpose of this project we only want to allow build time demand for emissions molecule nodes, and these values will be exempt from scaling, this solution should work (for now).

##### The validation (`MV(molecules_other_utilisation_delayed_emitted_co2, demand)`):
Before - GM055_rotterdam empty scenario - ZeroMoleculeNodes removed:
Pro: 0
Local: 68,733,382

After - GM055_rotterdam empty scenario ZeroMoleculeNodes re-instated and `scaling_exempt` skip clause added:
Pro: 0
Local: 0

Let me know what you think about this solution @kndehaan / @mabijkerk. I have pushed the atlas code change to `emissions-no-zeroing` but not updated the engine/etsource gemfiles.

kndehaan:
The proposed solution of using `scaling_exempt` as a flag for not zeroing demands seems to work. I think this is good enough for at least a temporary solution. 

One additional change that I want to do then is to remove `scaling_exempt` from the two existing nodes `bunkers_total_useful_demand_planes` and `bunkers_total_useful_demand_ships`. According to [this PR](https://github.com/quintel/etsource/issues/1452), it was originally introduced here to avoid scaling of demand for international transport. However, since `graph_methods` are defined on these nodes (I assume they were added later), the `scaling_exempt` doesn't have any effect on the nodes anymore and I think it was forgotten to remove the attribute. Thus the attribute can be safely removed from these nodes (I tested this, no changes in results).

This means that `scaling_exempt` will only be defined now on the molecule nodes containing the emissions data from the emissions.csv.

kndehaan:
After some back and forth, the approach taken to ensure the emissions values are set correctly on derived datasets is to lift the general set to zero functionality, and to create graph methods for the three molecule nodes from which the demand was set with `demand` function. 

This deviates from what is described in this PR description, but I'll keep it there for reference. 

**Implemented changes**
- emissions.csv dummy files added for regional (Dutch) datasets
- attribute `scaling_exempt` added to all dataset molecule nodes to avoid scaling of emissions.csv value
- graph methods added for 
  - industry_final_demand_for_chemical_fertilizers_network_gas_non_energetic_co2
  - industry_chemicals_fertilizers_emitted_co2
  - molecules_other_utilisation_delayed_emitted_co2

**Goes with**
- https://github.com/quintel/atlas/pull/193
- https://github.com/quintel/transformer/pull/22
- https://github.com/quintel/etsource/pull/3448
- https://github.com/quintel/etlocal/pull/694

mabijkerk:
I don't get why the interface elements are all zero:
image

But then the molecule graph says something different:
image

mabijkerk:
I get it now. After discussion with Kyra we've decided to drop the ETLocal branch and just insert the 0 placeholder values in the SGQ directly, ideally with a comment to an issue on ETEngine (or whatever repo is most relevant).

The idea is that we've now exposed an issue in the molecule graph initialisation. Users were not aware of this before (and don't have to be). Adding the placeholder 0 values to ETLocal therefore only creates confusion.

mabijkerk:
Functionally this works!

Semaphore fails on the new SGQs, the error message is quite incomprehensible (see below), but it may be that it expects and ETLocal counterpart...

image

According to AI it is because Atlas does not accept molecule edges for SGQs yet? In `SparseGraphQuery#graph_part` it suggests:

```ruby
def graph_part?
  EnergyNode.exists?(part)   || EnergyEdge.exists?(part) ||
  MoleculeNode.exists?(part) || MoleculeEdge.exists?(part)
end
```

noracato:
The test says that the graph method `demand` is not allowed for the node `industry_chemicals_fertilizers_emitted_co2` for SGQs.

We indeed did not update SGQ validation for the molecule graph. So your AI is right.

mabijkerk (approved):

PR #3451, opened 2026-06-22
Remove outdated queries and update missing query references

mabijkerk:
#### Context
Queries can reference other queries: `Q(query_key)`. Sometimes queries are removed or renamed. In this case, you have to do a manual check that they are no longer referenced by another query. There was no automatic check for this.

#### Implemented changes
I had AI write a spec to check for missing queries referenced in queries. Some commments:

-  There is a spec called `renamed_functions_spec`, which seems somewhat similar but I don't really understand its purpose. Should we retire that spec @aaccensi?
- This new spec does not check if a query is still referenced in ETModel, so there is still some manual labour involved - this will be difficult to add I expect?
- Ideally, after this new spec has been merged we would add another increment that checks whether queries refer to outdated nodes, edges, node groups. Or we can add a spec for orphaned queries.

From the first spec run came 15 failures. These have been addressed in individual commits. Goes with the clean up of the investment table: https://github.com/quintel/etmodel/pull/4735.

#### Review
Ideally, @aaccensi would check the spec and @kndehaan would check the actions for each query. I have added an Excel for this on OneDrive.

I would like to merge this with a merge commit and not squash the commits - to preserve commit history on the addressed queries.


noracato (approved):

aaccensi (approved):
The new `query_validation_spec` looks good to me!

Regarding the `renamed_functions_spec`, I don't think we should delete it.
What it does is to check a list of GQL functions that have been renamed at some point in time are not still being used anywhere with the old name. If we ever make more of this name changes (or we know of a name change that has happened and wasn't included in this test) we should add them to the list:
- INPUT_LINKS → INPUT_EDGES
- LG → EDGE_GROUP
- LINK → EDGE
- LINK_GROUP → EG
- OUTPUT_LINKS → OUTPUT_EDGES

gqueries/general/emissions/supply/power/emissions_captured_power_non_biogenic.gql
kndehaan, line 1: This query and the queries referenced by it are not used anymore, so isn't it better to remove them instead of correcting it?

gqueries/output_elements/output_series/sankey/crude_oil_in_source_of_biomass_products_industry_transformation_in_sankey.gql
kndehaan, line 1: This query is not used anymore, `crude_oil_in_source_of_biomass_products_industry_transformation_in_sankey` is used instead. Should we therefore remove this query?

gqueries/general/emissions/emissions_overview_energetic_built_environment_co2.gql
kndehaan, line 1: All queries in this folder starting with `emissions_overview` don't seemed to be used anymore. Should we just remove them?

gqueries/general/emissions/supply/power/emissions_captured_power_biogenic.gql
kndehaan, line 1: This query is in the end used by a query that's not used anymore. Remove this as well?

kndehaan (commented):
Looking good! See file specific comments and here a general comment:
 
ETFlex queries are still mentioned in etsource/gqueries/README.mdown. Actually, this file is very outdated and can be removed. The most important part of this file are now in the docs. Can you remove this file as well?

mabijkerk:
Good idea about the gqueries README, I have not removed it, but simplified it and referred to the docs @kndehaan.

I did see some styling that was not yet included in the docs. I added it there in a first draft PR. Will revise that later.

gqueries/general/emissions/supply/power/emissions_captured_power_biogenic.gql
mabijkerk: All the queries in the folder appear not to be used. I deleted the entire folder `emissions>supply>power`.

gqueries/general/emissions/emissions_overview_energetic_built_environment_co2.gql
mabijkerk: Done!

gqueries/output_elements/output_series/sankey/crude_oil_in_source_of_biomass_products_industry_transformation_in_sankey.gql
mabijkerk: I'm okay with leaving this query in for completeness.

PR #3457, opened 2026-07-08
Update ROI 2019 dataset ahead of Island of Ireland merge

claudiavalkenier:
Closed this PR and made a new one: https://github.com/quintel/etsource/pull/3489 (new branch instead of rebasing previous one)

#### Context
As part of a project to combine the Republic of Ireland (ROI) and Northern Ireland datasets into a single Island of Ireland dataset, this PR refines the existing ROI 2019 dataset.

#### Related
- https://github.com/quintel/etlocal/pull/699
- https://github.com/quintel/etdataset/pull/1093

## Checklist

Tagging @kndehaan as reviewer


datasets/IE_ireland/IE_ireland.derived.ad
```diff
@@ -9,7 +9,7 @@
 - has_semi_detached_houses = true
 - has_coastline = true
 - has_industry = true
-- has_lignite = true
+- has_lignite = false
```
kndehaan: Is this something you changed? If so, what was the reason that this was changed?

claudiavalkenier: This isn't something I directly changed actually... However, I note that in the previous dataset, a lignite key (input_energy_power_ultra_supercritical_lignite_production) used to be 6991 TJ and is changed to 0 TJ. (there is only one single coal-fired power plant in Ireland, and no lignite). Maybe this has changed automatically?

claudiavalkenier: @noracato Is it possible that `has_lignite` automatically changes?

claudiavalkenier: The `has_lignite` is not an ETLocal attribute. So if you re-export the dataset, it will be overwritten with the default value for `has_lignite`, which is false.

I'm changing this by hand for now and will open a separate issue for this.

kndehaan: Thanks for looking further into this @claudiavalkenier, good that you're creating an issue for this.

noracato: I created an issue with some background: see quintel/etlocal#700

claudiavalkenier:
@kndehaan I've exported the updated dataset from ETLocal to ETSource again, there a few corresponding adjustments. I also manually set `has_lignite` to `true` as we've discussed above.

PR #3459, opened 2026-07-08
Add sector mapping CSV, `sector_label`/`use` rollout, and validation

louispt1:
#### Context
- `config/sector_mapping.csv`: the canonical mapping from ETM sector labels to external classification schemes (IPCC CRT codes, klimaattafels, national sectors).
- `sector_label` attribute added across energy and molecule node files, joined by `(sector_label, use)`; `use` also added/aligned on molecule nodes. Stale per-node `emissions_*` groups removed now that mapping supersedes them.
- `spec/sector_mapping_spec.rb`: validates every labelled node's `(sector_label, use)` pair exists in the mapping, the join key (`sector_label == slug(sector)_slug(subsector)`) hasn't drifted, emissions-store coverage per mapping pair × GHG, scheme names don't collide with emissions-store sector keys, and a static scan of gqueries/inputs for literal `SECTOR`/`MSECTOR`/scheme-form `EMISSIONS` calls against unknown schemes or values.
- Bumps Atlas to pick up `Atlas::SectorMapping`.

Note: The spec is failing currently, presumably due to a mismatch with the mock sector_mapping file.

#### Related
[atlas](https://github.com/quintel/atlas/pull/198)
[etengine](https://github.com/quintel/etengine/pull/1777)
[etsource](https://github.com/quintel/etsource/pull/3459)

## Checklist


kndehaan:
Nice work! The mapping seems to generally work well. 

I updated the sector_mapping.csv to represent the correct one. An important change here is the removal of the "klimaattafel" mapping, as we want to first implement the ipcc crt code mapping; the klimaattafel mapping will be added in a later increment. 

I think now some parts of the ETSource spec don't work as they should anymore. Could you have a look and update this? I think it's best to use the ipcc_crt_code_agg instead. (Sorry for forgetting to tell you that Klimaattafel will be added in a later increment). Perhaps etengine and atlas need to be checked as well?

kndehaan:
I found some issues when a value of the columns contains white spaces:
For example, `SECTOR(emissions_subsector, "Steel")` works well, but for Electricity and heat production it gives this error:

image

Or another error example:
image

So it's something about handling the blank spaces that needs some fixing. @louispt1 could you look into this?

kndehaan:
Seems to work well now! I'll make the PR official, and:

Tagging @noracato to review the spec
Tagging @mabijkerk for general review of the changed logic on the nodes and the mapping file

config/sector_mapping.csv
mabijkerk, line 1: Shouldn't this file be called `emissions_sector_mapping` or `direct_emissions_sector_mapping`?

gqueries/modules/multi_year_charts/flexible_capacity_chart/myc_flexible_capacity_in_load_shifting_other_industry.gql
mabijkerk, line 1: Why is this query change part of this PR?

louispt1: One of the changes meant that whitespaces are properly respected in GQL. It shouldn't matter for the queries, but changing this structure makes the whitespacing consistent across the GQL (same goes for the * 100 changes etc). Without this change queries like: https://github.com/quintel/etsource/pull/3459#issuecomment-4924315269 would break

config/sector_mapping.csv
kndehaan: For now we leave it as is. Even though it's currently implemented only for emission-relevant nodes, it has the potential to be used for categorisation (e.g. costs, Klimaattafel for demands) other than emissions.

noracato (approved):

config/sector_mapping.csv
mabijkerk, line 1: In this file, why is for Buildings, Households and Agriculture the aggregated code the same as the non-aggregated code? Shouldn't we use 1A4 there?

More generally, the ipcc_crt_code_agg feels a bit random. Why don't we have a breakdown like crt_code_1st_order, crt_code_2nd_order etc.

mabijkerk, line 1: In the Chemical industry:

- For `Petrochemical and Carbon Black Production`, shouldn't we mention `Petrochemical and Carbon Black Production excl. Methanol`?
- And for `Other`, `Other excl. Hydrogen Production`?

kndehaan: This column is needed to get the desired categories for the IPCC chart. We discussed before that it's desired to split out Households, Buildings, Agriculture from category 1.A.4, therefore this has been split the way it is in ipcc_crt_code_agg.

kndehaan: I suggest to discuss this briefly and add the suggested columns in the next increment where we'll be adding the IPCC chart and add IPCC columns to the data export.

kndehaan: Good point, processed!

PR #3463, opened 2026-07-16
Direct emissions chart and dashboard items, mapping file, 1990 data

kndehaan:
#### Context
This PR adds the queries for a new chart and dashboard items, using a newly introduced mapping file structure. 

#### Implemented changes
- Queries for GHG emissions chart and dashboard items
- New mapping structure including mapping file and new sector label attribute
- Modified setup of the data export using the mapping structure
- 1990 emission data added to emissions.csv, following the emissions data pipeline

#### Related
Goes with pull requests:
- https://github.com/quintel/atlas/pull/199
- https://github.com/quintel/etengine/pull/1779
- https://github.com/quintel/etsource/pull/3463
- https://github.com/quintel/etmodel/pull/4753
- https://github.com/quintel/etdataset/pull/1095
- https://github.com/quintel/documentation/pull/343

## Checklist


mabijkerk:
I had AI do a full review of the branches, it flagged the following:

**LULUCF removals get the wrong sign in the mapped form**

The two 1990 LULUCF queries disagree about what "LULUCF" means:

- `direct_emissions_lulucf_co2_1990` names the pairs explicitly and negates
  removals:
  `SUM(EMISSIONS(lulucf_emissions, …), NEG(EMISSIONS(lulucf_removals, …)))`
- `direct_emissions_lulucf_other_ghg_1990` uses the new mapped form
  `EMISSIONS(emissions_sector, "LULUCF", other_ghg, 1990)`, and `sum_pairs`
  sums over **both** `lulucf_emissions` and `lulucf_removals`.

It's harmless today only because no dataset has an
`LULUCF,Removals,…,other_ghg` row — `nl` has removals for co2 only. The day
someone adds one, it gets added instead of subtracted, silently. Worth either
using the explicit-pair form for consistency, or leaving a comment noting the
mapped form is only safe while removals stay CO2-only.

config/direct_emissions_csv.yml
```diff
@@ -22,19 +39,26 @@
...
+# — any column of `config/sector_mapping.csv` may be surfaced, in any order, under any header label.
+# For example:
+#   - name: Use
+#     label: "Use"
+#     type: sector_mapping
+#     value: use
+#   - name: IPCC_CRT_code_agg
+#     label: "IPCC CRT code (aggregated)"
+#     type: sector_mapping
+#     value: ipcc_crt_code_agg
```
mabijkerk, line 52: Is this still valid?

kndehaan: It's still valid, it gives an example of how we could add a new column to the data export and how this relates to the columns in the mapping file.

kndehaan:
> It's harmless today only because no dataset has an
LULUCF,Removals,…,other_ghg row — nl has removals for co2 only. The day
someone adds one, it gets added instead of subtracted, silently. 

By IPCC definition, removals in LULUCF only apply for CO2. It's not possible to have removals for other GHG. I updated the descriptions in the doc strings to further clarify this. 

Note to ourselves: good practice to provide clear doc strings when queries deviate from "standard queries". This will also gives context to the AI review to understand why the query is different.

mabijkerk (approved):
Let's do it!

PR #3469, opened 2026-08-10
Simplified heat sliders buildings

jort-wolda:
#### Context
This PR simplifies the input sliders for the buildings.

#### Implemented changes
- Changed the new buildings input slider
- Removed the existing buildings input slider and replaced by demolished buildings
- Changed the insulation sliders for buildings to be a reduction in % instead of the typical heat demand in kWh/m2
- Added queries for a new figure, in which the number of buildings is shown alongside the housing stock

#### Related
Goes with pull requests:
- ...

## Checklist


inputs/demand/buildings/buildings_insulation/buildings_insulation_existing_buildings.ad
```diff
@@ -0,0 +1,21 @@
+# Sets the reduction in heating demand of existing buildings due to improved insulation. It updates:
+# - Useful demand for space heating
+# - Typical useful heat demand area attribute
+# - Investment costs for insulation
+#
+# Priority is set to 1 to go after number of buildings but before behaviour sliders.
+
+- query =
+    relative_change = DIVIDE(100 + USER_INPUT(), 100);
```
mabijkerk: It is good practice to use `100.0` in calculations like these. I think this ensures the engine will read it as a float instead of an integer.

mabijkerk:
If I leave the number of existing buildings as is and add the same amount of new buildings, I would expect both demands to be the same. Why is this not the case? Am I missing something? For a blank `nl2023` scenario I set the new buildings slider to the max:

image

image

**Update:** so this is because we the existing buildings come from the energy balance, while we use the following for the future buildings?
```
    insulation_value_new_buildings = AREA(typical_useful_demand_for_space_heating_buildings_future);
    typical_surface_area_per_unit = AREA(area_per_building_residence_equivalent);
    heat_demand_per_unit = insulation_value_new_buildings * typical_surface_area_per_unit * MJ_PER_KWH;
```

Didn't we want to make the new buildings have the same heat demand per building as the existing buildings?

**Conclusion:** we will address this in the next increment, together with the insulation costs.

inputs/demand/buildings/buildings_behaviour/buildings_space_heating_behaviour.ad
```diff
@@ -0,0 +1,17 @@
...
+# existing and new buildings.
+# Priority is set to 0 to go after number of buildings and insulation sliders.
+
+- query =
+    relative_change = DIVIDE(100 + USER_INPUT(), 100);
+    EACH(
+      UPDATE_WITH_FACTOR(V(buildings_useful_demand_for_space_heating_buildings_present), preset_demand, relative_change),
+      UPDATE_WITH_FACTOR(V(buildings_useful_demand_for_space_heating_buildings_future), preset_demand, relative_change)
+    )
+- priority = 0
```
mabijkerk, line 11: How does this relate to the `flexibility_outdoor_temperature` input that also sets the buildings heat demand with a factor and priority 0?

```
          UPDATE_WITH_FACTOR(V(buildings_useful_demand_for_space_heating_buildings_present), preset_demand, heat_factor),
          UPDATE_WITH_FACTOR(V(buildings_useful_demand_for_space_heating_buildings_future), preset_demand, heat_factor),
          UPDATE_WITH_FACTOR(V(buildings_useful_demand_cooling), preset_demand, cool_factor),
          UPDATE_WITH_FACTOR(V(agriculture_useful_demand_useable_heat), preset_demand, heat_factor)
        )
      },
      -> {},
    )
- priority = 0
```

inputs/demand/buildings/buildings_behaviour/buildings_space_heating_behaviour.ad
mabijkerk, line 1: The other development of demand sliders are set with a percentage per year change. What do we think is appropriate @kndehaan?

mabijkerk:
Done, with the review. All in all nice work! There are some points of discussion left. To final points:

- I didn't review the insulation costs, might be good to do this if/after we have revised it.
- It is still good to have a stress test: there are so many different factors that influence each other here. Perhaps with MT?

inputs/demand/buildings/buildings_behaviour/buildings_space_heating_behaviour.ad
mabijkerk: We want to keep it as a %, but to separate it clearly from the other sliders it should have its own group header.

inputs/demand/buildings/buildings_behaviour/buildings_space_heating_behaviour.ad
jort-wolda: It works fine like this. These inputs (flexibility_outdoor_temperature & buildings_space_heating_behaviour)

Have the same priority (0). Therefore the order is alphabetical (buildings_space_heating_behaviour goes first). But since both update with factor the order doesn't matter anyway. 

Just to be sure I did a quick test. If you reduce outdoor temperature, heat demand increases. If you then reduce the demand through behaviour by 50%, the increased heat demand is halved.

inputs/demand/buildings/buildings_insulation/buildings_insulation_existing_buildings.ad
jort-wolda: Changed it

inputs/demand/buildings/buildings_behaviour/buildings_space_heating_behaviour.ad
jort-wolda: Done

mabijkerk (approved):

PR #3472, opened 2026-08-19
Heat pump revisions for nl2023

mabijkerk:
#### Context

t.b.d.

#### Implemented changes

- **Node source analysis heat pumps**
    The following node source analysis have been updated. Most importantly the COPs and COP per degree. But other 
    values such costs are updated as well:
   - households_space_heater_heatpump_air_water_electricity.converter
   - households_water_heater_heatpump_air_water_electricity.converter
   - households_space_heater_hybrid_heatpump_air_water_electricity.converter
   - households_water_heater_hybrid_heatpump_air_water_electricity.converter
   
- **Residence source analysis nl2023**: updated the number of HHP, ...
- **Residence analysis:** the hardcoded HHP specs for space heating are updated, for hot water the input of network gas is set to 1.0 to match the COP cutoff of 6.0.

#### Related

Closes: https://github.com/quintel/etdataset/issues/1073
Partly addresses: https://github.com/quintel/etdataset/issues/900

Goes with: https://github.com/quintel/etdataset/pull/1104

## To do

- [x] Add updated node specs
- [ ] Review rake task or drop/cherry-pick to a different branch if not in scope


graphs/energy/nodes/households/households_water_heater_hybrid_heatpump_air_water_electricity.ad
```diff
@@ -23,24 +25,23 @@
...
 - fever.group = households_hot_water
-- fever.type = producer
 - fever.technology_curve_type = default
+- fever.type = producer
 - network_gas.profile = self: network_gas_input_curve
 - network_gas.type = consumer
 - electricity_output_capacity = 0.0
 - free_co2_factor = 0.0
-- heat_output_capacity = 0.0049
+- heat_output_capacity = 0.198
```
mabijkerk: @jort-wolda this seems rather high? 198 kW?

mabijkerk: And didn't we want to align it with the space heater?

graphs/energy/nodes/households/households_space_heater_heatpump_air_water_electricity.ad
```diff
@@ -1,43 +1,44 @@
...
+- fever.cop_per_degree = 0.0593
 - fever.defer_for = 4
 - fever.efficiency_balanced_with = ambient_heat
 - fever.efficiency_based_on = electricity
 - fever.group = space_heating
-- fever.type = producer
 - fever.technology_curve_type = tech_constant
-- storage.cost_per_mwh = 154736.8
+- fever.type = producer
+- storage.cost_per_mwh = 0.0
```
kndehaan, line 37: Is it correct that this value has been set to zero?

graphs/energy/nodes/households/households_water_heater_hybrid_heatpump_air_water_electricity.ad
jort-wolda: I see what went wrong. It accidentally uses 0.22 as a capacity for the boiler instead of 0.022
fever.capacity.network_gas is also changed from 0.022 to 0.22 for this one

And you're right, we did discuss to align with the space heater. I'll fix that

graphs/energy/nodes/households/households_space_heater_heatpump_air_water_electricity.ad
jort-wolda: Old node source analysis did not include storage cost, nor did the other heat pump node source analysis. The storage volume is also set to 0 so it should not matter.

graphs/energy/nodes/households/households_space_heater_hybrid_heatpump_air_water_electricity.ad
```diff
@@ -1,14 +1,19 @@
+- input.ambient_heat = 0.561977
+- input.electricity = 0.226111
+- input.network_gas = 0.20861
```
kndehaan: Note that below the input.ambient_heat, electricity and network_gas parameters are also defined with `~`, retrieving from (full) datasets and in`graph_methods` with sparse graph queries, retrieving from derived datasets. What is the intended data route: fixed efficiencies of dataset-dependent efficiencies?

kndehaan: I think the intended route is the dataset-dependent efficiencies, meaning that these hard-coded values should be removed.

graphs/energy/nodes/households/households_water_heater_hybrid_heatpump_air_water_electricity.ad
```diff
@@ -23,24 +25,23 @@
 - variable_operation_and_maintenance_costs_for_ccs_per_full_load_hour = 0.0
 - variable_operation_and_maintenance_costs_per_full_load_hour = 0.0
 - wacc = 0.02
-- fever.base_cop = 1.86
-- fever.capacity.electricity = 0.001294498
-- fever.capacity.network_gas = 0.022
+- fever.base_cop = 2.1044
+- fever.capacity.electricity = 0.001704
+- fever.capacity.network_gas = 0.22
```
kndehaan: Is the factor 10 increase of `fever.capacity.network_gas` compared to the old value correct?

graphs/energy/nodes/households/households_space_heater_hybrid_heatpump_air_water_electricity.ad
kndehaan: If these values are processed somewhere in the dataset, note that these shares currently sum to 1.0. It might be that this should be corrected somewhere.

graphs/energy/nodes/households/households_water_heater_hybrid_heatpump_air_water_electricity.ad
kndehaan: I think this increase might be related to the increase of `fever.capacity.network_gas`, I'm not sure which parameter "goes first" and determines the other.

graphs/energy/nodes/households/households_water_heater_heatpump_air_water_electricity.ad
```diff
@@ -21,20 +23,19 @@
...
 - fever.efficiency_based_on = electricity
 - fever.group = households_hot_water
-- fever.type = producer
 - fever.technology_curve_type = default
-- storage.cost_per_mwh = 50713.75
+- fever.type = producer
+- storage.cost_per_mwh = 0.0
 - storage.decay = 0.0
-- storage.volume = 0.005
+- storage.volume = 0.0
```
kndehaan: Is it correct that this value has been set to zero?

graphs/energy/nodes/households/households_water_heater_hybrid_heatpump_air_water_electricity.ad
jort-wolda: No it is not, see my comment on the heat_output_capacity comment from Matthijs. I've changed it in the NSA file on etdataset and manually changed it here.

graphs/energy/nodes/households/households_space_heater_hybrid_heatpump_air_water_electricity.ad
jort-wolda: You are right, these values should not be here. 
image

This was somewhat of a mistake to put them here but we wanted to keep te calculations. They should not have been included in the rake

graphs/energy/nodes/households/households_water_heater_heatpump_air_water_electricity.ad
jort-wolda: On second thought. It might not be. In the other three NSAs there's no storage volume but it does make sense to have hot water storage for a full-electric heat pump. While it might not be necessary for space heating or for a hybrid heat pump, as it uses the gas boiler for hot water, it should be included here. 

We might need to think about the storage costs as well then. I am not sure if it makes sense to include storage costs separately though. Since the investment costs cover the total cost for the heat pump already.

graphs/energy/nodes/households/households_space_heater_heatpump_air_water_electricity.ad
jort-wolda: Checked it again. For the heat pumps we assume 1 value for the total average investment costs. No separate storage costs are included.

kndehaan (approved):
Looking good!

PR #3474, opened 2026-08-20
Emissions inputs

robindesmit-quintel:
#### Context
To implement the direct emission method in the ETM, we need to create inputs in ETsource which retrieve the co2 emissions from sliders in the frontend.

#### Implemented changes
Creating new input files for the direct emissions.

## Checklist


kndehaan:
@mabijkerk could you review and test the changes related to international transport in these two commits: https://github.com/quintel/etsource/pull/3474/changes/a1fd4846310176f9dbb53cd4a6aeea3a76845ce8, https://github.com/quintel/etsource/pull/3474/changes/82e7897be1fd803deeeb3a7cba33a5c7bab6b793. 

These inputs are connected and I updated one of the priorities. Note that the user value for the emissions slider is corrected for the share of international transport that is accounted for. This is also a user instruction that we should communicate to the client.

kndehaan (commented):
Nice work! I only improved the LULUCF co2 slider since the nodes weren't updated properly for all use cases. This is now fixed. Only awaiting @mabijkerk review of the bunkers sliders now.

kndehaan:
@mabijkerk additional question: what do you think of the slider max? I now set it to three times the max value of all static emissions in the start year. But this number might not be recognisable to users (if eventually in the end they will be shown in the front-end). I also considered using the total GHG emissions in the start year as max value, but that turns out to be very high compared to some slider start year values...

inputs/emissions/greenhouse_gases/direct_emissions/emissions_agriculture_non_specified_non_energetic_co2.ad
```diff
@@ -0,0 +1,13 @@
...
+      MV(agriculture_non_specified_non_energetic_co2),
+      preset_demand,
+      PRODUCT(USER_INPUT(), THOUSANDS)
+    )
+- priority = 0
+- max_value_gql = present:PRODUCT(Q(direct_emissions_static_emissions_max_value), 3.0)
+- min_value = 0.0
+- start_value_gql = present:DIVIDE(MV(agriculture_non_specified_non_energetic_co2, demand), THOUSANDS)
+- step_value = 0.1
+- unit = T
```
mabijkerk: I would use kton for all these sliders. Even a minor municipality like Schiermonnikoog has emissions in kton. If they want to set stuff in more detail I'd rather set the step_value to 0.01 instead.

gqueries/general/emissions/direct_emissions/emissions_sector/miscellaneous/direct_emissions_static_emissions_bunkers.gql
mabijkerk, line 1: I'm not entirely convinced about the use of 'static'. This would suggest that they cannot change,  which is not the case: they can change through the inputs that we're adding here. Perhaps we can find a more fitting name?

inputs/demand/bunkers/bunkers_allocated_percentage_shipping.ad
```diff
@@ -1,7 +1,9 @@
 # Updates both useful demand for international shipping and other_ghg emissions value
 # Start year value for other_ghg is taken from emissions file and converted from kton to kg
+# Priority set to 1 to go before bunkers_useful_demand_ships and
```
mabijkerk, line 3: Sharp, this should work fine.

inputs/emissions/greenhouse_gases/direct_emissions/emissions_bunkers_international_aviation_energetic_other_ghg.ad
```diff
@@ -0,0 +1,22 @@
...
+      preset_demand,
+      PRODUCT(
+        PRODUCT(USER_INPUT(), THOUSANDS),
+        international_share
+      )
+    )
+- priority = 0
+- max_value_gql = present:PRODUCT(Q(direct_emissions_static_emissions_max_value), 3.0)
+- min_value = 0.0
+- start_value_gql = present:PRODUCT(EMISSIONS(bunkers_international_aviation, energetic, other_ghg), THOUSANDS)
```
mabijkerk: This might be confusing for users. The start value could say 300 kton but these would not show up in the emissions charts etc. until you've change the allocation percentage. I don't see a way to set up the slider differently, but it is definitely something to clarify when we show it in the frontend.

Perhaps we should stress in the input as well that these emissions only 'count' according to the allocation percentage?

mabijkerk:
International emissions inputs work as expected, well done!

gqueries/general/emissions/direct_emissions/emissions_sector/miscellaneous/direct_emissions_static_emissions_bunkers.gql
mabijkerk: Additionally, should we add these queries to gqueries/inputs_elements/max_value instead?

mabijkerk:
> what do you think of the slider max

The way it is set up right now feel a bit overly complicated to me, especially the MAX of the bunkers and the other emissions.

I would use one query to set the max, so 3 times the 'static' emissions might be fine. You could also go for 0.25 times the total emissions or something like that. I would put that factor in the root query, instead of repeating it in every input. 
Additionally, I would add to each input MAX(2 times its start value, the root query) to prevent some inputs from exceeding their max with the start value.

mabijkerk:
@kndehaan I made some final changes:

- Updated the max values (and min value for LULUCF)
- Changed the units from T to kT

Could you do a final review?

mabijkerk (approved):

inputs/emissions/greenhouse_gases/direct_emissions/emissions_bunkers_international_aviation_energetic_other_ghg.ad
mabijkerk: I added this to the comment.

PR #3476, opened 2026-08-21
Update the transport nodes of electric bus, car and motorcycle.

robindesmit-quintel:
#### Context
Some of the current ETM transport efficiencies are not in agreement with the STREAM reports by CE Delft.
We want to update the transport efficiencies of electric passenger vehicles that deviate substantially.

#### Implemented changes
Update the transport efficiencies in the nodes of:
- electric transport: bus, car, motorcycle

#### Related
Goes with pull [#1106](https://github.com/quintel/etdataset/pull/1106)

## Checklist


kndehaan (approved):
Looking good!

PR #3480, opened 2026-08-27
Split households space heating and hot water electricity and network gas keys

kndehaan:
#### Context
Before, the same technology split of electricity and network gas to households space heating and hot water was applied for ETLocal datasets. These keys have now been split so the technology split can be set individually. 

#### Implemented changes
- Three new keys introduced for network gas to hot water technologies 
- Four new keys introduced for electricity to hot water technologies
- Modified sparse graph queries to include the new ETLocal keys

#### Related

Goes with pull requests https://github.com/quintel/etlocal/pull/712

## Checklist

### DIFF
--- a/sparse_graph_queries/households/households_water_heater_hybrid_heatpump_air_water_electricity+input.ad
+++ b/sparse_graph_queries/households/households_water_heater_hybrid_heatpump_air_water_electricity+input.ad
@@ -7,10 +7,10 @@
 # if total_heat is zero (i.e. there are no HHPs in the start year) the input slots shares are set equal to the default node values
 
 - query =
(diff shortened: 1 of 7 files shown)


sparse_graph_queries/households/households_water_heater_hybrid_heatpump_air_water_electricity+input.ad
mabijkerk, line 1: I made some final changes to the HHP SGQ @kndehaan, can you check this? Additionally, shouldn't we update the space heating SGQ as well? It should not make a difference currently, but it is more robust right?

kndehaan: I updated the space heating SGQ as well now, using the same setup of the hot water one. 

Changing the 0.0 to an actual calculation seems fine, but I think the result will anyhow always be zero here.

mabijkerk (approved):

PR #3479, opened 2026-08-27
Households insulation chart

jort-wolda:
#### Context
The etsource changes for the charts for the households - insulation section.

#### Implemented changes
- Added heat demand / house queries
- Restructured the heat, useful_demand, households queries by moving existing queries in a 'totals' folder and the new queries in 'per_household' folder
- Added queries for the new chart
- Deleted obsolete queries (for the removed total bar)

#### Related
Goes with pull requests:
- https://github.com/quintel/etmodel/pull/4796

## Checklist

- [x] I have tested these changes
- [ ] I have updated documentation as needed (I don't think documentation is needed for this)
- [x] I have tagged the relevant people for review

PR #3481, opened 2026-08-28
Heat new households improvement

jort-wolda:
#### Context
The previous calculation method used a kWh/m2 and m2/residence value to calculate the heat demand of new houses. This deviated from the method for existing houses and resulted in much higher heat demands than expected for new houses. To compensate this, the start value for heat demand for new houses is set equal to heat demand of the most recent existing building of that type.

This was not done for the heat demand per non-residential building since this 'issue' arise there.

#### Implemented changes
- Set heat demand for new residences equal to heat demand of residences from 2005-present

#### Related
This branch was created from the one below, which PR is still in review. That branch contains the new chart for residential heat demand.

Goes with pull requests:
- https://github.com/quintel/etsource/pull/3479

## Checklist


kndehaan (approved):
Looking good!

PR #3483, opened 2026-08-28
Emissions Klimaatsector

robindesmit-quintel:
#### Context
Add chart of direct emissions for the klimaattafel categories.

#### Implemented changes
- create new queries in folder gqueries>general>emissions>direct_emissions>klimaattafel
- create new queries in folder gqueries>output_elements>output_series>direct_emissions_per_klimaattafel_category_chart
These queries follow the klimaattafel categories. 
Pay attention to `internationaal_transport` and `landgebruik` as these queries deviate from the rest. 

#### Related
Goes with pull requests:
- https://github.com/quintel/etsource/pull/3483
- https://github.com/quintel/etmodel/pull/4799
- https://github.com/quintel/mechanical_turk/pull/228

## Checklist


gqueries/general/emissions/direct_emissions/klimaattafel/1990/per_category/other_ghg/direct_emissions_klimaattafel_internationaal_transport_other_ghg_1990.gql
```diff
@@ -0,0 +1,7 @@
+# Uses the query for ETM sector bunkers, which matches klimaattafel category Internationaal
+# transport.
+# This is done since using the klimaattafel mapping could not distinguish between
+# aviation and navigation for other GHG.
```
jort-wolda: Not a big deal but the explanation could be a little bit better. This query is used because this uses the sliders that indicate whether you want to include international transport or not, right?

Otherwise you could just use the sector mapping like for all other categories.

gqueries/general/emissions/direct_emissions/klimaattafel/1990/per_category/other_ghg/direct_emissions_klimaattafel_landgebruik_other_ghg_1990.gql
```diff
@@ -0,0 +1,6 @@
+# Uses the query for ETM sector LULUCF, which matches klimaattafel category Landgebruik.
```
jort-wolda: I don't understand why this one needs to be different. I see that this works as well but why not use the same format as everywhere else like so:

unit = MT
query =
DIVIDE(
EMISSIONS(klimaattafel, "Landgebruik", other_ghg, 1990),
THOUSANDS
)

jort-wolda (commented):
I've not looked at the sector mapping csv. Besides that it looks good! Only had a question about one query

kndehaan:
Note: I just got notified that it's more common to refer to "Klimaatsector" instead of "Klimaattafel". I processed this in the labels, queries, docstrings, folder names, etc.

kndehaan (approved):
Looking good!

kndehaan:
One more todo that came from the sprint review: put indirect emissions under Other sector instead of a separate category, which would be in line with how it's done for ETM sectors.

kndehaan:
On second thought, we leave the indirect emissions as separate category since the there's no Klimaatsector "Other".

PR #3486, opened 2026-09-03
Add new inputs for external coupling to CTM

robindesmit-quintel:
#### Context
The end results of emissions in industry and waste in the ETM coupled to the CTM, is not in agreement with the end result of emissions in the CTM alone. Therefore, new external coupling inputs are created such that the resulting emissions of the CTM can be entered directly into the ETM. 

#### Implemented changes
- Create new external coupling inputs for the emissions of industry and waste. 
- Add the correct coupling groups to the new inputs.
- Edit the original emissions inputs to include the coupling group in `disabled by coupling` .

## Checklist


jort-wolda (approved):
Looks good, no comments!

kndehaan (approved):
Looking good! Only awaiting one decision regarding kton unit with client before merging, but I expect no required changes.

PR #3487, opened 2026-09-04
External coupling co2 emissions

jort-wolda:
#### Implemented changes
External coupling co2 emissions:
- Steel
- Fertilizers transformations
- Refineries transformations
- Chemicals other transformations

#### Related
Goes with pull requests:
- https://github.com/quintel/etsource/pull/3487
- https://github.com/quintel/etengine/pull/1808
- https://github.com/quintel/etmodel/pull/4804

## Checklist


graphs/molecules/nodes/industry/industry_steel_external_coupling_node_co2.ad
```diff
@@ -1,14 +1,9 @@
...
-- from_energy.conversion.cokes = edges: primary_co2_emission_of_bio_and_fossil_without_capture_factor
-- from_energy.conversion.coal = carrier: co2_conversion_per_mj
-- from_energy.conversion.crude_oil = carrier: co2_conversion_per_mj
-- from_energy.conversion.network_gas = edges: primary_co2_emission_of_bio_and_fossil_without_capture_factor
-- from_energy.conversion.wood_pellets = carrier: co2_conversion_per_mj
+- groups = [preset_demand, direct_emissions]
 - input.co2 = 1.0
-- use = energetic
+- scaling_exempt = true
+- use = non_energetic
```
kndehaan: `use` should be the same for the energy node counterpart, which is `energetic` for the steel node. Could you change this here?

graphs/molecules/nodes/energy/energy_chemical_fertilizers_transformation_external_coupling_node_co2.ad
```diff
@@ -0,0 +1,9 @@
+# Attribute scaling_exempt set to true to avoid scaling of value read from emissions.csv
+
+- groups = [preset_demand, direct_emissions]
+- input.co2 = 1.0
+- scaling_exempt = true
```
kndehaan: After second thought, I don't think we need the `scaling_exempt` defined on the four molecule nodes. `scaling_exempt` was needed for the other molecule nodes since they read a dataset value in the `~ demand` line. A scaling was done for derived datasets (e.g. other country datasets), which was undesired. This was avoided with `scaling_exempt = true`.

For the four molecule nodes, however, the demand is set to 0, so I think scaling would not be applied. 

Could you test whether the line can be removed by removing it, setting values for the related input and see if you get strange results that might be related to the removed line. Test this please for both a full dataset (e.g. `nl2023`) and few derived datasets (other countries). If nothing strange happens, you can remove this line on all four molecule nodes.

If the attribute can be removed, you can also remove the docstring about this.

graphs/molecules/nodes/energy/energy_chemical_fertilizers_transformation_external_coupling_node_co2.ad
```diff
@@ -0,0 +1,9 @@
...
+- use = non_energetic
+- sector_label = industry_fertilizers
+
+~ demand = 0
```
kndehaan: Could you apply a float here (`0.0`), and also to the other molecule nodes?

kndehaan:
General discussion points:
- Unit of new inputs is in kton, all other absolute emission-related external coupling inputs (except new ones for static emissions) are in Mton. Should we apply Mton as well?
- Emission sliders added to the respective sector sections in the model, not to Greenhouse gas emissions section
- Emissions from the external coupling nodes are reported under Industry sector, whereas the node name indicated energy sector
- `industry_steel_external_coupling_node_co2` already existed, but it now works differently where CO2 is set via external coupling input. I therefore think we need a migration to preserve the "demand" on the node for existing scenarios.

inputs/modules/external_coupling/industry_chemical_fertilizers/external_coupling_energy_chemical_fertilizers_transformation_external_coupling_node_co2_emissions.ad
kndehaan, line 1: Could you move the four input files to the folders of the sub sector they belong to? From a quick glance, it seems that the sub folders also represent the sub sector that is in `coupling_groups`. So then it seems more logic that the emissions input files are also put into the folder that represents their coupling group.

kndehaan:
From the general discussion points, I'll discuss the kton or Mton decision with the client. The scenario migration will be picked up. The other two points don't need any action.

graphs/molecules/nodes/energy/energy_chemical_fertilizers_transformation_external_coupling_node_co2.ad
jort-wolda: I deleted it and it indeed still worked fine

kndehaan:
Confirmed with client that we can use Mton for the four inputs added with this PR. @jort-wolda could you process the unit change in the inputs and in the ETModel locales?

inputs/modules/external_coupling/industry_chemical_fertilizers/external_coupling_energy_chemical_fertilizers_transformation_external_coupling_node_co2_emissions.ad
```diff
@@ -0,0 +1,17 @@
...
+      preset_demand,
+      PRODUCT(USER_INPUT(), BILLIONS)
+    )
+- priority = 0
+- max_value_gql =
+    present:Q(direct_emissions_industry_total_ghg)
+- min_value = 0.0
+- start_value_gql = present:DIVIDE(MV(energy_chemical_fertilizers_transformation_external_coupling_node_co2, demand), BILLIONS)
+- step_value = 0.001
+- unit = kT
```
aaccensi: Now that is Mtons, this unit still needs to change I guess. This is the same for the 4 external coupling inputs.

jort-wolda: You're right, I somehow overlooked that. Changed it just now

kndehaan (approved):
Looking good! Wait for the migration to be finished and then we can merge.

